PR #11, opened 2018-09-20
Coded MerkleTree class to be used by both Rooter and Observer. 

thehenrytsai:


tests/lib/MerkleTree.spec.ts
```diff
@@ -0,0 +1,110 @@
+import Cryptography from '../../src/lib/Cryptography';
```
sylar217: Its good to use "baseUrl": "src/lib", in the compilerOptions property in tsconfig.json file to prevent relative path imports

src/lib/MerkleTree.ts
```diff
@@ -0,0 +1,219 @@
...
+/**
+ * Class containing a Merkle receipt structure and its related operations.
+ */
+class MerkleReceipt {
+}
+
+/**
+ * SHA256 hash function.
+ */
+function sha256hash (value: Buffer): Buffer {
```
dialtone4web: Shouldn't this be changed to a const crypto.createHash('sha256') rather the creating on every invocation?

src/lib/MerkleTree.ts
```diff
@@ -0,0 +1,246 @@
...
+   * Hashing function usinged by this Merkle tree.
+   */
+  private hash: (value: Buffer) => Buffer = Cryptography.sha256hash;
+
+  /**
+   * Intermediate structure for storing different sizes of balanced Merkle trees that will eventually form one final tree.
+   * [0] stores tree of 1 leaf, [1] stores tree of 2 leaves, [2] -> 4 leaves, [3] -> 8 leaves and so on.
+   */
+  private subtrees: (MerkleNode | undefined) [] = [];
+  private valueToMerkleNodeMap = new Map<Buffer, MerkleNode>();
```
sylar217: In sidetree's context why do we need this map, as we will be storing the values in IPFS. If we need it for linearization, then shouldn't that be taken care by the consumer of this class?

tests/lib/MerkleTree.spec.ts
thehenrytsai: Hi Anmol, I wanted to use baseUrl, but I am not able to get Jasmine to resolve the paths properly after 30 min of trying. I have created an [issue](https://github.com/decentralized-identity/sidetree-core/issues/12), feel free to make it work.

src/lib/MerkleTree.ts
thehenrytsai: No, you can create only one digest/hash per instance of Crypto.Hash created by createHash(…).

src/lib/MerkleTree.ts
thehenrytsai: The map is for quick construction of receipts. Linearization/serialization can be arguably implemented outside of this class, if and only if we require all values to be given in one shot, since the order of values is tightly couple with how the Merkle tree is constructed, so it makes sense to have the MerkleTree class dictate the order and the consumer can just call add(). But you brought up a good point that perhaps we require that all values needs to be given at the same time for now. So I will hide add(...) as a private function for now.

PR #16, opened 2018-09-24
Merkle tree - Added receipt creation and proof functions.

thehenrytsai:
Next will create the specification for the receipt in protocol doc.


sylar217:
As we are creating this libraries, we should also start creating the type definition files for them, to use it in a typescript file

thehenrytsai:
> As we are creating this libraries, we should also start creating the type definition files for them.
Not sure what you mean. since this is written in TS, type definition are generated when compiled already.

PR #18, opened 2018-09-25
Add did-cache interface

arvinda:


docs/implementation.md
```diff
@@ -12,9 +12,70 @@
...
 
+The DID cache holds most of the state of a Sidetree node. It is a singleton class with the following methods for state update and retrieval.
 
-# Merkle Rooter
+## Record
+
+This is the core method to update state with the following signature.
+
+```javascript
+public record(stateModifyOp: any, blockId: BlockId): URL
```
thehenrytsai: I'd like to change the term from "versionURL" to "operation hash". Because:

1. URL has  a specific meaning in the web world, would be good use another term to disambiguate. 
2. The hash ought to be the same hash as the "hash of an operation used to construct the Merkle tree", so the name corresponds to that explicitly.

thehenrytsai: I'd like to change the term from "versionURL" to "operation hash". Because:

1. URL has  a specific meaning in the web world, would be good use another term to disambiguate. 
2. The hash ought to be the same hash as the "hash of an operation used to construct the Merkle tree", so the name corresponds to that explicitly.

docs/implementation.md
```diff
@@ -12,9 +12,70 @@
...
+```
+The `stateModifyOp` is a JSON object representing a create, update, or a delete operation. Recall from the protocol description that the SHA256 hash of this object is the *versionURL* of the document version produced as the result of this operation. The `blockId` parameter identifies the blockchain block that this operation is anchored to.
```
thehenrytsai: Does the record method really need to return URL (operation hash) since the caller has the operation in the first place? Is this more of a convenience/sanity return value?

docs/implementation.md
```diff
@@ -12,9 +12,70 @@
...
+v0 -> v1 -> v2 -> v3
+```
+If we record a new update operation `v0 -> v4`, the new chain for the DID would be:
+```
+v0 -> v4
+```
+The above semantics of two updates referencing the same version is used to handle recovery. In the above example, a DID was compromised after version `v0` which resulted in unauthorized updates `v1, v2, v3`. A recovery operation was performed which invalidated `v1, v2, v3` and added version `v4`.
+
+In the above description, *older* refers to the logical time of the operation derived from the position of the operation in the blockchain.
+> TODO: Currently, blockId is used as the timestamp of an operation, but perhaps we need something with finer granularity (blockId, position within block). The following sequence of unlikely events would leave the DID cache without enough information to figure out the correct resolution in the current api - an update operation and a subsequent recovery operation are anchored in the same block and the CAS (IPFS) delivers these operations in reverse order.
```
thehenrytsai: I am inclined to say for v1 protocol, disallow multiple operations to the same DID in one batch. That way we table the discussion and implementation later.

docs/implementation.md
```diff
@@ -12,9 +12,70 @@
...
+```javascript
+public first(versionURL: URL): URL
+public last(versionURL: URL): URL
+public next(versionURL: URL): URL
+public prev(versionURL: URL): URL
+```
+
+## Resolve
+
+The resolve method returns the latest document for a given DID. It is implemented as `last(lookup(did))`.
```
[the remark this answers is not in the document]
thehenrytsai: Fixed.

docs/implementation.md
thehenrytsai: We will try to support multiple operation of the same DID in same batch.

docs/implementation.md
```diff
@@ -12,9 +12,70 @@
...
+
+When two update operations reference the same (prior) version of a DID document, the cache removes from consideration the older of the two operations and all prior operations referencing the deleted operation. This ensures that the document versions of a particular DID form a chain without any forks. For illustration, assume we have recorded four operations for a particular DID producing the following chain:
+```
+v0 -> v1 -> v2 -> v3
+```
+If we record a new update operation `v0 -> v4`, the new chain for the DID would be:
+```
+v0 -> v4
+```
+The above semantics of two updates referencing the same version is used to handle recovery. In the above example, a DID was compromised after version `v0` which resulted in unauthorized updates `v1, v2, v3`. A recovery operation was performed which invalidated `v1, v2, v3` and added version `v4`.
```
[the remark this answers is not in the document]
thehenrytsai: Updated the document and removed the recovery portion.

docs/implementation.md
thehenrytsai: Removed return type for now, but if we do, it will probably be a Buffer representing a multihash.

PR #43, opened 2018-10-23
Fixed Windows vs Linux file issue in unit test + updated operation chaining diagram in protocol doc

thehenrytsai:
Test - Fixed Windows vs Linux based file read differences causing unit test to fail.
package.json - Renamed 'tests' script to 'test' to adhere to npm convention thus removed warning.
Protocol doc - Updated and replaced the DID operation chaining diagram.
README - Added references to protocol and implementation docs.


tests/RequestHandler.spec.ts
```diff
@@ -24,14 +24,19 @@ describe('RequestHandler', () => {
     }
     blockchain.setLaskBlock(mockLastBlock);
 
-    const createRequest = readFileSync('./tests/requests/create.json');
+    // Read create operation request from file.
+    const requestString = readFileSync('./tests/requests/create.json', 'utf8');
+    const requestStringWithoutCarriageReturn = requestString.replace(/[\r]+/g, ''); // Remove all Carriage Returns in case of file on Windows.
```
sylar217: Instead of handling line endings in the code, can we force git to store file in either Windows or Linux format to avoid this ambiguity?

thehenrytsai: Hi @sylar217, great suggestion, let me try play around with it.

PR #45, opened 2018-10-24
Initial code for DID cache

arvinda:
Near-duplicate of a previous pull request that we accidentally merged and reverted.


src/DidCache.ts
```diff
@@ -1,16 +1,338 @@
...
+ * This usage is guaranteed to produce unique VersionId's since the operation contains
+ * as one of its properties the previous VersionId. Since the operation hash is
+ * just a string we alias VersionId to string.
+ *
+ * With this usage, the operation hash serves two roles (1) an identifier for an operation
+ * (2) an identifier for the DID document produced by the operation. In the code below,
+ * we always use VersionId in places where we mean (2) and an OperationHash defined below
+ * when we mean (1).
  */
-export interface DidCache {
```
thehenrytsai: Please keep this interface and perhaps rename the implementation class to InMemoryDidCache (until it's not), my upcoming unit test still depends on the mock DidCache.

tests/mocks/MockDidCache.ts
```diff
@@ -1,13 +0,0 @@
-import Transaction from '../../src/Transaction';
```
thehenrytsai: Can we please bring back this file? My upcoming unit test still depends on the mock DidCache.

src/DidCache.ts
```diff
@@ -1,16 +1,338 @@
...
+  /**
+   * Map a operation hash to the OperationInfo which contains sufficient
+   * information to reconstruct the operation.
+   */
+  private opHashToInfo: Map<OperationHash, OperationInfo> = new Map();
+
+  /**
+   * Apply (perform) a specified DID state changing operation.
+   */
+  public apply (operation: WriteOperation): string | null {
```
thehenrytsai: [TypeScript team guideline](https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines#null-and-undefined) suggests to use `undefined` instead of `null`, which makes sense. Apply the same guideline to all instances.

tests/DidCache.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+      "@type": "UserServiceEndpoint",
+      "instances": ["did:bar:456", "did:zaz:789"]
+    }
+  }]
+};
+
+// Implementation of Did document create - return a dummy did document
+function didDocumentCreate (_operation: WriteOperation): DidDocument {
+  return new DidDocument(didDocJson);
+}
```
thehenrytsai: Code from line 7 ~ 35 duplicated in MockDidDocGen.ts?

src/index.ts
```diff
@@ -1,21 +1,22 @@
...
-import MockDidCache from '../tests/mocks/MockDidCache'; // TODO: Replace once real implementation comes in.
+import { DidCache } from './DidCache';
 import Observer from './Observer';
 import RequestHandler from './RequestHandler';
 import Rooter from './Rooter';
 import { BlockchainClient } from './Blockchain';
 import { CasClient } from './Cas';
 import { Config, ConfigKey } from './Config';
 import { toHttpStatus, Response } from './Response';
+import { didDocumentUpdate, didDocumentCreate } from '../tests/mocks/MockDidDocGen';
```
thehenrytsai: Avoid using abbreviation. Not sure what Gen stands for, but perhaps MockDidDocumentGenerator?

tests/DidCache.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+
+
+// Implementation of a dummy cas class for testing - a simple hash map
+class DummyCas implements Cas {
```
thehenrytsai: Please remove this class and move the implementation logic to MockCas so we don't have multiple test CAS classes.

tests/DidCache.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+  bufs: Buffer[] = [];
+
+  public async write (content: Buffer): Promise<string> {
+    this.bufs.push(content);
+    return (this.bufs.length - 1).toString();
+  }
+
+  public async read (address: string): Promise<Buffer> {
+    const idx = +address;
```
thehenrytsai: What does idx stand for?

tests/DidCache.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+    return this.bufs[idx];
+  }
+}
+
+function createCreateOpBuf(): Buffer {
```
thehenrytsai: Team convention is to not use abbreviation unless needed. I am okay with abbreviation with internal parameters, but method/function definitions should follow convention. So: createCreateOperationBuffer. Same goes for all instances.

tests/DidCache.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+    createPayload,
+    signature: 'signature',
+    proofOfWork: 'proof of work'
+  };
+  return Buffer.from(JSON.stringify(createOpRequest));
+}
+
+async function createRootedOp(opBuf: Buffer, cas: Cas): Promise<WriteOperation> {
+  const batch: Buffer[] = [ opBuf ];
+  const batchBuffer = Buffer.from(JSON.stringify(batch));
```
thehenrytsai: Slightly different to schema currently defined in protocol doc:

{
  "operations": [
    "Base58 encoded operation",
    "Base58 encoded operation",
    ...
  ]
}

tests/DidCache.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+  const createPayload = Base58.encode(Buffer.from(JSON.stringify(didDocJson)));
+  const createOpRequest = {
+    createPayload,
+    signature: 'signature',
+    proofOfWork: 'proof of work'
+  };
+  return Buffer.from(JSON.stringify(createOpRequest));
+}
+
+async function createRootedOp(opBuf: Buffer, cas: Cas): Promise<WriteOperation> {
```
thehenrytsai: Function name misleading, an operation is not rooted unless blockchain layer is invoked to anchor the operation. I feel the function is mixing to actions together, but perhaps rename storeOperation(...)

src/DidCache.ts
thehenrytsai: Minor: not seeing the need for returning the operation hash externally.

tests/DidCache.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+describe('DidCache', () => {
+
+  it('should return non-null url for create op', async () => {
+    const dummyCas = new DummyCas();
+    const didCache = new DidCache(dummyCas, didDocumentUpdate, didDocumentCreate);
+    const createOp = await createRootedOp(createCreateOpBuf(), dummyCas);
+    expect(didCache.apply(createOp)).not.toBeNull();
+  });
+
+  it('first(did) should be did', async () => {
```
thehenrytsai: See [outstanding work](https://github.com/decentralized-identity/sidetree-core/issues/38), first(did) is no longer the DID.

tests/DidCache.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+    const createOp = await createRootedOp(createCreateOpBuf(), dummyCas);
+    const createRet = didCache.apply(createOp);
+    expect(createRet).not.toBeNull();
+
+    const did = createRet as string;
+    const firstVersion = await didCache.first(did);
+    expect(firstVersion).toBe(did);
+  });
+
+  it('last(did) should be did', async() => {
```
thehenrytsai: Misleading test description? Also seem to me that to really test last() and first(), shouldn't there be a chain of operations?

tests/DidCache.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+    const didCache = new DidCache(dummyCas, didDocumentUpdate, didDocumentCreate);
+    const createOp = await createRootedOp(createCreateOpBuf(), dummyCas);
+    const createRet = didCache.apply(createOp);
+    expect(createRet).not.toBeNull();
+
+    const did = createRet as string;
+    expect(didCache.last(did)).toBe(did);
+  });
+
+  it('prev(did) should be null', async() => {
```
thehenrytsai: Misleading/incomplete test description. Should follow [BDD style description](https://jasmine.github.io/tutorials/your_first_suite): e.g. "prev(did) should be null when ..."

tests/DidCache.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+    const dummyCas = new DummyCas();
+    const didCache = new DidCache(dummyCas, didDocumentUpdate, didDocumentCreate);
+    const createOp = await createRootedOp(createCreateOpBuf(), dummyCas);
+    const createRet = didCache.apply(createOp);
```
thehenrytsai: The initialization appears to be identical for all tests, consider refactoring using beforeEach, or even just moving the code out before each test.

tests/DidCache.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+
+  it('should resolve created did', async () => {
+    const dummyCas = new DummyCas();
+    const didCache = new DidCache(dummyCas, didDocumentUpdate, didDocumentCreate);
+    const createOp = await createRootedOp(createCreateOpBuf(), dummyCas);
+    const did = didCache.apply(createOp);
+    expect(did).not.toBeNull();
+
+    const resolvedDid = await didCache.resolve(did as string);
+    // TODO: can we get the raw json from did? if so, we can write a better test.
```
thehenrytsai: Isn't DidDocument already an object? Or are you asking for just the JSON string? Feel free to add to the library, but I don't see how it will benefit the test.

src/DidCache.ts
```diff
@@ -1,16 +1,338 @@
...
+ * the actual function hides details of json patching from Did cache.
+ */
+export interface DidDocumentUpdater {
+  (didDoc: DidDocument, operation: WriteOperation): DidDocument;
+}
+
+/**
+ * Function type that creates a Did document given a create/recover operation.
+ */
+export interface DidDocumentCreator {
```
thehenrytsai: Design:
1. I feel it is an overkill to define a DidDcoumentCreator interface because there is really only one implementation and we ought to test that implementation whenever we can, and since this implementation does not involve network calls, I don't see huge benefit for mocking and injecting it, but see the downside of more complex code (well, not crazy complex, but still).
2. Similar sentiment for DidDocumentUpdator, the desire to not hide patching details from the DidCache is valid, but I see it can be done by just declaring the function `static`.

src/DidCache.ts
```diff
@@ -1,16 +1,338 @@
...
+  public get lastProcessedTransaction (): Transaction | undefined {
+    return undefined;
+  }
+
+  /**
+   * Get a cryptographic hash of the write operation. Currently, uses
+   * SHA256 to get hashes (TODO: Fix it to be consistent DID generation)
+   */
+  private static getHash (operation: WriteOperation): OperationHash {
+    const sha256HashCode = 18;
```
thehenrytsai: Can't hard-code hash function to use, has to depend on protocol version ie. Protocol.getProtocol(...), this means either the operationHash or blockNumber needs to be given in the WriteOperation instances. Again, I personally think operationHash would do, and thus no need to return the hash again when apply() is invoked.

src/DidCache.ts
```diff
@@ -1,16 +1,338 @@
...
+ * Information about a write operation relevant for the DID cache, a subset of the properties exposed by
+ * WriteOperation.
+ */
+interface OperationInfo extends OperationTimestamp {
+  readonly batchFileHash: string;
+  readonly type: OperationType;
+}
+
+/**
+ * DIDCache is a singleton class whose instance holds most of the state in Sidetree node.
```
thehenrytsai: Nit: fix Casing: DidCache. There is no singleton pattern applied here, so consider rephrase the sentence.

src/DidCache.ts
```diff
@@ -1,16 +1,338 @@
...
+/**
+ * The timestamp of an operation. We define a linear ordering of
+ * timestamps using the function lesser below.
+ */
+interface OperationTimestamp {
+  readonly transactionNumber: number;
+  readonly operationIndex: number;
+}
+
+function lesser (ts1: OperationTimestamp, ts2: OperationTimestamp): boolean {
```
thehenrytsai: Consider rename to "earlier"

tests/DidCache.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+    const prev = await didCache.prev(did);
+    expect(prev).toBeNull();
+  });
+
+  it('should resolve created did', async () => {
+    const dummyCas = new DummyCas();
+    const didCache = new DidCache(dummyCas, didDocumentUpdate, didDocumentCreate);
+    const createOp = await createRootedOp(createCreateOpBuf(), dummyCas);
+    const did = didCache.apply(createOp);
+    expect(did).not.toBeNull();
```
thehenrytsai: Sanity: assuming you took the suggestion of using `undefined` instead of `null`, then need to changing these null checks also.

src/DidCache.ts
```diff
@@ -1,16 +1,338 @@
...
+   */
+  public apply (operation: WriteOperation): string | null {
+    const opHash = DidCache.getHash(operation);
+
+    // Ignore operations without the required metadata - any operation anchored
+    // in a blockchain should have this metadata.
+    if (operation.transactionNumber === undefined ||
+        operation.operationIndex === undefined ||
+        operation.batchFileHash === undefined) {
+      return null;
```
thehenrytsai: I think throwing error is more appropriate here.

src/DidCache.ts
```diff
@@ -1,16 +1,338 @@
...
+    // For operations that have a previous version, we need additional
+    // bookkeeping
+    if (operation.previousOperationHash) {
+      this.applyOpWithPrev(opHash, opInfo, operation.previousOperationHash);
+    }
+
+    return opHash;
+  }
+
+  public constructor (
```
thehenrytsai: Constructor goes above other member functions.

src/DidCache.ts
```diff
@@ -1,16 +1,338 @@
...
+    const sha256HashCode = 18;
+    const multihash = Multihash.hash(operation.operationBuffer, sha256HashCode);
+    const multihashBase58 = Base58.encode(multihash);
+    return multihashBase58;
+  }
+
+  /**
+   * Apply state changes for operations that have a previous version (update, delete, recover)
+   */
+  private applyOpWithPrev (opHash: OperationHash, opInfo: OperationInfo, version: VersionId): void {
```
thehenrytsai: Rename version to previousVersionId.

thehenrytsai: Sanity: Should opInfo: OperationInfo really be opTimestamp: OperationTimestamp?

thehenrytsai: Consider no abbreviation in method names. Also since `this.opHashToInfo.set(opHash, opInfo);` is actually not done here, this method is only applying partial/additional change (which asks the question if there is an alternative code structure), the function name is misleading.

src/DidCache.ts
```diff
@@ -1,16 +1,338 @@
...
+
+    // Ignore operations without the required metadata - any operation anchored
+    // in a blockchain should have this metadata.
+    if (operation.transactionNumber === undefined ||
+        operation.operationIndex === undefined ||
+        operation.batchFileHash === undefined) {
+      return null;
+    }
+
+    // opInfo is operation with derivable properties projected out
```
thehenrytsai: Following the same logic, opHash should be part of OperationInfo?

src/DidCache.ts
```diff
@@ -1,16 +1,338 @@
...
+  }
+
+  /**
+   * Apply state changes for operations that have a previous version (update, delete, recover)
+   */
+  private applyOpWithPrev (opHash: OperationHash, opInfo: OperationInfo, version: VersionId): void {
+    // We might already know of an update to this version. If so, we retain
+    // the older of previously known update and the current one
+    const prevUpdateHash = this.nextVersion.get(version);
+    if (prevUpdateHash !== undefined) {
```
thehenrytsai: This is very confusing to me, shouldn't the previous update hash be `version`? If so shouldn't `prevUpdateHash` be something like 'knownUpdateToGivenPreviousUpdate'?

src/DidCache.ts
```diff
@@ -1,16 +1,338 @@
...
+      const prevUpdateInfo = this.opHashToInfo.get(prevUpdateHash) as OperationInfo;
+      if (lesser(prevUpdateInfo, opInfo)) {
+        return;
+      }
+    }
+
+    this.nextVersion.set(version, opHash);
```
thehenrytsai: Sanity: should make sure timestamp of `opHash` is greater than `version`?

src/DidCache.ts
```diff
@@ -1,16 +1,338 @@
...
+
+  /**
+   * Apply state changes for operations that have a previous version (update, delete, recover)
+   */
+  private applyOpWithPrev (opHash: OperationHash, opInfo: OperationInfo, version: VersionId): void {
+    // We might already know of an update to this version. If so, we retain
+    // the older of previously known update and the current one
+    const prevUpdateHash = this.nextVersion.get(version);
+    if (prevUpdateHash !== undefined) {
+      const prevUpdateInfo = this.opHashToInfo.get(prevUpdateHash) as OperationInfo;
```
thehenrytsai: Should this be OperationTimestamp instead of OperationInfo? Since the code make a point of separating them out.

src/DidCache.ts
```diff
@@ -1,16 +1,338 @@
...
+   * hashmaps storing the core Did state and removes all entries with
+   * a greater transaction number.  In future, the implementation should be optimized
+   * for the common case by keeping a sliding window of recent operations.
+   */
+  public rollback (transactionNumber: number) {
+
+    // Iterate over all nextVersion entries and remove all versions
+    // with "next" operation with transactionNumber greater than the provided
+    // parameter.
+    this.nextVersion.forEach((opHash, version, map) => {
```
thehenrytsai: (version, nextVersion, map) to be consistent, since the code makes a point of making a distinction.

src/DidCache.ts
```diff
@@ -1,16 +1,338 @@
...
+      const opInfo = this.opHashToInfo.get(opHash) as OperationInfo;
+      if (opInfo.transactionNumber > transactionNumber) {
+        map.delete(version);
```
thehenrytsai: Sanity: do you mean map.delete(opHash)?

src/DidCache.ts
```diff
@@ -1,16 +1,338 @@
...
+  public constructor (
+    private readonly cas: Cas,
+    private readonly didDocUpdate: DidDocumentUpdater,
+    private readonly didDocCreate: DidDocumentCreator) {
+
+  }
+
+  /**
+   * Rollback the state of the DidCache by removing all operations
+   * with transactionNumber greater than the provided parameter value.
```
thehenrytsai: I think the definition and implementation of `rollback` needs to change to:

_"... greater than OR EQUAL to the provided parameter value."_

It doesn't make sense to say "rollbase X", but it actually mean roll back AFTER X.

src/DidCache.ts
```diff
@@ -1,16 +1,338 @@
...
+    // transactionNumber greater than the provided parameter.
+    this.opHashToInfo.forEach((opInfo, opHash, map) => {
+      if (opInfo.transactionNumber > transactionNumber) {
+        map.delete(opHash);
+      }
+    });
+  }
+
+  /**
+   * Resolve a did.
```
thehenrytsai: Nit since this is public function: casing: DID.

src/DidCache.ts
```diff
@@ -1,16 +1,338 @@
...
+   */
+  public async resolve (did: VersionId): Promise<DidDocument | null> {
```
thehenrytsai: @csuwildcat made a point recently that DID is no should be hash of the initial DID Document, not hash of operation, so the type should just be `string`. So the code probably needs another look up map of DID -> createOperationHash. There is an [issue](https://github.com/decentralized-identity/sidetree-core/issues/38) created to modify code already written, but we should not commit more code that violates this requirement.

src/DidCache.ts
```diff
@@ -1,16 +1,338 @@
...
+    } else {
+      return this.last(nextVersionId);
+    }
+  }
+
+  /**
+   * Get the last processed transaction.
+   * TODO: fix this after discussing the intended semantics.
+   */
+  public get lastProcessedTransaction (): Transaction | undefined {
```
thehenrytsai: This returned value is used as an offset marker to new transactions from blockchain.

src/DidCache.ts
```diff
@@ -1,16 +1,338 @@
...
 
   /**
-   * Applies the given DID operation to the DID Cache.
+   * Return the operation given its (access) info.
    */
-  apply (operation: WriteOperation): void;
+  private async getOperation (opInfo: OperationInfo): Promise<WriteOperation> {
+    const batchBuffer = await this.cas.read(opInfo.batchFileHash);
+    const batch = JSON.parse(batchBuffer.toString());
+    const opBuffer = Buffer.from(batch[opInfo.operationIndex].data);
```
thehenrytsai: Batch file not following protocol specified schema.

src/DidCache.ts
```diff
@@ -1,16 +1,338 @@
...
+   * Return the latest (most recent) version of a DID document. Return null if
+   * the version is unknown.
+   */
+  public last (versionId: VersionId): VersionId | null {
+    const opInfo = this.opHashToInfo.get(versionId);
+    if (opInfo === undefined) {
+      return null;
+    }
+
+    const nextVersionId = this.next(versionId);
```
thehenrytsai: Consider a non-recursive implementation to avoid stack overflow.

src/DidCache.ts
```diff
@@ -1,16 +1,338 @@
...
+    }
+    return null;
+  }
+
+  /**
+   * Return the first version of a DID document given a possibly later version.
+   * A simple recursive implementation using prev; not very efficient but should
+   * not matter since this method is not hit for any externally exposed DID
+   * operations.
+   */
```
thehenrytsai: Not sure what you mean by _"not hit for any of the externally exposed DID operations."_? Is there any internal usage of this method? Coding guideline says remove unused code until needed, in fact linter/compiler "normally" should pick this type of issue up.

src/DidCache.ts
```diff
@@ -1,16 +1,338 @@
...
+      } else {
+        return this.didDocUpdate(prevDidDoc, op);
+      }
+    }
+  }
+
+  /**
+   * Return the previous version id of a given DID version. The implementation
+   * is inefficient and involves an async cas read. This should not be a problem
+   * since this method is not hit for any of the externally exposed DID operations.
```
thehenrytsai: Why not store the previous version also and be efficient since the reverse (ie next version) is already stored? If the reasoning is that the method is not being used currently, then I'd say just remove it.

src/DidCache.ts
```diff
@@ -1,16 +1,338 @@
...
+    }
+
+    // Construct the operation using a CAS lookup
+    const op = await this.getOperation(opInfo);
+
+    if (this.isInitialVersion(opInfo)) {
+      return this.didDocCreate(op);
+    } else {
+      const prevVersion = op.previousOperationHash as VersionId;
+      const prevDidDoc = await this.lookup(prevVersion);
```
thehenrytsai: Consider non-recursive implementation to avoid stack overflow, though in this case it can be annoying.

src/DidCache.ts
```diff
@@ -1,16 +1,338 @@
...
+  /**
+   * Apply state changes for operations that have a previous version (update, delete, recover)
+   */
+  private applyOpWithPrev (opHash: OperationHash, opInfo: OperationInfo, version: VersionId): void {
+    // We might already know of an update to this version. If so, we retain
+    // the older of previously known update and the current one
+    const prevUpdateHash = this.nextVersion.get(version);
+    if (prevUpdateHash !== undefined) {
+      const prevUpdateInfo = this.opHashToInfo.get(prevUpdateHash) as OperationInfo;
+      if (lesser(prevUpdateInfo, opInfo)) {
```
thehenrytsai: Design: seems to me that with the approach of evaluating operations at the time of lookup, all the possible links needs to be preserved, not just the one with lower timestamp, because one with lower timestamp might be invalid such as bad signature.

src/DidCache.ts
```diff
@@ -1,16 +1,338 @@
...
+    }
+
+    return this.lookup(latestVersion);
+  }
+
+  /**
+   * Returns the Did document for a given version identifier.
+   */
+  public async lookup (versionId: VersionId): Promise<DidDocument | null> {
+    // Version id is also the operation hash that produces the document
```
thehenrytsai: It is unclear to me where signature verification will be performed, we don't have to do them now, but can you add a TODO comment inline indicating where the code will be invoked?

src/DidCache.ts
arvinda: Sure, retained the old interface and added an implementation cache.

src/DidCache.ts
arvinda: Done.

tests/mocks/MockDidCache.ts
arvinda: Done - but since the DidCache interface has new methods I had to extend the class a bit.

src/index.ts
```diff
@@ -1,21 +1,22 @@
 import * as getRawBody from 'raw-body';
 import * as Koa from 'koa';
 import * as Router from 'koa-router';
-import MockDidCache from '../tests/mocks/MockDidCache'; // TODO: Replace once real implementation comes in.
+import { DidCache } from './DidCache';
```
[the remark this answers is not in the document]
arvinda: Sorted by what? Can you please clarify?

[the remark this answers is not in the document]
arvinda: Pending your clarification I have sorted by filename ...

tests/mocks/MockDidDocGen.ts
```diff
@@ -0,0 +1,32 @@
+import { DidDocument } from '@decentralized-identity/did-common-typescript';
+import { WriteOperation } from '../../src/Operation'
+
+// Implementation of Did document update - no-op for testing purposes
+export function didDocumentUpdate (didDoc: DidDocument, _operation: WriteOperation): DidDocument {
+    return didDoc;
+  }
+
+  const didDocJson = {
```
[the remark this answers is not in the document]
arvinda: Done

src/index.ts
arvinda: Changed MockDidDocGen.ts => MockDidDocumentGenerator.ts

tests/DidCache.spec.ts
arvinda: DummyCas actually is more general than MockCas. I simply moved the DummyCas to MockCas and removed the old MockCas.

tests/DidCache.spec.ts
arvinda: index (into the bufs array) - added a comment to this effect.

tests/DidCache.spec.ts
arvinda: Changed the name to createOperationWithSingletonBatch which is probably more accurate

tests/DidCache.spec.ts
arvinda: I was looking at rooter.ts: getOperations() and rooter.ts: rootOperations() which stringifies/serializes a Buffer[] object. Leaving this unchanged now pending further discussion.

src/DidCache.ts
arvinda: apply returns a string which can be used as an input parameter to lookup/first/next/prev/next. This makes writing unit tests for didcache simple - the observer and rooter who do not need this return value can simply ignore it. Added @returns ...

tests/DidCache.spec.ts
arvinda: I am curious what the rationale for this change is - let us discuss this offline.

tests/DidCache.spec.ts
arvinda: I do intend to add more general tests with longer chain of operations soon. When I do so if these tests are subsumed by the general ones, we can remove them.

tests/DidCache.spec.ts
arvinda: it is not clear why. The previous of a first version should always be null, isn't it?

tests/DidCache.spec.ts
arvinda: Fixed it.

tests/DidCache.spec.ts
arvinda: if DID is not the hash of the createOp, then the implementation of did cache should be updated to reflect this, but this test itself should continue to be valid: it is making simple checks on the version list

tests/DidCache.spec.ts
arvinda: I did update the test description.

src/DidCache.ts
arvinda: This is mostly to get a working DidCache without needing to have a fully implemented json patch functions. Once we have those we could simply substitute those implementations. 

I have taken your suggestion and removed these from being exported to users of the class (e.g., the constructor now no longer needs these functions).

src/DidCache.ts
arvinda: I don't understand your suggestion. Even if apply() does not return anything, the did cache needs to match previousOperationHash in an update operation with a previous operation that has this hash. This means that the cache should either (1) have the ability to compute hashes of operations or (2) the caller (e.g., observer) needs to provide the operation hash as part of an Operation. 

If we go with (1) we need to provide the blockNumber in WriteOperation; if we go with (2) we need to provide operationHash in WriteOperation. I don't have a strong preference between (1) and (2).

src/DidCache.ts
arvinda: Fixed.

src/DidCache.ts
arvinda: Done

tests/DidCache.spec.ts
arvinda: The compiler forces you to :-)

src/DidCache.ts
arvinda: Makes sense. Fixed. (These are actually "asserts" since the observer can be trusted to never provide invalid operations - should we remove them altogether?)

src/DidCache.ts
arvinda: Fixed.

src/DidCache.ts
arvinda: Done.

arvinda: The current definition of OperationInfo extends OperationTimestamp was confusing. Changed it to OperationInfo { timestamp: OperationTimestamp } - I think the code reads better now. Please check.

arvinda: How about applyVersionChainUpdates()? If you can think of a more descriptive name please feel free to change or suggest ...

src/DidCache.ts
arvinda: This is not clear to me. The opHash is the key of the hashmap used to retrieve OperationInfo, so whenever we access OperationInfo we should have the opHash (otherwise we could not have looked up the hashmap). In other words, would this not be like storing a hash map (key, {key, value}) with key being duplicated in the value field? There is no harm in storing, but it seems unnecessary. Am I missing something?

src/DidCache.ts
arvinda: Just to make sure, this is a purely terminology change suggestion, right? What you suggest makes sense and I changed it accordingly - but just checking I did not miss something deeper in the logic of the code.

src/DidCache.ts
arvinda: Good point. Assuming observer is correct we should never get two operations with the same timestamp - so not checking for equality should be fine. Please double check.

src/DidCache.ts
arvinda: This issue should be auto-resolved given the fix for a previous comment (making timestamp a field/property of OperationInfo) - please check.

src/DidCache.ts
arvinda: Fixed per suggestion.

src/DidCache.ts
arvinda: Nice catch! Fixed the bug.

src/DidCache.ts
arvinda: I took your suggestion.

src/DidCache.ts
arvinda: Fixed.

src/DidCache.ts
arvinda: This has a dependency on how we fix an earlier issue (copied here):

"Even if apply() does not return anything, the did cache needs to match previousOperationHash in an update operation with a previous operation that has this hash. This means that the cache should either (1) have the ability to compute hashes of operations or (2) the caller (e.g., observer) needs to provide the operation hash as part of an Operation.

If we go with (1) we need to provide the blockNumber in WriteOperation; if we go with (2) we need to provide operationHash in WriteOperation. I don't have a strong preference between (1) and (2)"

If we go with (2) all hashes are provided by observer (or rooter may be) and there is not much change to DidCache code. If we go with (1) what you say is correct. Holding off on this until we resolve (1) vs (2) offline.

src/DidCache.ts
arvinda: You are correct - but I think this is consistent with current rooter code (that was what I used as a reference) - please check.

src/DidCache.ts
arvinda: Moved to an iterative implementation.

src/DidCache.ts
arvinda: Related to the previous issue. If we do decide to keep it would be for purely testing purposes where performance is not a critical issue. Whether or not we retain prev/first methods, I think we should not keep a hashmap version->prevversion since we probably want the state of the lean cache (without yet to be added perf optimizations) as small as possible. Does this sound reasonable?

src/DidCache.ts
arvinda: Moved to a non-recursive implementation.

src/DidCache.ts
arvinda: Added a TODO where signature verification could go.

src/DidCache.ts
arvinda: Will this issue not go away if we do signature checking before checking if the update can be applied?

tests/DidCache.spec.ts
thehenrytsai: Added code to remove duplicated DID Document in tests.

src/DidCache.ts
thehenrytsai: Fixed remaining null references in documentation and tests.

tests/DidCache.spec.ts
thehenrytsai: You are right, will make the corresponding change and add a test for the rooter also.

src/DidCache.ts
thehenrytsai: I don't see the added @returns, I will add it.

tests/DidCache.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+
+    const resolvedDid = await didCache.resolve(did as string);
+    // TODO: can we get the raw json from did? if so, we can write a better test.
+    expect(resolvedDid).not.toBeNull();
```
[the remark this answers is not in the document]
thehenrytsai: 1. Feel free to submit PR to did-common-typescript for DidDocument extensions.
2. We can still verify what is currently available in the DidDocument, even if it's just DID and public keys. I will file an issue for improving the DID Cache tests and assigning it to you to track this TODO.

tests/DidCache.spec.ts
thehenrytsai: Compiler can't force you to call a certain function. I have made the changes.

src/DidCache.ts
thehenrytsai: I updated method summary and implementation doc accordingly.

src/DidCache.ts
thehenrytsai: I have filed and assigned to you an [issue](https://github.com/decentralized-identity/sidetree-core/issues/51) to track this.

src/DidCache.ts
thehenrytsai: I've fixed this.

src/DidCache.ts
thehenrytsai: Fixed a bug and improved implementation.

src/DidCache.ts
thehenrytsai: This is still recursive, Created an [issue](https://github.com/decentralized-identity/sidetree-core/issues/52) and assigned to you.

src/DidCache.ts
thehenrytsai: We are in agreement, creating an issue to track this issue.

src/DidCache.ts
thehenrytsai: Not feeling to strongly about this, basically this stems from the thought that if opHash was part of opInfo, then you only need to pass opInfo around instead of passing them separately. Filing an issue to track this.

src/DidCache.ts
thehenrytsai: Created an issue to track this.

src/DidCache.ts
thehenrytsai: 1. Not asking for checking for equality, but the case of an invalid operation pointing to an operation further down the chain from it. Using this [issue](https://github.com/decentralized-identity/sidetree-core/issues/53) to track it.

src/DidCache.ts
thehenrytsai: I do not believe so, because signature checking can't be performed if there is a missing parent operation while there is a fork. Opened [issue](https://github.com/decentralized-identity/sidetree-core/issues/53) to track this.

PR #50, opened 2018-10-29
avoid using the same name for two APIs of the blockchain service

srinathsetty:
We were using /transactions to both anchor and fetch Sidetree transactions from the underlying blockchain. This commit fixes the naming scheme.


srinathsetty:
closing the PR, as discussed.

PR #58, opened 2018-11-14
Remove implementation-specific labels and branding

jcnelson:
This PR updates the overview diagram to remove the Bitcoin logo and IPFS, in favor of "blockchain peer network" and "content-addressed storage system."  The specification does not mandate these particular implementations; as an implementer, this PR would make the diagram more consistent with the text.


thehenrytsai:
Thanks Jude, never got around to updating this diagram from the original spec.

PR #59, opened 2018-11-15
Arvind/operationprocessor

arvinda:
Redesigned operation processor based on recent discussions.


src/OperationProcessor.ts
```diff
@@ -113,33 +151,56 @@ function earlier (ts1: OperationTimestamp, ts2: OperationTimestamp): boolean {
...
 /**
- * Information about a write operation relevant for the DID cache, a subset of the properties exposed by
- * WriteOperation.
+ * Information about a write operation relevant for maintaining OperationProcessor state.
  */
 interface OperationInfo {
   readonly batchFileHash: string;
   readonly type: OperationType;
   readonly timestamp: OperationTimestamp;
+  readonly parent?: OperationHash;
```
thehenrytsai: Nit: parent type is more appropriate/consistent to be `VersionId` (like `missingAncestor)?

src/OperationProcessor.ts
```diff
@@ -113,33 +151,56 @@ function earlier (ts1: OperationTimestamp, ts2: OperationTimestamp): boolean {
...
 
   /**
    * Map a operation hash to the OperationInfo which contains sufficient
    * information to reconstruct the operation.
    */
   private opHashToInfo: Map<OperationHash, OperationInfo> = new Map();
 
+  /**
+   * Map a valid versionId to the next valid versionId whenever one exists. Due to validity checks,
+   * (condition 3 in comment above Valid), the next valid version is uniquely defined if it exists.
```
thehenrytsai: _...he next valid version is uniquely defined if it exists._ -> Can you reword? I don't get the statement.

src/OperationProcessor.ts
```diff
@@ -427,11 +464,134 @@ class DidCacheImpl implements DidCache {
...
+
+    return true;
+  }
+
+  private invalidatePreviouslyValidOperation (opHash: OperationHash) {
+    const opInfo = this.opHashToInfo.get(opHash) as OperationInfo;
+    const nextOperation = this.nextVersion.get(opHash);
+
+    if (nextOperation !== undefined) {
+      this.invalidatePreviouslyValidOperation(nextOperation);
```
thehenrytsai: Sanity: doesn't look like recursion is necessary since from line 282 ~ 295 the code is looping through each incorrect operation and rolling them back, so just remove nextVersion[parent] would do. Taking the suggested approach would also:

1. Allow a more descriptive method name, something like `removeParentsNextLink(...)`.
2. Avoid potential stack overflow due to recursion.

Note: pending CR over usage on line 531.

src/OperationProcessor.ts
```diff
@@ -427,11 +467,134 @@ class DidCacheImpl implements DidCache {
...
+        // fall through ...
+      }
+    }
+
+    opInfo.status = OperationStatus.Valid;
+    this.nextVersion.set(parentOpHash, opHash);
+    return;
+  }
+
+  private updateWaitingDescendants (missingAncestor: OperationHash, opHash: OperationHash) {
```
thehenrytsai: Consider renaming method to "addWaitingDescendant(...), seems more accurate.

src/OperationProcessor.ts
```diff
@@ -427,11 +467,134 @@ class DidCacheImpl implements DidCache {
...
+    this.processDescendantsWaitingOn(opHash);
+  }
+
+  private processCreateOperation (_opHash: OperationHash, opInfo: OperationInfo): void {
+    // TODO: Validate create operation (verify signature)
+    opInfo.status = OperationStatus.Valid;
+  }
+
+  private processOperationWithParent (opHash: OperationHash, opInfo: OperationInfo): void {
+    const parentOpHash = opInfo.parent as OperationHash;
```
thehenrytsai: Consider using `opInfo.parent!` instead, same goes for all occurrences.

src/OperationProcessor.ts
```diff
@@ -427,11 +467,134 @@ class DidCacheImpl implements DidCache {
...
+
+    // Assert: parentOpInfo.status === OperationStatus.Valid. Validate the operation
+    if (!this.validate(opHash, opInfo)) {
+      opInfo.status = OperationStatus.Invalid;
+      return;
+    }
+
+    // The operation is intrinsically valid. Before we set it to valid, we need
+    // to ensure that it is the earliest sibling among child nodes of its parent.
+    const curEarliestSiblingHash = this.nextVersion.get(parentOpHash);
```
thehenrytsai: Nit avoid abbreviation: consider `earliestSiblingHash` or `parentsNextHash`

src/OperationProcessor.ts
```diff
@@ -427,11 +467,134 @@ class DidCacheImpl implements DidCache {
...
+
+    if (curEarliestSiblingHash !== undefined) {
+      const curEarliestSiblingInfo = this.opHashToInfo.get(curEarliestSiblingHash) as OperationInfo;
+      if (earlier(curEarliestSiblingInfo.timestamp, opInfo.timestamp)) {
+        opInfo.status = OperationStatus.Invalid;
+        return;
+      } else {
+        this.invalidatePreviouslyValidOperation(curEarliestSiblingHash);
```
thehenrytsai: Can we fetch all the descendants then literatively call this method similar as the `foreach(...)` on line 282? I am basically trying to reuse the `invalidatePreviouslyValidOperation(...)` method while keeping the comment for line 571.

src/OperationProcessor.ts
arvinda: Sure, fixed.

src/OperationProcessor.ts
arvinda: Tried to reword for clarity. Please check.

src/OperationProcessor.ts
arvinda: I think we need to keep the recursion (or iterate) since this function is called from within processOperationWithParent() which relies on the fact that this function removes not just the version but all its descendants as well. Leaving it on for now.

src/OperationProcessor.ts
```diff
@@ -427,11 +467,134 @@ class DidCacheImpl implements DidCache {
...
+
+  private processOperationWithParent (opHash: OperationHash, opInfo: OperationInfo): void {
+    const parentOpHash = opInfo.parent as OperationHash;
+    const parentOpInfo = this.opHashToInfo.get(parentOpHash);
+
+    // If we do not know about the parent, then the ancestry of this operation is
+    // incomplete. We leave the status to be Unvalidated and update the waitingDescendants
+    // of the parent operation (hash).
+    if (parentOpInfo === undefined) {
+      // assert: opInfo.statue === OperationStatus.Unvalidated
```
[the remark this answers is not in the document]
arvinda: I fixed the typo. Are you sure you want the assignment here? We do this assignment in lines 241-247 just a couple of instructions before reaching this line (-> processInternal -> processOperationWithParent).

src/OperationProcessor.ts
```diff
@@ -427,11 +467,134 @@ class DidCacheImpl implements DidCache {
...
+      }
+    }
+
+    opInfo.status = OperationStatus.Valid;
+    this.nextVersion.set(parentOpHash, opHash);
+    return;
+  }
+
+  private updateWaitingDescendants (missingAncestor: OperationHash, opHash: OperationHash) {
+    const curWaitingDescendants = this.waitingDescendants.get(missingAncestor);
```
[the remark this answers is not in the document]
arvinda: Sure - took your suggestion and changed it to waitingDescendants

src/OperationProcessor.ts
arvinda: Good suggestion - incorporated it.

src/OperationProcessor.ts
arvinda: I did not know about this operator. Changed the code in a lot of places to use this operator.

src/OperationProcessor.ts
arvinda: Done

src/OperationProcessor.ts
```diff
@@ -427,11 +467,134 @@ class DidCacheImpl implements DidCache {
...
+    opInfo.status = OperationStatus.Invalid;
+  }
+
+  private processDescendantsWaitingOn (opHash: OperationHash) {
+    const waitingDescendants = this.waitingDescendants.get(opHash);
+    if (waitingDescendants === undefined) {
+      return;
+    }
+
+    for (const descHash in waitingDescendants) {
```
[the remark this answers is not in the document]
arvinda: Expanded desc -> descendant everywhere.

src/OperationProcessor.ts
arvinda: using the foreach(...) like in 282 is very inefficient since it iterates over all the known operations. (We should fix the implement of recover() soon to avoid this.) Instead I changed the implementation of invalidatePreviouslyValidOperation() to be non-recursive. Does that help?

src/OperationProcessor.ts
thehenrytsai: I see that you've removed the recursion.

src/OperationProcessor.ts
thehenrytsai: What you did is great, I was just looking for an iterative implementation.

thehenrytsai:
PR out of date. Abandoning the current PR and rebase and creating a new PR for you to review in case I missed/screwed up anything when merge the two branches.

PR #64, opened 2018-11-28
OperationProcessorOutOfOrderTests

arvinda:


tests/OperationProcessor.spec.ts
```diff
@@ -40,53 +48,230 @@ async function addBatchOfOneOperation (opBuf: Buffer, cas: Cas, transactionTime:
     batchFileHash: batchFileAddress
   };
 
-  const op = WriteOperation.create(opBuf, resolvedTransaction, 0);
+  const op = WriteOperation.create(opBuf, resolvedTransaction, operationIndex);
   return op;
 }
 
+function getHash (operation: WriteOperation): string {
```
thehenrytsai: Avoid code duplication. Use the same implementation used by Operation Processor. You'll probably need to move it else where to use it. A logical place to put it would be in the WriteOperation class.

tests/OperationProcessor.spec.ts
```diff
@@ -40,53 +48,230 @@ async function addBatchOfOneOperation (opBuf: Buffer, cas: Cas, transactionTime:
...
+  const sha256HashCode = 18;
```
thehenrytsai: As mentioned previously can't hardcode this. Can use the same pattern found in WriteOperation.toDidDcoument().

tests/OperationProcessor.spec.ts
```diff
@@ -40,53 +48,230 @@ async function addBatchOfOneOperation (opBuf: Buffer, cas: Cas, transactionTime:
...
+  }
+  return factorial;
+}
+
+// Return a permutation of a given size with a specified index among
+// all possible permutations. For example, there are 5! = 120 permutations
+// of size 5, so by passing index values 0..119 we can enumerate all
+// permutations
+function getPermutation (size: number, index: number): Array<number> {
+  let permutation: Array<number> = [];
```
thehenrytsai: This is some crazy awesome implementation that I will not understand unless I spend a long time, so I will assume this works. Only nit is to use `const` for `permutation, j, t`.

tests/OperationProcessor.spec.ts
```diff
@@ -40,53 +48,230 @@ async function addBatchOfOneOperation (opBuf: Buffer, cas: Cas, transactionTime:
...
+      if (i === opHashes.length - 1) {
+        expect(await operationProcessor.next(opHashes[i])).toBeUndefined();
+      } else {
+        expect(await operationProcessor.next(opHashes[i])).toBe(opHashes[i + 1]);
+      }
+    }
+  });
+
+  it('should correctly process updates in every (5! = 120) order', async () => {
+    const ops = new Array(createOp!);
```
thehenrytsai: Consider generalizing the setup portion and verification portion of the code so similar code doesn't repeat 3 times.

tests/OperationProcessor.spec.ts
arvinda: Good suggestion. Moved it to WriteOperation class

tests/OperationProcessor.spec.ts
arvinda: The toDidDocument reference was helpful. Fixed.

tests/OperationProcessor.spec.ts
arvinda: Done (const suggestion)

PR #66, opened 2018-11-29
Various updates and fixes to enable Update REST API

thehenrytsai:
Protocol doc - Fixed typos in Update and Delete REST API examples + other minor fixes.
Operation Processor - Updated resolve() to accept fully qualified DID as input.

With this change Update REST API is officially working.


csuwildcat:
The DID Method prefix variable should be a config value, not a hard coded var.

thehenrytsai:
>           The DID Method prefix variable should be a config value, not a hard coded var.
It is a config value already.

csuwildcat:
Oh, I must have not read that right in the code snippet - sorry :)

PR #67, opened 2018-12-01
Added Delete REST API support + unit tests + standardized error + basic logger.

thehenrytsai:
* REST API - Added Delete operation support.
* Logger - Added a basic logger and stopped logs from polluting the screen during tests.
* Tests - Remove and replaced mock Operation Processor with the real one.
* Tests - Added Delete REST API unit tests for Request Handler.
* Tests - Added Update API tests to incease code coverage to 90+% for RequestHandler.


src/OperationProcessor.ts
```diff
@@ -252,6 +257,15 @@ class OperationProcessorImpl implements OperationProcessor {
     // previous info if it exists
     this.opHashToInfo.set(opHash, opInfo);
 
+    if (operation.type === OperationType.Delete ||
+        operation.type === OperationType.Recover) {
+      // NOTE: only assuming and hanldling delete currently.
+      // TODO: validate recovery key.
+
+      this.deletedDids.add(operation.did!);
```
arvinda: Don't we have to verify the signature for delete operation? This requirement introduces a dependency between the delete operation and its ancestors (which might define the public key to be used for verification) - which means that if some ancestor is missing and therefore the delete operation is unvalidated, we might have to postpone adding the deleted dids to the deletedDids set.

arvinda: Sounds good - thanks.

PR #82, opened 2018-12-18
AddOperationStoreClass

arvinda:
Add OperationStore class to cache and reduce OperationReconstruction costs

### DIFF
--- a/src/Cache.ts
+++ b/src/Cache.ts
@@ -1,0 +1,68 @@
+/**
+ * Standard cache interface that allows for the storage (store) and retrieval (lookup) of
+ * key-value pairs with false negatives: The lookup of a key might return undefined even
+ * if the key was previously stored, but if the lookup does return a defined value, the value is
+ * guaranteed to be that of the most recent store for the key.
+ */
+export interface Cache<KeyType, ValueType> {
+
+  /**
+   * Store a key-value pair in the cache.
+   */
+  store (key: KeyType, value: ValueType): void;
+
+  /**
+   * Lookup the value for a specified key with possible false negatives. See comment
+   * above the interface definition.
+   */
+  lookup (key: KeyType): ValueType | undefined;
+}
+
+/**
+ * A simple associative cache that stores the cached objects (key-value pairs)
+ * in a fixed size array along with a hashmap indexing the position of each
+ * stored key. The current implementation uses a simple random replacement
+ * policy.
+ */
+class FixedSizeCache<KeyType, ValueType> implements Cache<KeyType, ValueType> {
+
+  // Array storing the cached values
+  private readonly cachedKeys: Array<KeyType>;
+
+  private readonly cacheSlotValid: Array<boolean>;
+
+  // For any key currently stored in the cache, the map from the key
+  // to its slot in the cachedObjects array.
+  private readonly keyToValue: Map<KeyType, ValueType> = new Map();
+
+  public constructor (private readonly size: number) {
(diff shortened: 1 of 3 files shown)


src/Cache.ts
```diff
@@ -0,0 +1,68 @@
...
+   * Lookup the value for a specified key with possible false negatives. See comment
+   * above the interface definition.
+   */
+  lookup (key: KeyType): ValueType | undefined;
+}
+
+/**
+ * A simple associative cache that stores the cached objects (key-value pairs)
+ * in a fixed size array along with a hashmap indexing the position of each
+ * stored key. The current implementation uses a simple random replacement
```
thehenrytsai: Consider a equally simple (if not simpler) but more effective (in the initial phase where all slots will guaranteed to be used.) scheme/implementation: first in first out/circular cache.

src/OperationProcessor.ts
```diff
@@ -157,6 +158,63 @@ interface OperationInfo {
...
+   * Store an operation in the store
+   */
+  public store (opHash: OperationHash, operation: WriteOperation) {
+    this.operationCache.store(opHash, operation);
+  }
+
+  /**
+   * Lookup an operation from the store
+   */
+  public async lookup (opHash: OperationHash, opInfo: OperationInfo): Promise<WriteOperation> {
```
thehenrytsai: Design: Since this is an abstraction layer, I think opInfo should not be passed in for a lookup, else it defeats the purpose of an abstraction; ie. the caller shouldn't need to pass in data needed in case of a miss-hit by the internal cache. This probably leads to the need to store all the opInfo within the OperationStore, and to avoid duplication, the opHashToInfo probably can be moved within the OperationStore, which is probably the right decision following this design.

src/Cache.ts
arvinda: Interestingly, the random strategy has provably good performance. That said, I think a good strategy would be to use something like a clock algorithm used in OS caches. I will transition it to that strategy sometime when you are away. For now, I will keep the random strategy - the random strategy should not be much worse than clock and we should get decent enough performance with large cache sizes (as you correctly point out).

src/OperationProcessor.ts
arvinda: Very nice suggestion. I implemented the suggestion - now lookup does not require an opInfo parameter.

I made one minor tweak to the second part of your suggestion. The OperationInfo interface contains a lot of details specific to OperationProcessor (e.g., status) which the OperationStore need not know about. I split the OperationInfo struct into two - OperationInfo (required by OperationProcessor) and OperationAccessInfo(required by OperationStore). The OperationAccessInfo struct contains informatoin such as batchFilehash which is required to reconstruct the operation from the cas.

src/OperationProcessor.ts
```diff
@@ -157,6 +158,63 @@ interface OperationInfo {
   missingAncestor?: VersionId;
 }
 
+/**
+ * An abstraction of a *complete* store for operations, exposing methods to store and
+ * subsequently retrieve operations using OperationInfo. Internally relies on a
+ * cache to lookup recent and/or heavily accessed operations; on a cache miss relies on
+ * an expensive CAS lookup to reconstruct the operation.
+ */
+class OperationStore {
```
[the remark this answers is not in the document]
arvinda: Done.

src/Cache.ts
thehenrytsai: Sounds good.

PR #83, opened 2018-12-18
Tests - Added code for generating files used for load testing using Vegeta.

thehenrytsai:


tests/generators/VegetaLoadGenerator.ts
```diff
@@ -0,0 +1,78 @@
...
+
+/**
+ * Class for generating files used for load testing using Vegeta.
+ */
+export default class VegetaLoadGenerator {
+
+  /**
+   * Creates a Create request followed by an Update request for each DID.
+   */
+  public static async generateCreateLoadFiles (uniqueDidCount: number, endpointUrl: string, absoluteFolderPath: string) {
```
sylar217: Everything looks great, Can you please add the params defintion?

sylar217: One more thing, how are planning to use this method?

thehenrytsai: Parameter description added.

I am using this for load testing request throughput and operation processing throughput.

thehenrytsai: This can be used for both local and cloud load testing.

PR #101, opened 2019-01-23
Arvind/deferred process

arvinda:
This PR implements deferred processing where most of the work of processing a write operation is pushed to the next resolve of the corresponding did.


src/OperationProcessor.ts
```diff
@@ -233,10 +240,22 @@ class OperationProcessorImpl implements OperationProcessor {
...
+    const did = this.getDid(operation, opHash);
+    this.getDeferredOperationsList(did).append(opHash);
+
+    return opHash;
+  }
+
+  /**
+   * Processes a specified DID state changing operation.
+   */
+  private async processDeferred (operation: WriteOperation, opHash: OperationHash): Promise<void> {
```
thehenrytsai: Consider renaming to `processOperation`.

thehenrytsai: Move this private method below all public methods (after `processDeferredOperationsOfDid`).

src/OperationProcessor.ts
```diff
@@ -268,8 +287,6 @@ class OperationProcessorImpl implements OperationProcessor {
 
     // Else the operation is a create or an update.
     await this.processInternal(opHash, opInfo);
-
-    return opHash;
   }
 
   /**
```
thehenrytsai: Rollback of `operationStore` and `deferredOperations` not handled.

src/OperationProcessor.ts
```diff
@@ -199,6 +199,12 @@ class OperationProcessorImpl implements OperationProcessor {
 
   private readonly operationStore: OperationStore;
 
+  /**
```
thehenrytsai: Sanity: On line 25, is it supposed to be "incomplete" as opposed to "complete"?

src/OperationProcessor.ts
```diff
@@ -233,10 +240,22 @@ class OperationProcessorImpl implements OperationProcessor {
...
+
+    return opHash;
+  }
+
+  /**
+   * Processes a specified DID state changing operation.
+   */
+  private async processDeferred (operation: WriteOperation, opHash: OperationHash): Promise<void> {
     // opInfo is operation with derivable properties projected out
     const opTimestamp: OperationTimestamp = {
```
thehenrytsai: Design: Should `OperationTimestamp` and `OperationInfo` be revisited? Since `this.opeartionStore` is readily available and is being used throughout already, maybe there is no point in copying/duplicating most of the same info like between line 255 ~ 266.

Perhaps `OperationInfo` can be evolved into just `OperationState`:

```javascript
interface OperationState {
  status: OperationStatus;
  missingAncestor?: VersionId;
}

```

src/OperationProcessor.ts
```diff
@@ -209,12 +215,13 @@ class OperationProcessorImpl implements OperationProcessor {
    *            1. The operation (of the same hash) is not process before; or
    *            2. The operation is processed before but this operation has an earlier timestamp.
    *          Returns undefined if the same operation with an earlier timestamp was processed previously.
+   *
```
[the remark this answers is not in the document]
arvinda: I am not so sure. It is still returning the opHash - like we discussed long back it is true that the opHash is not used by the observer but the tests still use it like before. 

But now that the unit tests know how to hash operations, we have the option of removing the opHash return. Is this what you are suggesting?

src/OperationProcessor.ts
arvinda: Done

src/OperationProcessor.ts
```diff
@@ -615,6 +634,47 @@ class OperationProcessorImpl implements OperationProcessor {
...
+
+  /**
+   * Iterate over the deferred (unapplied) operations of a did and process the same.
+   */
+  private async processDeferredOperationsOfDid (did: string): Promise<void> {
+    const deferredOperationsList = this.deferredOperations.get(did);
+    if (deferredOperationsList !== undefined) {
+      for (const deferredOpHash of deferredOperationsList) {
+        const deferredOperation = await this.operationStore.lookup(deferredOpHash);
+        await this.processDeferred(deferredOperation, deferredOpHash);
```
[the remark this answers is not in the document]
arvinda: Good suggestion. Implemented.

src/OperationProcessor.ts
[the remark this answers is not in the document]
thehenrytsai: It is fine to return the operation hash if tests still uses it, but the comments still appear obsolete? E.g, the method doesn't return undefined like it claims it could.

src/OperationProcessor.ts
```diff
@@ -199,6 +199,12 @@ class OperationProcessorImpl implements OperationProcessor {
 
   private readonly operationStore: OperationStore;
 
+  /**
+   * The list of deferred (unapplied) operations, stored as a mapping from each did to the list
+   * of deferred operations for the did.
+   */
+  private deferredOperations: Map<OperationHash, LinkedList<OperationHash>> = new Map();
```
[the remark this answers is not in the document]
thehenrytsai: I agree that perf optimizations need not be persisted and that we should not trying to persist it until operation store persistence is done.

I do not understand what you mean by deferredOperations being subsumed by OperationStore, I thought OperationStore should not know anything about the deferred operations, but a pure storage with perhaps some caching capability.

Let's chat further if needed.

src/OperationProcessor.ts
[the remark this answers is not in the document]
arvinda: yeah - makes sense. I fixed the signature to return Promise<string> - changed the OperationProcessor interface as well.

PR #125, opened 2019-02-13
Arvind/prep persistence

arvinda:
This PR (a) significantly simplifies the OperationProcessor code by avoiding any computation caching (all work is done at resolve time); (b) changes the OperationStore interface to prepare for persistence.

### DIFF
--- a/src/OperationProcessor.ts
+++ b/src/OperationProcessor.ts
@@ -1,69 +1,11 @@
 import * as Protocol from './Protocol';
+import { Cas } from '../src/Cas';
 import Cryptography from './lib/Cryptography';
 import DidPublicKey from './lib/DidPublicKey';
 import Document from './lib/Document';
-import { Cas } from './Cas';
 import { DidDocument } from '@decentralized-identity/did-common-typescript';
-import { LinkedList } from 'linked-list-typescript';
-import { getOperationHash, WriteOperation, OperationType } from './Operation';
-import { OperationStore } from './OperationStore';
-
-/**
- * Each operation that is submitted to OperationProcessor for processing
- * has one of the following status at any given point.
- *
- * Terminology useful for various comments below:
- *
- * Parent/Prev: For a non-create operation o this refers to the previous operation (version)
- * specified by o.
- * Ancestor: Transitive closure of the Parent relation.
- * Descendant: Inverse of Ancestor
- * Known Ancestry/Complete Ancestry: A predicate/property of an operation that indicates whether all the
- * ancestors of the operation are known (by the OperationProcessor).
- * Siblings: Two or more operations that point to the same parent.
- */
-enum OperationStatus {
-  /**
-   * An operation is in Unvalidated state if its ancestry is incomplete.
-   */
-  Unvalidated,
-
-  /**
-   * An operation is in Valid state if all the following conditions hold:
-   * (1) It's ancestry is complete
-   * (2) Validation checks such as signature verification pass
-   * (3) It is the earliest sibling in terms of its timestamp (see earliest interface below).
-   */
-  Valid,
-
-  /**
-   * An operation is in Invalid state if its ancestry is complete but at least one of the three conditions
-   * above is false.
-   */
-  Invalid
-}
-
-/**
- * VersionId identifies the version of a DID document. We use the hash of the
- * operation that produces a particular version of a DID document as its versionId.
- * This usage is guaranteed to produce unique VersionId's since the operation contains
- * as one of its properties the previous VersionId. Since the operation hash is
- * just a string we alias VersionId to string.
- *
- * With this usage, the operation hash serves two roles (1) an identifier for an operation
- * (2) an identifier for the DID document produced by the operation. In the code below,
- * we always use VersionId in places where we mean (2) and an OperationHash defined below
- * when we mean (1).
- *
- * Since we identify versions using the operation it is meaningful to apply OperationStatus
- * to versions.
- */
-export type VersionId = string;
-
-/**
- * Alias OperationHash to string - see comment above
- */
-export type OperationHash = string;
+import { WriteOperation, OperationType } from './Operation';
+import { createOperationStore, OperationStore } from './OperationStore';
 
 /**
  * Represents the interface used by other components to process DID operations
@@ -78,11 +20,11 @@ export interface OperationProcessor {
    * operation.
    * @returns the hash of the operation.
    */
-  process (operation: WriteOperation): Promise<string>;
+  process (operation: WriteOperation): Promise<void>;
 
   /**
    * Remove all previously processed operations with transactionNumber
-   * greater than the provided parameter value.
+   * greater or equal to the provided parameter value.
    * The intended use case for this method is to handle rollbacks
    * in the blockchain.
    */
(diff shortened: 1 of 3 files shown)


src/OperationProcessor.ts
```diff
@@ -1,69 +1,11 @@
 import * as Protocol from './Protocol';
+import { Cas } from '../src/Cas';
```
thehenrytsai: './Cas' should do.

src/OperationProcessor.ts
```diff
@@ -76,576 +18,145 @@ export interface OperationProcessor {
...
    * operation.
-   * @returns the hash of the operation.
    */
-  process (operation: Operation): Promise<string>;
+  process (operation: Operation): Promise<void>;
 
   /**
    * Remove all previously processed operations with transactionNumber
-   * greater than the provided parameter value.
+   * greater or equal to the provided parameter value.
```
thehenrytsai: This changes the current contract with the Observer, is it possible to keep it the same? We went over the technical reason why a while back, but we can talk about it again.

src/OperationStore.ts
```diff
@@ -1,89 +1,164 @@
...
+
+  /**
+   * Get an iterator that returns all operations with a given
+   * did ordered by (transactionNumber, operationIndex)
+   * ascending.
+   */
+  get (did: string): Promise<Iterable<Operation>>;
+
+  /**
+   * Delete all operations with at least a given transactionNumber.
```
thehenrytsai: Nit:
Delete all operations with transaction time greater or equal to the given transactionNumber.

(Or just greater if you took the earlier comment).

src/OperationStore.ts
```diff
@@ -1,89 +1,164 @@
...
+   * Store an operation.
+   */
+  put (operation: Operation): Promise<void>;
+
+  /**
+   * Get an iterator that returns all operations with a given
+   * did ordered by (transactionNumber, operationIndex)
+   * ascending.
+   */
+  get (did: string): Promise<Iterable<Operation>>;
```
thehenrytsai: Rename `did` param and in the comment to `didUniqueSuffix` to disambiguate.

src/OperationProcessor.ts
```diff
@@ -76,576 +18,145 @@ export interface OperationProcessor {
...
-    const op = await this.operationStore.lookup(opHash);
+      // If either of these is defined, then we have seen a previous create operation.
+      if (previousOperation || currentDidDocument) {
+        return false;
+      }
 
-    if (this.isInitialVersion(opInfo)) {
-      const protocolVersion = Protocol.getProtocol(op.transactionTime!);
-      return Document.from(op.encodedPayload, this.didMethodName, protocolVersion.hashAlgorithmInMultihashCode);
+      // TODO: Add signature verification for create operation
```
thehenrytsai: Need to add the logic else the regression test I added in master will fail.

After rebase, you should be able to use `Document.from(...)` and `operation.verifySignature(...) to do so.

src/OperationProcessor.ts
```diff
@@ -76,576 +18,145 @@ export interface OperationProcessor {
...
+    const didUniqueSuffix = did.substring(this.didMethodName.length);
 
-    // lastVersion === undefined implies we do not know about the did
-    if (latestVersion === undefined) {
-      return undefined;
+    const didOps = await this.operationStore.get(didUniqueSuffix);
+    for (const operation of didOps) {
+      if (await this.isValid(operation, previousOperation, didDocument)) {
+        didDocument = this.getUpdatedDocument(didDocument, operation);
+        previousOperation = operation;
```
thehenrytsai: Consider adding a short-circuit check here:

// If this is a delete operation, this will be the last valid operation for this DID.
if (operation.type === OperationType.Delete) { break; }

src/OperationProcessor.ts
arvinda: Fixed.

src/OperationProcessor.ts
arvinda: I simply forgot if we had decided on >= or >. Will switch to >.

src/OperationStore.ts
arvinda: Fixed.

src/OperationStore.ts
arvinda: Done.

src/OperationProcessor.ts
arvinda: Added.

src/OperationProcessor.ts
arvinda: Good catch. Done.

src/OperationProcessor.ts
```diff
@@ -75,580 +17,160 @@ export interface OperationProcessor {
...
 
-    // Construct the operation using a CAS lookup
-    const op = await this.operationStore.lookup(opHash);
+    if (operation.type === OperationType.Create) {
 
-    if (this.isInitialVersion(opInfo)) {
-      if (opInfo.status === OperationStatus.Invalid) {
-        return undefined;
+      // If either of these is defined, then we have seen a previous create operation.
+      if (previousOperation || currentDidDocument) {
```
thehenrytsai: What if there are only two operations in chronological order: U1, C1, in this logic C1 is consider invalid, but I should be valid.

src/OperationProcessor.ts
```diff
@@ -75,580 +17,160 @@ export interface OperationProcessor {
...
-      return;
-    }
-
-    for (const descendantHash of waitingDescendants) {
-      const descendantInfo = this.opHashToInfo.get(descendantHash)!;
-      // assert: descendantInfo.status === Unvalidated and descendantInfo.missingAncestor === opHash
-      await this.processInternal(descendantHash, descendantInfo);
-    }
+  // Get the initial DID document for a create operation
+  private getInitialDocument (createOperation: Operation): DidDocument {
```
thehenrytsai: rename to `getOriginalDocument`, as _original document_ is now the official term used in protocol spec.

tests/OperationProcessor.spec.ts
```diff
@@ -132,6 +112,18 @@ function getPermutation (size: number, index: number): Array<number> {
   return permutation;
 }
 
+function getPublicKey (didDocument: DidDocument, keyId: string): DidPublicKey | undefined {
```
thehenrytsai: Great reusable library, let's move it to `Document.ts`.

src/OperationProcessor.ts
```diff
@@ -75,580 +17,160 @@ export interface OperationProcessor {
...
-   */
-  private invalidatePreviouslyValidOperation (opHash: OperationHash) {
-    const opInfo = this.opHashToInfo.get(opHash)!;
-
-    if (opInfo.parent) {
-      this.nextVersion.delete(opInfo.parent);
+  // Update a document with a specified operation.
+  private getUpdatedDocument (didDocument: DidDocument | undefined, operation: Operation): DidDocument {
+    if (operation.type === OperationType.Create) {
+      return this.getInitialDocument(operation);
```
thehenrytsai: Consider avoid reconstructing the _original document_ (calling `getInitialDocument`) twice in case of create, one way is by making `isValid()` return the latest document when it is valid and undefined otherwise.

src/OperationStore.ts
```diff
@@ -1,89 +1,165 @@
...
+
+    this.ensureDidEntriesExist(did);
+    // Append the operation to the operation array for the did ...
+    this.didToOperations.get(did)!.push(operation);
+    // ... which leaves the array unsorted, so we record this fact
+    this.didTouchedSinceLastSort.set(did, true);
+  }
+
+  /**
+   * Implement OperationStore.put
```
thehenrytsai: Typo. `.get`

src/OperationStore.ts
```diff
@@ -1,89 +1,165 @@
...
-      operationBuffer,
-      resolvedTransaction,
-      operationAccessInfo.operationIndex);
-
-    return operation;
+  /**
+   * Gets the DID unique suffix of an operation. For create operation, this is the operation hash;
+   * for others the DID included with the operation can be used to obtain the unique suffix.
+   */
+  private getDidUniqueSuffix (operation: Operation): string {
```
thehenrytsai: Let's move this to `Operation.ts`

src/OperationStore.ts
```diff
@@ -1,89 +1,165 @@
...
+ * A simple in-memory implementation of operation store.
+ */
+class OperationStoreImpl {
+  // Map DID to operations over it stored as an array. The array might be sorted
+  // or unsorted by blockchain time order.
+  private readonly didToOperations: Map<string, Array<Operation>> = new Map();
+
+  // Map DID to a boolean indicating if the operations array for the DID is sorted
+  // or not.
+  private readonly didTouchedSinceLastSort: Map<string, boolean> = new Map();
```
thehenrytsai: Nit: consider `Updated` or `Changed` instead of `touched`?

src/OperationProcessor.ts
arvinda: U1, C1 should leave C1 in a valid state just as you expect. When U1 is being processed we check if previous is defined - in this case U1 would be invalid, so prevoiusOperation would be null, so this if condition would not be taken.

thehenrytsai: You are right.

PR #163, opened 2019-04-08
Fix for Issue 151

arvinda:
Make mongodb index unique to reject duplicate operations

### DIFF
--- a/src/MongoDbOperationStore.ts
+++ b/src/MongoDbOperationStore.ts
@@ -39,7 +39,9 @@ export default class MongoDbOperationStore implements OperationStore {
       this.collection = db.collection(operationCollectionName);
     } else {
       this.collection = await db.createCollection(operationCollectionName);
-      await this.collection.createIndex({ didUniqueSuffix: 1, transactionNumber: 1, operationIndex: 1 });
+      // create an index on didUniqueSuffix, transactionNumber, operationIndex to make get() operations more efficient
+      // this is an unique index, so duplicate inserts are rejected.
+      await this.collection.createIndex({ didUniqueSuffix: 1, transactionNumber: 1, operationIndex: 1 }, { unique: true });
     }
   }
 
@@ -54,7 +56,14 @@ export default class MongoDbOperationStore implements OperationStore {
       batch.insert(mongoOperation);
     }
 
-    await batch.execute();
+    try {
+      await batch.execute();
+    } catch (e) {
+      // Swallow duplicate insert errors (error code 11000); rethrow others
+      if (e.name !== 'BulkWriteError' || e.code !== 11000) {
+        throw e;
+      }
+    }
   }
 
   /**
--- a/src/MongoDbOperationStore.ts
+++ b/src/MongoDbOperationStore.ts
@@ -58,10 +58,10 @@ export default class MongoDbOperationStore implements OperationStore {
 
     try {
       await batch.execute();
-    } catch (e) {
+    } catch (error) {
       // Swallow duplicate insert errors (error code 11000); rethrow others
-      if (e.name !== 'BulkWriteError' || e.code !== 11000) {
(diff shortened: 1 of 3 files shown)


src/MongoDbOperationStore.ts
```diff
@@ -54,7 +56,14 @@ export default class MongoDbOperationStore implements OperationStore {
       batch.insert(mongoOperation);
     }
 
-    await batch.execute();
+    try {
+      await batch.execute();
+    } catch (e) {
```
thehenrytsai: Nit: `error` instead of `e` (group decision).

tests/OperationStore.spec.ts
```diff
@@ -125,6 +125,31 @@ describe('OperationStore', async () => {
     expect(returnedOperationHash).toEqual(updateOperation.getOperationHash());
   });
 
+  it('should ignore duplicate updates', async () => {
```
thehenrytsai: Sanity: aren't all operation store unit tests pointless since they are testing just the mock store?

tests/mocks/MockOperationStore.ts
```diff
@@ -73,6 +73,9 @@ export default class MockOperationStore implements OperationStore {
     // Sort needed if there was a put operation since last sort.
     if (updatedSinceLastSort) {
       didOps.sort(compareOperation);       // in-place sort
+      didOps = didOps.filter((elem, index, self) => {  // remove duplicates
+        return (index === 0) || compareOperation(elem, self[index - 1]) !== 0;
```
thehenrytsai: This looks a bit like hack that is designed to handle the above tests, why not just prevent duplicate in the `putBatch()`?

src/MongoDbOperationStore.ts
arvinda: Done.

tests/OperationStore.spec.ts
arvinda: Yes and no. I use them with minor edits to test MongoDbOperationStore before I checkin any changes. If you have better ideas for testing MongoDbOperationStore, please suggest.

tests/mocks/MockOperationStore.ts
arvinda: I am not convinced this is a hack: (1) since we are anyway sorting before returning from get(), it is efficient to use the sorted operations to deduplicate; (2) the result of deduplication is "materialized" from this point on and the next get() operation would not do the deduplication. 

If we push this logic to putBatch() we would have to push the sorting as well (or use an inefficient method) - but we would be doing more work in the process(batch) -> putBatch() code path and our design with deferred processing is to make this code path very inexpensive.

Does this make sense?

tests/OperationStore.spec.ts
thehenrytsai: I get how you would be testing MongoDbOperationStore with this class, but this would be unknown/hidden to people who does not know that this code exists.

We can add a funciton in the `OperationStore.spec.ts`:

```javascript
function skipTestIfMongoDbNotConfigured (): boolean {
  const mongoDbFound = ...;
  if (!mongoDbFound) {
    pending('MongoDB not configured');
  }
}
```

then call this method in each of the `OperationStore` tests.

Essentially we are introducing a new class of tests that are not unit tests.

tests/mocks/MockOperationStore.ts
thehenrytsai: Sort of. I'm not going to be too picky on this since this is a mock, but my point was that if this was a real store, then I think we wouldn't have stored duplicates, so the comment was trying to align with what a real store would behave.

tests/OperationStore.spec.ts
thehenrytsai: Talked to Arvind offline, this will be addressed in PR #170.

PR #169, opened 2019-04-09
Fix delete operation bugs in operation processor

arvinda:
Fix various bugs in delete processing within operation processor. Add a test to uncover and fix these bugs.


tests/OperationProcessor.spec.ts
```diff
@@ -224,4 +224,27 @@ describe('OperationProcessor', async () => {
     const didDocument = await operationProcessor.resolve(didUniqueSuffix);
     expect(didDocument).toBeUndefined();
   });
+
+  it('should return undefined for deleted did', async () => {
+    const numberOfUpdates = 1;
```
thehenrytsai: Sanity: did you intend to just have a 1 update?

tests/OperationProcessor.spec.ts
```diff
@@ -224,4 +224,27 @@ describe('OperationProcessor', async () => {
...
+    const ops = await createUpdateSequence(didUniqueSuffix, createOp!, cas, numberOfUpdates, privateKey);
+
+    for (let i = 0 ; i < ops.length ; ++i) {
+      await operationProcessor.processBatch([ops[i]]);
+    }
```
thehenrytsai: Sanity: why wouldn't the code use `processBatch(ops)` directly?

src/OperationProcessor.ts
```diff
@@ -67,64 +67,85 @@ export default class OperationProcessor {
    * @param operation The operation to apply against the given current DID Document (if any).
    * @param previousOperation The previously operation applied if any. Used for operation validation.
    * @param currentDidDocument The DID document to apply the given operation against.
-   * @returns undefined if any validation fails; the updated document otherwise.
+   * @returns [isOperationValid, updatedDidDocument]; isOperationValid is a boolean that indicates if the
+   *          operation is valid given the operation context. If the operation is valid, updatedDidDocument
+   *          contains the updated document, Otherwise, it contains currentDidDocument (unchanged).
```
thehenrytsai: Nit: consider calling out that `updatedDidDocument` would be `undefined` in case of valid delete operation.

tests/OperationProcessor.spec.ts
arvinda: No, debugging artifact. Fixed.

tests/OperationProcessor.spec.ts
arvinda: No reason. Fixed it (and also the source this was cut-n-pasted from).

src/OperationProcessor.ts
arvinda: Done.

Issue #165, opened 2019-04-08
DID Document `publicKey` element must contain `controller` field per DID spec


thehenrytsai:
Follow the DID spec such that `controller` will be a required field in the returned DID Document.

mudiali:
The controller field for the publicKey section is supposed to point to the owner did (https://w3c.github.io/did-core/#public-keys). When registering the doc, the client may not actually have the did id to populate in there. 

So it probably makes sense for the node to add that field during resolution. So the protocol can block the clients from putting any value in there. @csuwildcat we probably want to update the section 5.3 (https://w3c.github.io/did-core/#public-keys) with a note saying that the individual methods may introduce an intermediate state where the controller value is not present. It would be similar to the NOTE section 5.2 which is:

> DID method specifications MAY create intermediate representations of a DID document that do not contain the id key, such as when a DID resolver is performing resolution. However, the fully resolved DID document MUST contain a valid id property.

csuwildcat:
Why do we think a spec change is required here? The DID spec does not mandate any specific properties, values, or structures on the internal operation data of Methods, only the resolved documents.

mudiali:
@csuwildcat I would think that the update is needed for the same reason that we have an existing note in section 5.2.

PR #170, opened 2019-04-09
Add additional operation processor tests

arvinda:


lib/OperationProcessor.ts
```diff
@@ -82,10 +82,7 @@ export default class OperationProcessor {
         return [false, currentDidDocument];
       }
 
-      const originalDidDocument = this.getOriginalDocument(operation);
-      if (originalDidDocument === undefined) {
```
thehenrytsai: Sanity: did you mean to delete the if statement? If so, why?

tests/OperationProcessor.spec.ts
```diff
@@ -233,11 +250,273 @@ describe('OperationProcessor', async () => {
...
+    expect(didDocument).toBeUndefined();
+  });
+
+  it('should return undefined for deleted did', async () => {
+    const numberOfUpdates = 10;
+    const ops = await createUpdateSequence(didUniqueSuffix, createOp!, cas, numberOfUpdates, privateKey);
+
+    for (let i = 0 ; i < ops.length ; ++i) {
+      await operationProcessor.processBatch([ops[i]]);
+    }
```
thehenrytsai: Use just `processBatch(ops);`, fix for all occurrences.

tests/OperationProcessor.spec.ts
```diff
@@ -233,11 +250,273 @@ describe('OperationProcessor', async () => {
...
+
+    const didDocument = await operationProcessor.resolve(didUniqueSuffix);
+    expect(didDocument).toBeDefined();
+    const publicKey2 = Document.getPublicKey(didDocument!, 'key2');
+    expect(publicKey2).toBeDefined();
+    expect(publicKey2!.owner).toBeDefined();
```
thehenrytsai: Just wondering if we can avoid using `owner` for new tests as it is no longer a valid property, it is being superseded by `controller` and there is a pending bug to fix existing code (#165).

tests/OperationProcessor.spec.ts
```diff
@@ -233,11 +250,273 @@ describe('OperationProcessor', async () => {
     expect(publicKey2!.owner).toBeDefined();
     expect(publicKey2!.owner!).toEqual('did:sidetree:updateid' + (numberOfUpdates - 1));
 
-    const deleteOperationBuffer = await OperationGenerator.generateDeleteOperation(didUniqueSuffix, '#key1', privateKey);
```
thehenrytsai: Design sanity: how can the test without `await operationProcessor.processBatch([createOp!]);` being called? Are each subsequent test dependent on the previous test? I hope not.

tests/OperationProcessor.spec.ts
```diff
@@ -233,11 +250,273 @@ describe('OperationProcessor', async () => {
...
+    expect(publicKey2).toBeDefined();
+    expect(publicKey2!.owner).toBeUndefined(); // if update above went through, the owner would be defined
+  });
+
+  it('should pick earlier of two conflicting updates', async () => {
+    await operationProcessor.processBatch([createOp!]);
+
+    const update1Payload = {
+      didUniqueSuffix,
+      operationNumber: 1,
```
thehenrytsai: Unless it is going to break tests, consider not having adding dependencies to `operationNumber` since we agreed to remove it (issue #172).

tests/OperationProcessor.spec.ts
```diff
@@ -233,11 +250,273 @@ describe('OperationProcessor', async () => {
...
+          owner: 'did:sidetree:updateid2',
+          publicKeyPem: process.hrtime() // Some dummy value that's not used.
+        }
+      }]
+    };
+
+    const updateOperation1Buffer = await OperationGenerator.generateUpdateOperationBuffer(update1Payload, '#key1', privateKey);
+    const updateOperation1 = await addBatchFileOfOneOperationToCas(updateOperation1Buffer, cas, 1, 1, 0);
+    await operationProcessor.processBatch([updateOperation1]);
+
```
thehenrytsai: For extra coverage, do you think we can flip the order or processing and process operation 2 first?

tests/OperationStore.spec.ts
```diff
@@ -64,21 +71,69 @@ async function constructAnchoredUpdateOperation (
     }]
   };
 
-  const updateOperationBuffer = await OperationGenerator.generateUpdateOperation(updatePayload, '#key1', privateKey);
+  const updateOperationBuffer = await OperationGenerator.generateUpdateOperationBuffer(updatePayload, '#key1', privateKey);
   return constructAnchoredOperation(updateOperationBuffer, transactionNumber, transactionTime, operationIndex);
 }
 
+const databaseName = 'sidetree';
```
thehenrytsai: Consider expose these constants in `MongoDbOperationStore` and use them here.

tests/OperationStore.spec.ts
```diff
@@ -64,21 +71,69 @@ async function constructAnchoredUpdateOperation (
...
+    return false;
+  }
+  return true;
+}
+
+/**
+ * Clear a mongo collection - used to remove test state.
+ */
+async function clearMongoCollection () {
+  const client = await MongoClient.connect('mongodb://localhost:27017');
```
thehenrytsai: Pass in server URL like line 84.

tests/OperationStore.spec.ts
```diff
@@ -64,21 +71,69 @@ async function constructAnchoredUpdateOperation (
...
+  const db = client.db(databaseName);
+  const collections = await db.collections();
+  const collectionNames = collections.map(collection => collection.collectionName);
+
+  // If the operation collection exists, use it; else create it then use it.
```
thehenrytsai: Sanity: I don't see the else portion? Also are you able to use the collection if it is dropped?

tests/OperationStore.spec.ts
```diff
@@ -64,21 +71,69 @@ async function constructAnchoredUpdateOperation (
...
+  const operationStore = new MongoDbOperationStore(operationStoreUri);
+  await operationStore.initialize();
+  return operationStore;
+}
+
 describe('OperationStore', async () => {
   const versionsOfProtocolParameters = require('../json/protocol-parameters-test.json');
   ProtocolParameters.initialize(versionsOfProtocolParameters);
 
+  // Change to OperationStoreType.Mongo to test Mongo-based store
```
thehenrytsai: This line no longer applicable?

tests/OperationStore.spec.ts
```diff
@@ -64,21 +71,69 @@ async function constructAnchoredUpdateOperation (
...
+  const collectionNames = collections.map(collection => collection.collectionName);
+
+  // If the operation collection exists, use it; else create it then use it.
+  if (collectionNames.includes(operationCollectionName)) {
+    const collection = db.collection(operationCollectionName);
+    await collection.drop();
+  }
+}
+
+async function getOperationStore (operationStoreUri: string): Promise<OperationStore> {
```
thehenrytsai: Rename to `createOperationStore`

tests/OperationStore.spec.ts
```diff
@@ -212,4 +267,177 @@ describe('OperationStore', async () => {
...
+      expect(returnedOperation.getOperationHash()).toEqual(batch[i].getOperationHash());
+    }
+
+    // Create another instance of the operation store
+    operationStore = await getOperationStore(config.operationStoreUri);
+
+    // Check if we have all the previously put operations
+    returnedOperations = Array.from(await operationStore.get(didUniqueSuffix));
+
+    expect(returnedOperations.length).toEqual(batchSize);
```
thehenrytsai: It appears that most of the verification code and creation code are identical between tests, can we refactor them to reusable methods?

tests/OperationStore.spec.ts
```diff
@@ -212,4 +267,177 @@ describe('OperationStore', async () => {
...
+      expect(returnedOperation.operationIndex).toBeDefined();
+      expect(returnedOperation.operationIndex!).toEqual(0);
+      expect(returnedOperation.transactionTime).toBeDefined();
+      expect(returnedOperation.transactionTime!).toEqual(i);
+      expect(returnedOperation.didUniqueSuffix).toEqual(didUniqueSuffix);
+      expect(returnedOperation.getOperationHash()).toEqual(batch[i].getOperationHash());
+    }
+  });
+
+  it('should get all operations in transaction time order', async () => {
```
thehenrytsai: Sanity: is this not covered by previous tests? If so, consider removing.

lib/OperationProcessor.ts
arvinda: Yes, I believe there is no way we would ever get undefined at this point because the same checks are applied when the operation is created.

arvinda: Please check this claim...

tests/OperationProcessor.spec.ts
```diff
@@ -65,7 +65,7 @@ async function createUpdateSequence (
       }]
```
thehenrytsai: Rename this file to `MongoDbOperationStore.spec.ts`

tests/OperationProcessor.spec.ts
arvinda: Fixed. Please note that many of the remaining non-batch ones exist since we want to process the operations in a particular order (e.g., reverse) or a sub array. I can always remove them by generating a new array (e.g,. using Array.splice) - let me know if you have a strong preference for this...

tests/OperationStore.spec.ts
```diff
@@ -64,21 +71,69 @@ async function constructAnchoredUpdateOperation (
...
+  }
+}
+
+async function getOperationStore (operationStoreUri: string): Promise<OperationStore> {
+  const operationStore = new MongoDbOperationStore(operationStoreUri);
+  await operationStore.initialize();
+  return operationStore;
+}
+
 describe('OperationStore', async () => {
```
thehenrytsai: Rename this file to `MongoDbOperationStore.spec.ts` and describe as `MongoDbOperationStore`.

tests/OperationProcessor.spec.ts
arvinda: Removed all operationNumber in the tests.

tests/OperationStore.spec.ts
arvinda: Sure, I have made the change. However are we convinced that we are only going to have MongoDbOperation store going forward? The tests are quite general and will be useful for any new operation store that we might have in the future.

tests/OperationProcessor.spec.ts
arvinda: Sure, makes sense. Took the suggestion.

tests/OperationStore.spec.ts
arvinda: Made these public readonly static members of MongoDbOperationStore.

tests/OperationStore.spec.ts
arvinda: Done.

tests/OperationStore.spec.ts
arvinda: Sorry, cut-n-paste legacy. the comment is invalid. Fixed it and renamed the function.

tests/OperationStore.spec.ts
arvinda: Done.

tests/OperationStore.spec.ts
arvinda: Good catch. Removed.

tests/OperationStore.spec.ts
arvinda: i don't think so - the operation insert order is different.

tests/OperationProcessor.spec.ts
arvinda: Switched update patches to use serviceendpoint/instance as we discussed offline.

PR #179, opened 2019-04-18
Observer - Added transaction ersistence using MongoDB

thehenrytsai:


lib/MongoDbTransactionStore.ts
```diff
@@ -0,0 +1,235 @@
...
+  retryAttempts: number;
+  nextRetryTime: number;
+}
+
+/**
+ * Implementation of TransactionStore that stores the transaction data in a MongoDB database.
+ */
+export default class MongoDbTransactionStore implements TransactionStore {
+  /** Default database name used if not specified in constructor. */
+  public static readonly defaultDatabaseName: string = 'sidetree';
```
arvinda: I am worried that this could lead to subtle race condition bugs: In MongoDb, when a non-existent database (here, sidetree) is referenced the database is auto-created. What happens when TransactionStore and OperationStore reference and thereby set off database creation around the same time? If you can find documentation that Mongo correctly takes locks and handles this situation then we should be fine.

thehenrytsai: Since we await initialization of `TransactionStore` and `OperationStore` we should be fine. let me know if you disagree.

PR #180, opened 2019-04-18
Integrate with cosmosdb/mongo api

arvinda:


lib/MongoDbOperationStore.ts
```diff
@@ -4,11 +4,13 @@ import { OperationStore } from './OperationStore';
 
 /**
  * Sidetree operation stored in MongoDb.
+ * Note: we use the shorter property name "opIndex" instead of "operatioIndex" due to a constraint imposed by CosmosDB/MongoDB:
```
thehenrytsai: Consider mentioning the constraint is specifically index key length.

arvinda: Done.

PR #190, opened 2019-04-29
Rename `putBatch` to `put`, and `processBatch` to `process` 

### LINKED ISSUE #177: Rename `putBatch` to `put`, and `processBatch` to `process`
The concern here is that the term "batch" here is not the same as the "operation batch" or "batch file" being used.

Atul9:
closes #177


Atul9:
@thehenrytsai done. I have updated the PR.

PR #193, opened 2019-05-08
Migrated `sidetree-ipfs` lib to `sidetree-core`.

thehenrytsai:


lib/index.ts
```diff
@@ -25,3 +25,12 @@ export {
   ISidetreeBitcoinConfig,
   SidetreeBitcoinService
 };
+
+// IPFS service exports.
```
rado0x54: Not sure if I like this "Super-Index". Sidetree-Bitcoin would not be interested in these dependencies. I think it would still be cleaner if the code would import individual sub-modules in `lib`.

src/bitcoin-config.json
```diff
@@ -1,4 +1,4 @@
-﻿{
+{
```
rado0x54: Thanks for removing those. 👍

lib/index.ts
thehenrytsai: The hope here is that an external repo using this lib will be able to know what are the explicitly exported classes without needing to sift through the folders and being overwhelmed by classes, that are irrelevant (used internally). If you feel strongly about this, feel free to file an issue, and we can discuss this further.

PR #198, opened 2019-05-15
Bitcoin Service overhaul

logangirvin:
* Revised Bitcoin service to talk directly to the underlying Bitcoin miner as a library.
* Added utilities for SidetreeError and StreamReader
* Added unit tests for the Bitcoin service (93.33% branch)
* Revised instructions to starting a Bitcoin miner (Bcoin).
    * Added a shell script to install it locally


lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -0,0 +1,581 @@
...
+      } catch (error) {
+        if (error instanceof FetchError) {
+          if (retryCount >= this.maxRetries) {
+            console.debug('Max retries reached. Request failed.');
+            throw error;
+          }
+          switch (error.type) {
+            case 'request-timeout':
+              console.debug(`Request timeout (${retryCount})`);
+              await this.waitFor(Math.round(Math.random() * this.defaultTimeout + timeout));
```
thehenrytsai: I don't get the intention behind `Math.random() * this.defaultTimeout`, it progressively less significant as retry increases, so can we remove it?

src/bitcoin.ts
```diff
@@ -75,36 +72,39 @@ app.use((ctx, _next) => {
...
 /**
  * Sets the koa response according to the Sidetree response object given.
  * @param response Response object fetched from request handler.
  * @param koaResponse Koa Response object to be filled
  * @param contentType Content type to be set for response, defaults to application/json
  */
-const setKoaResponse = (response: ISidetreeResponse, koaResponse: Koa.Response, contentType?: string) => {
-  koaResponse.status = SidetreeResponse.toHttpStatus(response.status);
+const setKoaResponse = (response: any, koaResponse: Koa.Response, contentType?: string) => {
+  koaResponse.status = SidetreeResponse.toHttpStatus(200);
```
thehenrytsai:
```suggestion
  koaResponse.status = 200;
```
Sanity: did you mean to do this?

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -0,0 +1,581 @@
...
+   * @param since A transaction number
+   * @param hash The associated transaction time hash
+   * @returns Transactions since that blocktime
+   */
+  public async transactions (since?: number, hash?: string): Promise<{
+    moreTransactions: boolean,
+    transactions: ITransaction[]
+  }> {
+    if (since && !hash) {
+      throw new SidetreeError(httpStatus.BAD_REQUEST);
```
thehenrytsai: It is my intention to not have direct "http" reference in the lib, consider using the internal `ResponseStatus` enum (which I acknowledge that I want to make it string enums after our discussion yesterday.)

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -0,0 +1,581 @@
...
+   * Fetches Sidetree transactions in chronological order from since or genesis.
+   * @param since A transaction number
+   * @param hash The associated transaction time hash
+   * @returns Transactions since that blocktime
+   */
+  public async transactions (since?: number, hash?: string): Promise<{
+    moreTransactions: boolean,
+    transactions: ITransaction[]
+  }> {
+    if (since && !hash) {
```
```suggestion
    if ((since && !hash) ||
        (!since && hash)) {
```

lib/bitcoin/BitcoinProcessor.ts
logangirvin: I had put it in when there was greater parallelization, which is still an option in this code. 
In the event that multiple calls are made at the same time and the bitcoin peer becomes overwhelmed, the random would begin to separate the request retries. Its only one timeout interval because I wanted to retain some amount of guaranteed timeout between retries, but should still be significant enough to work over multiple retry attempts.

lib/bitcoin/BitcoinProcessor.ts
logangirvin: For sure!
I was more or less just switching between the existing implementation, the implementation doc, and how Koa would need the error to be formatted to return correctly. As a true library, it should just be an enum so I'll make that change.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -0,0 +1,590 @@
...
+    if ((since && !hash) ||
+        (!since && hash)) {
+      throw new SidetreeError(httpStatus.BAD_REQUEST);
+    } else if (since && hash) {
+      if (!await this.verifyBlock(TransactionNumber.getBlockNumber(since), hash)) {
+        console.info('Requested transactions hash mismatched blockchain');
+        throw new SidetreeError(httpStatus.BAD_REQUEST, 'invalid_transaction_number_or_time_hash');
+      }
+    } else {
+      since = TransactionNumber.construct(this.genesisBlockNumber, 0);
```
thehenrytsai: Probably corner case, but technically incorrect since this would exclude the transactions contained in the genesis block. I'll make an update.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -0,0 +1,590 @@
...
+    };
+  }
+
+  /**
+   * Given a list of Sidetree transactions, returns the first transaction in the list that is valid.
+   * @param transactions List of transactions to check
+   * @returns The first valid transaction, or undefined if none are valid
+   */
+  public async firstValidTransaction (transactions: ITransaction[]): Promise<ITransaction | undefined> {
+    // sort so lower transaction numbers come first
```
thehenrytsai: No need to sort, literally return the first valid transaction in the array.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -0,0 +1,590 @@
...
+    const response = JSON.parse(responseData);
+
+    return response.success;
+  }
+
+  /**
+   * Will process transactions every interval seconds.
+   * @param interval Number of seconds between each query
+   */
+  private periodicPoll (interval: number = this.pollPeriod) {
```
thehenrytsai: Since this is private instanced method, consider not taking in `interval` argument and use `this.pollPeriod` directly internally.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -0,0 +1,590 @@
...
+    if (this.pollTimeoutId) {
+      clearTimeout(this.pollTimeoutId);
+    }
+    this.processTransactions(this.lastSeenBlock).then((syncedTo) => {
+      this.lastSeenBlock = syncedTo;
+      this.pollTimeoutId = setTimeout(this.periodicPoll.bind(this), 1000 * interval, interval);
```
```suggestion
      this.pollTimeout = setTimeout(() => this.periodicPoll(), 1000 * this.pollPeriod,);
```
Consider non "bind" notation. Though you can't just take the suggested line at face value because it has assumption.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -0,0 +1,590 @@
...
+    }).catch((error) => {
+      console.error(error);
```
thehenrytsai: Consider moving `setTimeout()` call to finally so next poll always gets scheduled regardless of processing outcome.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -0,0 +1,590 @@
...
+
+  /**
+   * Will process transactions every interval seconds.
+   * @param interval Number of seconds between each query
+   */
+  private periodicPoll (interval: number = this.pollPeriod) {
+    if (this.pollTimeoutId) {
+      clearTimeout(this.pollTimeoutId);
+    }
+    this.processTransactions(this.lastSeenBlock).then((syncedTo) => {
```
thehenrytsai: Nit: Consider using async/await style over .then.catch chaining.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -0,0 +1,590 @@
...
+      const exponentiallySpacedTransactions = await this.transactionStore.getExponentiallySpacedTransactions();
+
+      const firstValidTransaction = await this.firstValidTransaction(exponentiallySpacedTransactions);
+
+      let revertToTransactionNumber: number;
+
+      if (firstValidTransaction) {
+        revertToTransactionNumber = firstValidTransaction.transactionNumber;
+      } else {
+        const lowestHeight = exponentiallySpacedTransactions.reduce((height: number, transaction: ITransaction): number => {
```
thehenrytsai: No need for the reduce call to sort, just get the last transaction from `exponentiallySpacedTransactions`

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -0,0 +1,590 @@
...
+        // check for returned data for sidetree prefix
+        const hexDataMatches = script.asm.match(/\s*OP_RETURN ([0-9a-fA-F]+)$/);
+        if (!hexDataMatches || hexDataMatches.length === 0) {
+          continue;
+        }
+        const data = Buffer.from(hexDataMatches[1], 'hex').toString();
+        if (data.startsWith(this.sidetreePrefix)) {
+          // we have found a sidetree transaction
+          const sidetreeTransaction: ITransaction = {
+            transactionNumber: TransactionNumber.construct(block, anchorFilePosition),
```
thehenrytsai: I like your idea of using `transactionIndex` instead, you reckon we can?

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -0,0 +1,590 @@
...
+   * Gets all unspent coins of a given address
+   * @param address Bitcoin address to get coins for
+   */
+  private async getUnspentCoins (address: Address): Promise<Transaction.UnspentOutput[]> {
+    const addressToSearch = address.toString();
+    console.info(`Getting unspent coins for ${addressToSearch}`);
+    const requestPath = `/coin/address/${addressToSearch}`;
+
+    const fullPath = new URL(requestPath, this.bitcoinExtensionUri);
+    const response = await this.fetchWithRetry(fullPath.toString());
```
thehenrytsai: Any reason not to use the `rpcCall` method? Same goes to all 3 other occurrences.

package.json
```diff
@@ -10,6 +10,7 @@
     "@decentralized-identity/did-auth-jose": "0.1.5",
     "@decentralized-identity/did-common-typescript": "0.1.2",
     "base64url": "3.0.1",
+    "bitcore-lib": "^8.3.0",
```
thehenrytsai: Disucssion: do you think it is good to lock version here?

lib/bitcoin/BitcoinProcessor.ts
logangirvin: the `this.periodicPoll()` would lose reference to `this` when called by setTimeout (and theres a trailing comma).

lib/bitcoin/BitcoinProcessor.ts
logangirvin: This is the same behavior as the old code. This particular "index" is a counter incremented within this conditional block.

lib/bitcoin/BitcoinProcessor.ts
logangirvin: rpcCall assumes this is a `POST` request, which would fail these particular calls as they are `GET` requests to REST endpoints on the node.

package.json
logangirvin: I just did `npm install ...` so this was not an intentional decision. In my opinion, locking versions should be left to the `package-lock.json`, but that has a tendency to clash in git commits and just be overwritten, so if you'd like the redundancy, sure?

lib/bitcoin/BitcoinProcessor.ts
logangirvin: Talked with Henry and will be keeping the `interval` parameter, for now.

PR #236, opened 2019-06-08
Fixed Bitcoin service not returning invalid hash error code when blockchain forked + other fixes

thehenrytsai:
1. Bitcoin service - Fixed service not returning invalid hash error code when blockchain forked + regression test.
1. Core service - Added better handling of bad CAS fetch requests.
1. Tests - 100% code coverage on Cas.ts
1. Renamed clashed `SidetreeError` classes.
1. Consolidated duplicated enums + using string-based enums going forward.


lib/core/Cas.ts
```diff
@@ -38,7 +38,8 @@ export enum FetchResultCode {
   Success = 'success',
```
logangirvin: Should this be pulled out into its own file like how RequestError has a code enum and Response has response status codes?

src/bitcoin.ts
```diff
@@ -71,40 +105,35 @@ app.use((ctx, _next) => {
...
 // initialize the blockchain service and kick-off background tasks
+let server: any;
 let blockchainService: SidetreeBitcoinProcessor;
 try {
   blockchainService = new SidetreeBitcoinProcessor(config);
 
-  blockchainService.initialize()
+  // SIDETREE_TEST_MODE enables unit testing of this file by bypassing blockchain service initialization.
+  if (process.env.SIDETREE_TEST_MODE === 'true') {
+    server = app.listen(port);
```
logangirvin: shouldn't you still initialize() ?

tests/json/bitcoin-config-test.json
```diff
@@ -0,0 +1,11 @@
+{
+  "bitcoinPeerUri": "unused",
+  "bitcoinWalletImportString": "cVneRge89Va3mrxv7dehTYTknbgLVan8rY5hkbXUwdwXtZrrueGY",
```
logangirvin: And all of the scanners running over github for keys go wild... I'd generally generate a new key during testing, but this is fine 🔥🔥🔥

lib/core/Cas.ts
thehenrytsai: Done.

src/bitcoin.ts
thehenrytsai: No, purposely not going through the initialization.

PR #244, opened 2019-06-18
Switching all bcoin specific api calls to generic bitcoin core rpc calls

logangirvin:
* Removed all bcoin api calls
* Added bitcoin rpc calls for wallets
* Changed documentation to recommend the Satoshi client


lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -77,6 +80,7 @@ export default class BitcoinProcessor {
 
   public constructor (config: IBitcoinConfig) {
     this.bitcoinPeerUri = config.bitcoinPeerUri;
+    this.bitcoinAuthorization = Buffer.from(`${config.bitcoinRpcUsername}:${config.bitcoinRpcPassword}`).toString('base64');
```
AlexITC: It would be nice if the authentication is optional instead.

logangirvin: While I have made it optional, the documentation will treat RPC authorization as required as the satoshi client does [require an rpc password](https://bitcoin.org/en/developer-reference#remote-procedure-calls-rpcs).

tests/bitcoin/BitcoinProcessor.spec.ts
```diff
@@ -23,7 +23,9 @@ function randomBlock (above: number = 0): IBlockInfo {
 describe('BitcoinProcessor', () => {
 
   const testConfig: IBitcoinConfig = {
-    bitcoinPeerUri: 'http://localhost:18331',
+    bitcoinPeerUri: 'http://localhost:18332',
+    bitcoinRpcUsername: 'admin',
```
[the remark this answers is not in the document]
logangirvin: Need @thehenrytsai since this reaches into overarching config details, but theres no strict requirement on location and the port will change between testnet and mainnet. For the time being I've set the src/bitcoin-config to a local testnet node.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -120,6 +126,23 @@ export default class BitcoinProcessor {
...
+      console.debug(`Configuring bitcoin peer to Watch address ${this.privateKey.toAddress()}. This can take up to 10 minutes.`);
+      const request = {
+        method: 'importpubkey',
+        params: [
+          this.privateKey.toPublicKey().toBuffer().toString('hex'),
+          'sidetree',
+          true
+        ]
+      };
+      await this.rpcCall(request, undefined, false);
```
thehenrytsai: Sanity: so this RPC call blocks up to 10 minutes before returning without timing out? Is there a way to poll (thus print) the status?

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -120,6 +126,23 @@ export default class BitcoinProcessor {
   public async initialize () {
     console.debug('Initializing TransactionStore');
     await this.transactionStore.initialize();
+    const address = this.privateKey.toAddress();
+    console.debug(`Checking if bitcoin contains a wallet for ${address}`);
+    if (!await this.walletExists(address.toString())) {
+      console.debug(`Configuring bitcoin peer to Watch address ${this.privateKey.toAddress()}. This can take up to 10 minutes.`);
```
```suggestion
      console.debug(`Configuring bitcoin peer to watch address ${address}. This can take up to 10 minutes.`);
```

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -262,6 +288,22 @@ export default class BitcoinProcessor {
     console.info(`Successfully submitted transaction ${transaction.id}`);
   }
 
+  /**
+   * Gets the best hash for a given block height
```
```suggestion
   * Gets the block hash for a given block height
```

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -552,21 +532,49 @@ export default class BitcoinProcessor {
...
+    console.info(`Checking if bitcoin wallet for ${address} exists`);
+    const request = {
+      method: 'getaddressinfo',
+      params: [
+        address
+      ]
+    };
+
+    const response = await this.rpcCall(request);
+    return response.labels.length > 0 || response.iswatchonly;
```
thehenrytsai: Sanity: under what situation would the response fall on either side of the condition?

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -590,16 +598,20 @@ export default class BitcoinProcessor {
...
-  private async fetchWithRetry (uri: string, requestParameters?: RequestInit | undefined): Promise<Response> {
+  private async fetchWithRetry (uri: string, requestParameters?: RequestInit | undefined, setTimeout: boolean = true): Promise<Response> {
     let retryCount = 0;
     let timeout: number;
     do {
       timeout = this.requestTimeout * 2 ** retryCount;
-      const params = Object.assign({}, requestParameters, {
-        timeout
-      });
+      let params = Object.assign({}, requestParameters);
```
thehenrytsai: Would below work?
```suggestion
      let params = requestParameters;
```

lib/bitcoin/README.md
```diff
@@ -66,14 +47,16 @@ Linux users:
...
+If you installed Bitcoin locally, `localhost` will do.
 
-In the following configuration `bcoin.local` refers to your IP address, DNS address, or `localhost`.
+In the following configuration `bitcoin.local` refers to your IP address, DNS address, or `localhost` of the Bitcoin peer.
 
 Edit your bitcoin-config.json
 ```json
 {
-  "bitcoinPeerUri": "[FILL THIS IN!]",
+  "bitcoinPeerUri": "bitcoin.local:18332",
```
thehenrytsai: Sanity: is this kubernetes cluster artifacts?
```suggestion
  "bitcoinPeerUri": "http://localhost:18332",
```

lib/bitcoin/README.md
```diff
@@ -66,14 +47,16 @@ Linux users:
...
+  "bitcoinRpcUsername": "{{YOUR USERNAME}}",
```
thehenrytsai: Nit: keep to the same convention, either use [FILL THIS IN!], or replace line 60 with something like {{YOUR WIF STRING}}

src/bitcoin-config.json
```diff
@@ -1,5 +1,7 @@
 {
-  "bitcoinPeerUri": "[FILL THIS IN!]",
+  "bitcoinPeerUri": "http://localhost:18332",
+  "bitcoinRpcUsername": "",
```
thehenrytsai: Nit: Consider adding [FILL THIS IN!] for consistency.

lib/bitcoin/BitcoinProcessor.ts
logangirvin: So the actual command is quick if you don't care about history, but I gave the "rescan" flag, so it hangs and doesn't return until that is complete, which takes about 10 minutes. Theres no progress report that I found aside from whats printed on the bitcoin daemon.

lib/bitcoin/BitcoinProcessor.ts
logangirvin: This was my attempt when comparing two responses from the bitcoin daemon for address info, one of which is being watched while the other is not.
In the bitcoin service, I label the address "sidetree" which makes for an easy check on labels, although really any 'label' would indicate its being watched.
I'm sure theres a number of indicators for if the address is being watched from the addresses, but these were the two most obvious from what this code does.

lib/bitcoin/README.md
logangirvin: No. Per the line above, I did not want to assume where your bitcoin peer is. "bitcoin.local" would never resolve unless you had explicitly set your network domain to "local" or wrote in your host file, which makes it a safe domain to use for examples.

lib/bitcoin/README.md
logangirvin: Would you have a wallet at this point? Further on I mention that if you don't it'll complain and generate a WIF for you to use in this config value.

src/bitcoin-config.json
logangirvin: This was intentional. I had made RPC username and password optional, so leaving them blank will evaluate to false in javascript-type-conversion-anything-goes-land and thus RPC calls would be made without authorization headers.

Issue #262, opened 2019-07-11
Replace JSON patch with strict update payload schema


thehenrytsai:

OR13:
Why do we need to replace the patch, why not apply some validation to it instead?

I included some comments here:

https://github.com/decentralized-identity/sidetree/issues/229

csuwildcat:
I would venture a guess and say that we want to validate before the merge of a delta, meaning we need to strictly control and evaluate delta contents. This might be hard to do in a super specific and efficient way without some custom code. Henry, does that sound about right?

thehenrytsai:
@csuwildcat yes that's correct. I am now in the process of adding validation logic to the update payload at the observing time, upon until now it has just been "return true" regardless of what is in the patch. Validating a free-form JSON patch to meet the DID spec would be quite difficult if not at least tedious and error-prone IMO. Having a stricter schema make this validation logic a lot simpler (at least more readable).

@OR13's request on labeling keys in #229 is still valid and related, and will be looked at as its own issue.

I will upload a link to the proposed update operation API/schema change.

thehenrytsai:
https://github.com/decentralized-identity/sidetree/blob/henrytsai/120/docs/protocol.md#update-payload-schema

Note that we will need a better way at surfacing/documenting the allowed key algorithms and and supported format combination(s).

For now I am just listing the minimal set as a start.

Same goes with `serviceType`, we can add to the allowed list then what is currently allowed.

thehenrytsai:
Implemented in 2bca504dae9a36de6dc7330bc5baa1a2c628050a.

Issue #172, opened 2019-04-11
Remove concept of 'operation number' in update operation payload


thehenrytsai:

OR13:
Adding a bit of color here, I think the reason this is OK is the `previousOperationHash` makes `operationNumber` redundant.

`operationNumber` is only referenced a few places, should be an easy fix: https://github.com/decentralized-identity/sidetree/search?q=operationNumber&unscoped_q=operationNumber

thehenrytsai:
Addressed in 0bdc44367eef752c844dbc6428eefc254f905c7c.

Issue #120, opened 2019-02-08
Add schema validation for update request payload.


thehenrytsai:
Validation for create is already done.

thehenrytsai:
Implemented in 3419fc3a43d6f0c6a7db0d3aa11a1bc4381daf1e.

Issue #265, opened 2019-07-15
Add an API exposing the version of running Sidetree protocol


thehenrytsai:

mudiali:
PR #284 fixes this.

sambarnes:
@thehenrytsai just a reminder that this can be closed. Was looking for issues to contribute to and saw this

PR #267, opened 2019-07-18
 Issue #262 - Replaced JSON patch with strict update payload schema + other fixes

thehenrytsai:
- Issue #262 - Replaced JSON patch with strict update payload schema.
- Issue #262 - Documented update patch schema.
- Issue #172 - Removed concept of 'operation number' in update operation payload.
- Issue #120 - Added schema validation for update operation payload.
- Core service - Throwing more specific errors when operation fails validation.
- Core service - Added error code for malformed bad requests.
 - Tests - Fixed request generator used for load testing.


lib/common/ErrorCode.ts
```diff
@@ -18,7 +18,29 @@ enum ErrorCode {
...
+  OperationHeaderMissingKid = 'operation_header_missing_kid',
+  OperationHeaderMissingOrIncorrectAlg = 'operation_header_missing_or_incorrect_alg',
+  OperationHeaderMissingOrIncorrectOperation = 'operation_header_missing_or_incorrect_operation',
+  OperationMissingOrIncorrectPayload = 'operation_missing_or_incorrect_payload',
+  OperationMissingOrIncorrectSignature = 'operation_missing_or_incorrect_signature',
+  OperationUpdatePayloadDidUniqueSuffixInvalid = 'operation_update_payload_did_unique_suffix_invalid',
+  OperationUpdatePayloadPreviousOperationHashInvalid = 'operation_update_payload_previous_operation_hash_invalid',
+  OperationUpdatePayloadMissingOrInvalidDidUniqueSuffixType = 'operation_update_payload_missing_or_invalid_did_unique_suffix_type',
+  OperationUpdatePayloadMissingOrInvalidPreviousOperationHashType = 'operation_update_payload_missing_or_invalid_previous_operation_hash_type',
+  OperationUpdatePayloadMissingOrUnknownProperty = 'operation_update_missing_or_unknown_property',
```
gjgd: Shouldn't this be: `operation_update_payload_missing_or_unknown_property`

lib/common/ErrorCode.ts
```diff
@@ -18,7 +18,29 @@ enum ErrorCode {
...
+  OperationUpdatePatchesNotArray = 'operation_update_patches_not_array',
+  OperationUpdatePatchMissingOrUnknownAction = 'operation_update_patch_missing_or_unknown_action',
+  OperationUpdatePatchMissingOrUnknownProperty = 'operation_update_patch_missing_or_unknown_property',
+  OperationUpdatePatchPublicKeyHexMissingOrIncorrect = 'operation_update_patch_public_key_hex_missing_or_incorrect',
+  OperationUpdatePatchPublicKeyIdNotString = 'operation_update_path_public_key_id_not_string',
```
gjgd: operation_update_patch_public_key_id_not_string *

lib/core/Did.ts
```diff
@@ -25,11 +25,22 @@ export default class Did {
...
+    if (!did.startsWith('did:')) {
+      return false;
+    }
+
+    const uniqueSuffixWithMethodName = did.substring(4); // e.g. 'sidetree:abc'
+    const indexOfColonChar = uniqueSuffixWithMethodName.indexOf(':');
+
+    // ':' must exists and not be the first or last character.
+    if (indexOfColonChar <= 0 ||
+        indexOfColonChar === (uniqueSuffixWithMethodName.length - 1)) {
```
gjgd: According to the spec, `did:example:` is a valid did, but it would fail this test because ':' is tha last character

https://w3c-ccg.github.io/did-spec/#generic-did-syntax

tests/core/Did.spec.ts
```diff
@@ -0,0 +1,8 @@
+import Did from '../../lib/core/Did';
+
+describe('DID', async () => {
+  it('isDid() should return false if DID method is not in the DID.', async () => {
+    const isDid = Did.isDid('did:abcdefg');
+    expect(isDid).toBeFalsy();
+  });
```
gjgd: could add a test for a string that fails the first condition of isDid() like 'abcdefg'

tests/core/Did.spec.ts
```diff
@@ -0,0 +1,8 @@
...
+});
```
gjgd: coud add a positive test showing that 'did:abc:def' is a Did

tests/core/Operation.spec.ts
```diff
@@ -34,4 +36,220 @@ describe('Operation', async () => {
...
+
+    it('should throw error if `id` of a public key in an add-public-keys patch is not a string.', async () => {
+      const updatePayload = generateUpdatePayloadForPublicKeys();
+      (updatePayload.patches[0].publicKeys![0] as any).id = { invalidType: true };
+
+      const expectedError = new SidetreeError(ErrorCode.OperationUpdatePatchPublicKeyIdNotString);
+      expect(() => { Operation.validateUpdatePayload(updatePayload); }).toThrow(expectedError);
+    });
+
+    it('should throw error if the a secp256k1 public key in an add-public-keys patch is not in `publicKeyHex` format.', async () => {
```
OR13: I know we are using compressed public key format, we should maybe add a test here to ensure that the none compressed public key format won't be added, because that could cause some strange errors later...

In particular, ethereum users are likely to try and register uncompressed keys since, they are used to calculate addresesses normally...

https://bitcoin.stackexchange.com/questions/44260/can-ethereum-support-compressed-secp256k1-public-keys-to-compute-keccak-addresse

tests/generators/VegetaLoadGenerator.ts
```diff
@@ -42,12 +43,11 @@ export default class VegetaLoadGenerator {
       fs.writeFileSync(absoluteFolderPath + `/requests/create${i}.json`, createOperationBuffer);
 
       // Compute the DID unique suffix from the generated Create payload.
-      const didUniqueSuffix = Did.getUniqueSuffixFromEncodeDidDocument(createPayload, ProtocolParameters.get(1000000).hashAlgorithmInMultihashCode);
+      const didUniqueSuffix = Did.getUniqueSuffixFromEncodeDidDocument(createPayload, ProtocolParameters.get(blockchainTime).hashAlgorithmInMultihashCode);
```
OR13: I don't see tests in this PR for this, but I see there is an open issue for handling it later: https://github.com/decentralized-identity/sidetree/issues/265

lib/core/Did.ts
thehenrytsai: Won't fix unless there is a compelling reason. For the context of Sidetree, any DID without a unique DID suffix is invalid.

tests/core/Did.spec.ts
thehenrytsai: That condition is tested already by test `should throw error if any of the service endpoints in the add-service-endpoints patch is not a valid DID.` DID.ts has 100% code coverage.

tests/core/Did.spec.ts
thehenrytsai: DID.ts has 100% code coverage. The positive case has been hit 128 times by the tests.

tests/generators/VegetaLoadGenerator.ts
thehenrytsai: #265 is for something else, this change is more generic, it is simply removing a hardcoded value.

I think what you are looking for is a way to run load test using this tool, which is tracked by #119, though I am still guity for not doing it because it is not consider "mainnet" blocking.

tests/core/Operation.spec.ts
thehenrytsai: Will add a length check of 66 chars.

tests/generators/VegetaLoadGenerator.ts
OR13: yes, basically want to understand how to implement protocol upgrades, and have unit / integration tests around that.

Issue #269, opened 2019-07-23
Mechanism for implementing code-level protocol upgrades with backward compatibility


csuwildcat:

OR13:
I believe there is some linkage between `blockchainTime` and supported crypto / hashing algorithms.

It sounds like maybe this ticket is asking for official client version binding. There is no way a client that does not support new hash algorithm can process new DIDs, nor a new client old ones if an old hash algorithm is dropped.

There is this ticket for defining protocol version: https://github.com/decentralized-identity/sidetree/issues/265

We've thought about this a lot with element, in particular, can we support a great UX on top of multiple versions of the sidetree protocol? 

The pruning / dos change being a good example of a major change. We'd like to be able to support ongoing use and the upgraded version.

Protocol version could be embedded in the anchorFile.

Old clients could ignore new protocol version anchorFiles, but the state they built should probably not be trusted.

I think we need to test a series of client upgrades in integration tests, where the structure of files changes. Probably a blocking dependency for https://github.com/decentralized-identity/sidetree/issues/266

csuwildcat:
The way Henry is implementing this assumes that upgrades to the protocol that result in breaking/forking changes require nodes to run the upgraded version to remain in the same 'network'. There are upgrades that can be optional, if they only are optimizations to existing protocol rules that do not result in forking conditions, but anything that changes the state of what a node sees as valid DIDs, ops, and resulting DID Document state will require network-wide upgrades.

thehenrytsai:
Implemented.

PR #276, opened 2019-08-16
Issue #269 - Generalized anchorFileHash as anchorString in transactions.

thehenrytsai:
This is to remove the assumption that the string will always be just an Anchor File hash, which could be untrue as protocol version progresses and evolves.

### DIFF
--- a/docs/implementation.md
+++ b/docs/implementation.md
@@ -16,9 +16,9 @@ A full node offers the largest set of features and highest resolution performanc
 
 ### Light Node
 
-A light node is a node that retains the ability to independently resolve DIDs without relying on a trusted party or trusted assertions by other nodes, while minimizing the amount of bandwidth and data required to do so. Light nodes run a copy of the target system's node (e.g. a blockchain) and fetch all minimal _anchor file_ data required to create an independent mapping that enables just-in-time resolution of DIDs.
+A light node is a node that retains the ability to independently resolve DIDs without relying on a trusted party or trusted assertions by other nodes, while minimizing the amount of bandwidth and data required to do so. Light nodes run a copy of the target system's blockchain node and fetch only minimal Sidetree data required to create an independent lookup table that enables just-in-time resolution of DIDs.
 
-> NOTE: Light node support is in development, with release of a supporting node implementation in May 2019.
+> NOTE: Light node support is in development.
 
 ## Observer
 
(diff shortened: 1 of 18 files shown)


docs/implementation.md
```diff
@@ -216,7 +216,7 @@ GET /transactions?since=170&transaction-time-hash=00000000000000000000100158f474
       "transactionNumber": "A monotonically increasing number (need NOT be by 1) that identifies a Sidtree transaction.",
       "transactionTime": "The logical blockchain time this transaction is anchored. Used for protocol version selection.",
       "transactionTimeHash": "The hash associated with the transaction time.",
-      "anchorFileHash": "Hash of the anchor file of this transaction."
+      "anchorString": "The string written to the blockchain for this transaction."
```
OR13: The size limitations of this string should be documented.

OR13: 32 byte utf-8 string seems correct here.

In practice this string may be a protocol specific encoding of a multihash.

32 bytes is not enough for much more than a hash or a uuid.

I worry about the security issues arising from weakening this type.

For example, a uuid might work fine to resolve the anchor file, but provides no built in integrity check.

There is nothing preventing an implementer from returning content for an anchorFile that does not match its hash, but now there will be no way of knowing that is happening from looking at the ledger.

I suggest renaming CAS as well, if content addressing is not longer a core requirement.

in element, we use the term "storage" specifically to eliminate this confusion, we support ipfs, but also centralized databases that can return anchorFiles which don't match the anchorFileHash that was written to the ledger.

For GDPR Compliance, private consortium and development purposes, I think its fine if some anchorFileHashes are not resolved over IPFS / SWARM / Public P2P Storage, but I think that content addressing is a valuable feature of the protocol, and we should try and default to integrity checked storage interfaces.

thehenrytsai: Thanks @OR13,

Updated protocol documentation to include concrete parameter values used in the latest protocol version.

thehenrytsai: Hi @OR13, I did not refresh to get you second comment. You are right that the length of the anchor string will be protocol version specific. @csuwildcat and I had a lengthy chat on what type of checks should be done to prevent abusers from supplying false/bogus data for the anchor string, and we arrived at the conclusion that we can't really prevent abusers from doing it, so we will let protocol version dictate what validation is done at the observing time (remember strict checks are done at resolution time). To be clear, we see CAS still being integral part of the design, meaning the implementer should definitely fail the content if the hash does not match, this change is really to address the case if in the future we write more information than just the anchor file hash to the blockchain.

@csuwildcat feel free to add your 2 cents.

PR #284, opened 2019-09-18
Add support to return the service versions

mudiali:
Issue: #265 Adding a new REST api to return the version information for the core and the dependent bitcoin and ipfs services. The version is read from the package.json file for all the services and is cached during the service start up. If there's any changes to the dependent services version then the core service needs to be restarted to get the latest info.


lib/core/Cas.ts
```diff
@@ -12,7 +14,18 @@ export default class Cas implements ICas {
 
   private fetch = nodeFetch;
 
-  public constructor (public uri: string) { }
+  private versionUri: string // e.g. https://127.0.0.1/version
```
thehenrytsai: Consider hidding `versionUri` in `tryGetServiceVersion()`

lib/core/Blockchain.ts
```diff
@@ -150,4 +164,23 @@ export default class Blockchain implements IBlockchain {
     console.info(`Refreshed blockchain time: ${responseBodyString}`);
     return responseBody;
   }
+
+  /**
+   * Gets the version information by making the REST API call.
+   */
+  private async tryGetServiceVersion() : Promise<ServiceVersionModel> {
```
thehenrytsai: Do you think we should generalize line 174 ~ 179 as one "utility" method?

lib/core/Blockchain.ts
```diff
@@ -150,4 +164,23 @@ export default class Blockchain implements IBlockchain {
...
+
+    try {
+      const response = await this.fetch(this.versionUri);
+
+      const responseBodyString = await ReadableStream.readAll(response.body);
+      console.info("Received version response from the blockchain service: ", responseBodyString);
+
+      return JSON.parse(responseBodyString);
+    } catch (e) {
+      console.error("Ignoring the exception during blockchain service version retrieval: %s", JSON.stringify(e));
```
thehenrytsai: I'm surprised that linter didn't block double quotes, but we use [standardjs](https://standardjs.com/) style which uses single quotes.

In this case consider:

```suggestion
      console.error(`Ignoring the exception during blockchain service version retrieval: ${JSON.stringify(e)}`, );
```

lib/core/Blockchain.ts
```diff
@@ -7,6 +7,8 @@ import ReadableStream from '../common/ReadableStream';
 import SharedErrorCode from '../common/SharedErrorCode';
 import TransactionModel from '../common/models/TransactionModel';
 import { SidetreeError } from './Error';
+import ServiceVersionModel from '../common/models/ServiceVersionModel';
```
thehenrytsai: Nit: we sort imports, and `{..}` imports comes last.

lib/common/ServiceInfo.ts
```diff
@@ -0,0 +1,46 @@
...
+        return {
+            name: this.serviceName,
+            version: ServiceInfo.packageJson.version
+        };        
+    }
+
+    /**
+     * Gets an 'empty' service version object.
+     */
+    public static getEmptyServiceVersion(): ServiceVersionModel {
```
thehenrytsai: Design: consider not having this "dummy" method. I think `tryGetServiceVersion()` can itself return either:

`undefined` or
```
{
  name:  'service_name',
  version: 'undefined'
}

File: lib/common/ServiceInfo.ts [hunk 6]
```diff
@@ -0,0 +1,46 @@
+import ServiceVersionModel from './models/ServiceVersionModel'; 
+
+/**
+ * Encapsulates the functionality to get the information about the service such as
+ * version info.
+ */
+export default class ServiceInfo {
+
+    private static readonly packageJson = require('../../package.json');
```
thehenrytsai: We use [standardjs](https://standardjs.com/rules.html) style which uses 2 spaces for indents.

lib/common/ServiceInfo.ts
```diff
@@ -0,0 +1,46 @@
+import ServiceVersionModel from './models/ServiceVersionModel'; 
+
+/**
+ * Encapsulates the functionality to get the information about the service such as
+ * version info.
+ */
+export default class ServiceInfo {
```
thehenrytsai: Design: I was imagining this class would be doing the fetching and caching, so that `Cas` and `Blockchain` don't have to know about the implementation details.

lib/core/Blockchain.ts
gjgd: I would add that JSON.stringify doesn't work on error objects (returns `{}`) . 
You may want to consider using `JSON.stringify(err, Object.getOwnPropertyNames(err))` (see https://stackoverflow.com/questions/18391212/is-it-not-possible-to-stringify-an-error-using-json-stringify)

lib/core/Cas.ts
```diff
@@ -79,4 +92,30 @@ export default class Cas implements ICas {
...
+   * Gets the cached service version.
+   */
+  public getCachedServiceVersion(): ServiceVersionModel {    
+    return this.cachedVersionModel;
+  }
+
+  /**
+   * Gets the version information by making the REST API call.
+   */
+  private async tryGetServiceVersion() : Promise<ServiceVersionModel> {
```
thehenrytsai: Design: I was imagining this a `ServiceInfoFetcher` class would be doing the fetching and caching, so that `Cas` and `Blockchain` don't have to know about the implementation details.

lib/ipfs/RequestHandler.ts
```diff
@@ -117,4 +121,16 @@ export default class RequestHandler {
     console.error(`Wrote content '${base64urlEncodedMultihash}'.`);
     return response;
   }
+
+  /**
+   * Handles the get version request.
+   */
+  public async handleGetVersionRequest(): Promise<ResponseModel> {
+    var body = JSON.stringify(this.serviceInfo.getServiceVersion());
```
thehenrytsai: Use `const` whenever possible.

tests/bitcoin/BitcoinProcessor.spec.ts
```diff
@@ -156,23 +157,23 @@ describe('BitcoinProcessor', () => {
       }
     });
   });
-
+
```
thehenrytsai: Sanity: are these unintended trailing spaces?

lib/core/Core.ts
```diff
@@ -98,4 +103,21 @@ export default class Core {
     const response = requestHandler.handleResolveRequest(didOrDidDocument);
     return response;
   }
+
+  /**
+   * Handles the get version request. It gets the versions from the dependent services
+   * as well.
+   */
+  public async handleGetVersionRequest(): Promise<ResponseModel> {
```
thehenrytsai: Consider not having `Request` in method name.
```suggestion
  public async getServiceVersion(): Promise<ResponseModel> {
```

lib/core/ServiceVersionFetcher.ts
```diff
@@ -56,6 +55,18 @@ export default class ServiceVersionFetcher {
       console.error('Ignoring the exception during blockchain service version retrieval: %s', JSON.stringify(e, Object.getOwnPropertyNames(e)));
     }
 
-    return ServiceInfoProvider.emptyServiceVersion;
+    return this.emptyServiceVersion;
+  }
+
+  private get emptyServiceVersion (): ServiceVersionModel {
```
thehenrytsai: Nit: technically static method.

lib/core/ServiceVersionFetcher.ts
```diff
@@ -56,6 +55,18 @@ export default class ServiceVersionFetcher {
...
+  }
+
+  private get emptyServiceVersion (): ServiceVersionModel {
+    return {
+      name: 'undefined',
+      version: 'undefined'
+    };
+  }
+
+  private isEmptyServiceVersion (serviceVersion: ServiceVersionModel): boolean {
```
thehenrytsai: Nit: technically static method.

lib/core/Cas.ts
mudiali: Refactoring this code out to the new ServiceVersionFetcher.ts

lib/common/ServiceInfo.ts
mudiali: Refactoring this code out to the new ServiceVersionFetcher.ts

lib/common/ServiceInfo.ts
mudiali: Changed it to a ServiceInfoProvider.ts

lib/core/Core.ts
mudiali: Leaving this file the way it is and updating the Blockchain.ts to be consistent with the rest of functions within each respective file.

lib/core/ServiceVersionFetcher.ts
mudiali: Function removed. No longer needed.

PR #286, opened 2019-09-20
Fetch service version again if failed earlier + removed initialize().

thehenrytsai:


lib/core/ServiceVersionFetcher.ts
```diff
@@ -20,16 +20,14 @@ export default class ServiceVersionFetcher {
...
   /**
-   * Initializes the properites of this object.
+   * Gets the service version.
+   * Returns `undefined` service version if unable to fetch it.
    */
-  public async initialize () {
-    this.cachedVersion = await this.tryGetServiceVersion();
-  }
+  public async getVersion (): Promise<ServiceVersionModel> {
+    if (this.cachedVersion.version === 'undefined') {
```
mudiali: This is why I was thinking about a function in the ServiceInfo which can return true/false if the version is empty. For example, this call could be:
````
if (ServiceInfo.IsEmptyServiceVersion(this.cachedVersion)) {
   // try fetch again
}
````

thehenrytsai: Ideally, I think the return type can just be `Promise<ServiceVersionModel | undefined>`, but I wanted to keep this PR as isolated as I could.

Issue #229, opened 2019-05-29
Protocol rules for assigning a specific function/purpose to keys


OR13:
Per the DID Spec: https://w3c-ccg.github.io/did-spec/#example-5-various-public-keys

Public Keys must have an `id` property.

Unfortunately, the examples like  `did:example:123456789abcdefghi#keys-2`, don't provide a clear method for a key holder to determine an `id` for the key in a clear way, and thus, in order to determine if a key is present often times the properties of the key must be compared... I suggest we be stricter in a way that helps wallet developers and users. 

I'm borrowing from a couple of projects here:

https://github.com/decentralized-identity/github-did/blob/master/packages/lib/src/v2/func/getPublicKeyFromDIDDoc.js

https://github.com/transmute-industries/did-wallet/blob/master/src/keyToKID.js

I propose that a sidetree public key have and id constructed as follows:

`did:example:123#kid=base64Url(sha256(
  "publicKeyPem" | 
  "publicKeyJwk" | 
  "publicKeyHex" | 
  "publicKeyBase64" | 
  "publicKeyBase58" | 
  "publicKeyMultibase"))`

This will also allow wallets / other systems to index key pairs by public key, which will then aid in signing operations with the correct key.

as a wallet system interacting with sidetree, you will be wanting to retrieve keys easily in constant time, and be able to distinguish between key types / (signature purposes), such a update / recovery.

csuwildcat:
Perhaps we can title this Issue differently, assuming this is squarely about how to mark a key for a certain use/role? I would suggest something like "Protocol rules for assigning a specific function/purpose to keys"

If this is the desire, my opinion is that we should add a property called `function` or `purpose` and recognize a set of values (for example: `update`, `signing`, `encryption`, `recovery`, etc.) that can be added to an array assigned to that property.

OR13:
For example:

```
{
    "id": "did:example:123#kid=456",
    "type": "RsaVerificationKey2018",
    "controller": "did:example:123",
    "purpose": "recovery",
    "publicKeyPem": "-----BEGIN PUBLIC KEY...END PUBLIC KEY-----\r\n"
}
```

I think the note about the `id` property is more of a useful convention (optional), but we do need a way to know that a given key is the recovery key, since only that key can be used in `recover` and `delete` operations.

OR13:
It may be too complex here, but "update" is probably too permissive.

https://medium.com/@markherhold/validating-json-patch-requests-44ca5981a7fc

[...]

Issue #93, opened 2019-01-03
Implement recover operations


thehenrytsai:
Recovery mechanism will dictate how signature verification will be performed for delete and recovery operations.

csuwildcat:
We can quickly spec out one simple option to just include a recovery public key, and assume a payload will be signed with it.

OR13:
example recovery from element:

```
const payload = {
      did: `did:sidetree:${uid}`,
      operationNumber: 2,
      previousOperationHash,
      patch: [
        // first op should update recovery key.
        {
          op: 'replace',
          path: '/publicKey/1',
          value: {
            id: '#recovery',
            type: 'Secp256k1VerificationKey2018',
            publicKeyHex: recoveryKeypair2.publicKey,
          },
        },
        {
          op: 'replace',
          path: '/publicKey/0',
          value: {
            id: '#primary',
            type: 'Secp256k1VerificationKey2018',
            publicKeyHex: primaryKeypair2.publicKey,
          },
        },
      ],
    };
```
Per the spec, a new recovery key is required to be present, but other replacements, deletions are optional.

Recovery is identical to update, except that a specific patch operation is required and the recovery payload is to be signed by a "recovery" key... I took this to mean a key with the id `#recovery`, which means that recovery is not possible until such a key exists in a did document.

OR13:
on a related note, it seems very odd that delete is not also a patch operation, that removes all keys for the doc.

Per the DID spec, a public key that does not exist is considered revoked.

https://w3c-ccg.github.io/did-spec/#public-keys

so any credentials would be auto revoked if they were signed by `did:sidetree:uid#key`, and this key was not present in the resolved document.

Also consider renaming "delete" to "deactivate" 
https://w3c-ccg.github.io/did-spec/#deactivate

OR13:
I'm taking a second pass at this in element as I apply the updates to the protocol.

It seems like the operation processor needs to be aware of a couple things here:

1. A public key with id `#recovery` can only be updated by an op signed by that key, which includes a new public key with id `#recovery`. (implies update logic to prevent any patch that would violate this).
2. delete has no such requirement.

It seems like the generalization here is that sidetree has a way of recognizing a new / not very well spec'ed concept in JSON-LD Signatures called [Proof Purposes](https://github.com/digitalbazaar/jsonld-signatures#introduction), from some attribute of the publicKey, and that operations with a purpose that are signed by a key that supports that proof purpose are valid.

[...]

PR #296, opened 2019-09-28
Issue #229, #93 - Added usage property to public key entries.

thehenrytsai:
1. Added `usage` property to public key entries in preparation for DID recovery & data pruning support.
1. Added test for DID Document validation to increase `Document.ts` code coverage to 100%.
1. Added example tool to show how to use the `VegetaLoadGenerator`.
1. Added a FAQ in implementation doc.


csuwildcat:
Can you explain what this is and what it is intended to do?

csuwildcat:
From the code and some of the content, it appears this is adding a property to designate the intended purpose of the key within the DID public key descriptor? If so, I will start a larger discussion of this in the W3C to ensure what we do here is either adopted (in some form or fashion) or at least is not in conflict with some other plan around this general concept.

PR #299, opened 2019-09-30
Compressing batch and anchor files

mudiali:
The core now compresses both the batch and the anchor files before sending them off to the IPFS service. It uses the zlib library to compress the data in gzip.


lib/core/versions/latest/AnchorFile.ts
```diff
@@ -103,4 +111,15 @@ export default class AnchorFile {
 
     return false;
   }
+
+  /**
+   * Creates a buffer from the input so that the buffer can be persisted.
+   */
+  public static async createBufferFromAnchorFileModel (anchorFileModel: AnchorFileModel): Promise<Buffer> {
```
thehenrytsai: Consider rename method to `createCompressedBufferFromAnchorFileModel` or maybe just `compressAnchorFileModel`

mudiali: I feel like that 'compressing' information is an internal detail and should not be exposed to caller. Tomorrow, the AnchorFile.ts can decide to not compress and that detail should stay within.

lib/core/versions/latest/AnchorFile.ts
```diff
@@ -13,11 +14,18 @@ export default class AnchorFile {
...
+    try {
+      anchorFileDecompressedBuffer = await Compressor.decompressBuffer(anchorFileBuffer);
+    } catch {
+      throw new SidetreeError(ErrorCode.AnchorFileDecompressionFailure);
+    }
 
     let anchorFile;
     try {
-      anchorFile = JSON.parse(anchorFileBuffer.toString());
+      anchorFile = JSON.parse(anchorFileDecompressedBuffer.toString());
```
thehenrytsai: Consider:
```suggestion
      anchorFile = JSON.parse(anchorFileDecompressedBuffer);
```
Looks like you were able to do so for the batch file decompression.

lib/core/versions/latest/BatchWriter.ts
```diff
@@ -49,14 +51,14 @@ export default class BatchWriter implements IBatchWriter {
...
-    const anchorFile = {
+    const anchorFileModel: AnchorFileModel = {
       batchFileHash: batchFileHash,
       merkleRoot: encodedMerkleRoot,
       didUniqueSuffixes
     };
 
     // Make the 'anchor file' available in content addressable store.
-    const anchorFileJsonBuffer = Buffer.from(JSON.stringify(anchorFile));
+    const anchorFileJsonBuffer = await AnchorFile.createBufferFromAnchorFileModel(anchorFileModel);
```
thehenrytsai:
```suggestion
    const anchorFileCompressedBuffer = await AnchorFile.createBufferFromAnchorFileModel(anchorFileModel);
```

mudiali: Same comment as before, I don't think that the BatchFile necessary knows that the AnchorFile is compressing the buffer.

lib/core/versions/latest/util/Compressor.ts
```diff
@@ -0,0 +1,60 @@
...
+export default class Compressor {
+
+  private static readonly gzipAsync = util.promisify(zlib.gzip);
+  private static readonly gUnzipAsync = util.promisify(zlib.gunzip);
+
+  /**
+   * Compresses the data in gzip and returns it as base64url encoded string.
+   * @param input The buffer to compress.
+   */
+  public static async compressAsBase64Url (input: string): Promise<string> {
```
thehenrytsai: Sanity: doesn't appear to be used anywhere except for test?

lib/core/versions/latest/util/Compressor.ts
```diff
@@ -0,0 +1,60 @@
...
+
+    // Casting result to Buffer as that's what is returned by gzip
+    return result as Buffer;
+  }
+
+  /**
+   * Decompresses the base64url encoded input and returns it as a string.
+   * @param buffer The gzip compressed buffer as base64url encoded string.
+   */
+  public static async decompressBase64UrlData (base64UrlInput: string): Promise<string> {
```
thehenrytsai: Sanity: not used except for test? In which case consider removing.

lib/core/versions/latest/AnchorFile.ts
mudiali: Good catch cause the BatchFile is using JsonAsync where as here we weren't. Updated the code to use JsonAsync now.

lib/core/versions/latest/util/Compressor.ts
mudiali: Removing.

PR #300, opened 2019-10-01
Ignore the transactions which have multiple OP_RETURN operations.

mudiali:
Issue: #224

### DIFF
--- a/lib/bitcoin/BitcoinProcessor.ts
+++ b/lib/bitcoin/BitcoinProcessor.ts
@@ -514,34 +514,68 @@ export default class BitcoinProcessor {
         // console.debug(`Skipping transaction ${transactionIndex}: no output coins.`);
         continue;
       }
-      const outputs = transactions[transactionIndex].vout as Array<any>;
-      for (let outputIndex = 0; outputIndex < outputs.length; outputIndex++) {
-        // grab the scripts
-        const script = outputs[outputIndex].scriptPubKey;
 
-        // console.debug(`Checking transaction ${transactionIndex} output coin ${outputIndex}: ${JSON.stringify(script)}`);
-        // check for returned data for sidetree prefix
-        const hexDataMatches = script.asm.match(/\s*OP_RETURN ([0-9a-fA-F]+)$/);
-        if (!hexDataMatches || hexDataMatches.length === 0) {
-          continue;
-        }
-        const data = Buffer.from(hexDataMatches[1], 'hex').toString();
-        if (data.startsWith(this.sidetreePrefix)) {
-          // we have found a sidetree transaction
-          const sidetreeTransaction: TransactionModel = {
-            transactionNumber: TransactionNumber.construct(block, transactionIndex),
-            transactionTime: block,
-            transactionTimeHash: blockHash,
-            anchorString: data.slice(this.sidetreePrefix.length)
-          };
-          console.debug(`Sidetree transaction found; adding ${JSON.stringify(sidetreeTransaction)}`);
-          await this.transactionStore.addTransaction(sidetreeTransaction);
-          // stop processing future anchor strings. Protocol defines only the first should be accepted.
-          break;
-        }
+      try {
+        const outputs = transactions[transactionIndex].vout as Array<any>;
+
+        await this.addSidetreeTransactionsToTransactionStore(this.transactionStore, outputs, transactionIndex, block, blockHash);
+
+      } catch (e) {
+        const inputs = { block: block, blockHash: blockHash, transactionIndex: transactionIndex };
+        console.error('An error happened when trying to add sidetree transaction to the store. Moving on to the next transaction. Inputs: %s\r\nFull error: %s',
+                       JSON.stringify(inputs),
+                       JSON.stringify(e, Object.getOwnPropertyNames(e)));
       }
     }
+
     return blockHash;
+  }
+
+  private async addSidetreeTransactionsToTransactionStore (
+    transactionStore: MongoDbTransactionStore,
+    allVOuts: Array<any>,
+    transactionIndex: number,
+    transactionBlock: number,
+    transactionHash: any): Promise<void> {
+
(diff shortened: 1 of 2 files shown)


lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -514,36 +514,70 @@ export default class BitcoinProcessor {
...
+                       JSON.stringify(inputs),
+                       JSON.stringify(e, Object.getOwnPropertyNames(e)));
       }
     }
+
     return blockHash;
   }
 
+  private async addSidetreeTransactionsToTransactionStore (
+    transactionStore: MongoDbTransactionStore,
```
thehenrytsai: No need to pass `transactionStore`.

mudiali: We're using it on line 577.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -514,36 +514,70 @@ export default class BitcoinProcessor {
...
+      if (!hexDataMatches || hexDataMatches.length === 0) {
+        continue;
+      }
+
+      const data = Buffer.from(hexDataMatches[1], 'hex').toString();
+      const isSidetreeTx = data.startsWith(this.sidetreePrefix);
+      const oneSidetreeTxAlreadyFound = sidetreeTxToAdd !== undefined;
+
+      if (isSidetreeTx && oneSidetreeTxAlreadyFound) {
+        throw new Error('The transaction has more then one sidetree transactions.');
```
thehenrytsai: Consider:
```suggestion
        throw new Error('The transaction has more then one sidetree anchor strings.');
```

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -514,36 +514,70 @@ export default class BitcoinProcessor {
...
-        }
+
+      try {
+        const outputs = transactions[transactionIndex].vout as Array<any>;
+
+        await this.addSidetreeTransactionsToTransactionStore(this.transactionStore, outputs, transactionIndex, block, blockHash);
+
+      } catch (e) {
+        const inputs = { block: block, blockHash: blockHash, transactionIndex: transactionIndex };
+        console.error('An error happened when trying to add sidetree transaction to the store. Moving on to the next transaction. Inputs: %s\r\nFull error: %s',
```
thehenrytsai: I should this should be `console.debug`.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -514,36 +514,70 @@ export default class BitcoinProcessor {
...
-          console.debug(`Sidetree transaction found; adding ${JSON.stringify(sidetreeTransaction)}`);
-          await this.transactionStore.addTransaction(sidetreeTransaction);
-          // stop processing future anchor strings. Protocol defines only the first should be accepted.
-          break;
-        }
+
+      try {
+        const outputs = transactions[transactionIndex].vout as Array<any>;
+
+        await this.addSidetreeTransactionsToTransactionStore(this.transactionStore, outputs, transactionIndex, block, blockHash);
```
thehenrytsai: Consider renaming the method `addSidetreeTransactionsToTransactionStore` to something like `processTransactionVOuts` instead. Because the current name seems to imply that the transaction will for sure be a Sidetree transaction.

tests/bitcoin/BitcoinProcessor.spec.ts
```diff
@@ -839,15 +839,22 @@ describe('BitcoinProcessor', () => {
   describe('processBlock', () => {
 
     // creates a response object for Bitcoin
-    async function generateBlock (blockHeight: number, data?: () => string | undefined): Promise<any> {
+    async function generateBlock (blockHeight: number, data?: () => string | string[] | undefined): Promise<any> {
```
thehenrytsai: Why not just `data?: () => string[] | undefined`? Too invasive of a change?

tests/bitcoin/BitcoinProcessor.spec.ts
```diff
@@ -971,6 +978,45 @@ describe('BitcoinProcessor', () => {
...
       done();
     });
+
+    it('should ignore any transactions that have multiple OP_RETURN in them', async (done) => {
+      const block = randomNumber();
+      let shouldFindIDs: string[] = [];
+      const blockData = await generateBlock(block, () => {
+        const id = randomString();
+        const rand = Math.random();
+
```
thehenrytsai: Add comment for percentage.

lib/bitcoin/BitcoinProcessor.ts
mudiali: Renaming to addValidSidetreeTransactionsFromVOutsToTransactionStore(…)

tests/bitcoin/BitcoinProcessor.spec.ts
mudiali: Yea .. then the callers of the functions need to change to return an array

tests/bitcoin/BitcoinProcessor.spec.ts
mudiali: Comments added.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -514,36 +514,70 @@ export default class BitcoinProcessor {
...
+      // console.debug(`Checking transaction ${transactionIndex} output coin ${outputIndex}: ${JSON.stringify(script)}`);
+      // check for returned data for sidetree prefix
+      const hexDataMatches = script.asm.match(/\s*OP_RETURN ([0-9a-fA-F]+)$/);
+
+      if (!hexDataMatches || hexDataMatches.length === 0) {
+        continue;
+      }
+
+      const data = Buffer.from(hexDataMatches[1], 'hex').toString();
+      const isSidetreeTx = data.startsWith(this.sidetreePrefix);
```
thehenrytsai: Consider passing `this.sidetreePrefix` as method parameter.

PR #301, opened 2019-10-07
Implement an IVersionManager to avoid passing function pointers in Core.ts

mudiali:
- The VersionManager.ts now implements IVersionManager. 
- All components which require the 'versioned' objects now take in the IVersionManager instead of taking the function pointers as before.
- Also added tests for VersionManager.ts


tests/core/BatchScheduler.spec.ts
```diff
@@ -2,12 +2,18 @@ import * as retry from 'async-retry';
 import BatchScheduler from '../../lib/core/BatchScheduler';
 import MockBatchWriter from '../mocks/MockBatchWriter';
 import MockBlockchain from '../mocks/MockBlockchain';
+import MockVersionedProvider from '../mocks/MockVersionManager';
```
thehenrytsai: Nit:
```suggestion
import MockVersionManager from '../mocks/MockVersionManager';
```

tests/core/versions/testversion1/BatchWriter.ts
```diff
@@ -0,0 +1,18 @@
+import IBatchWriter from '../../../../lib/core/interfaces/IBatchWriter';
```
thehenrytsai: Nit: Consider following more of javascript convension of using hyphens between words.
`testversion1` -> `test-version-1` or maybe avoiding it altogether to `0.0.1`

mudiali: Renaming.

tests/core/BatchScheduler.spec.ts
mudiali: Good catch … I missed this during the name changes.

Issue #219, opened 2019-05-23
Do not revert transactions if last transaction time in Core is greater than Blockchain microservice


thehenrytsai:
Do not revert transactions if last transaction time in Core is greater than Blockchain microservice. This can mean Blockchain microservice is reconstructing transactions from scratch.

PR #311, opened 2019-10-16
Issue #219 - Prevented rollback if current Blockchain time is behind last txn time.

thehenrytsai:


lib/core/Observer.ts
```diff
@@ -114,14 +113,27 @@ export default class Observer {
           void this.processTransaction(transaction, awaitingTransaction);
         }
 
+        // NOTE: Blockchain reorg has happened for sure only if `invalidTransactionNumberOrTimeHash` AND
+        // latest transaction time is less or equal to blockchain service time.
+        // This check will prevent Core from reverting transactions if/when blockchain service is reinitializing its data itself.
+        let blockReorganizationDetected = false;
```
AlexITC: The only reliable way I know for detecting rollbacks is to look for the least-common ancestor from the stored headers and what bitcoin has, see [this](https://github.com/X9Developers/block-explorer/blob/develop/server/app/com/xsn/explorer/services/synchronizer/LedgerSynchronizationStatusService.scala#L42) for details.

thehenrytsai: You are right @AlexITC, this logic is handled in the blockchain (bitcoin) microservice.

PR #315, opened 2019-10-23
Arvind/pof

arvinda:
An initial pull request for the proof-of-fee design. I will send a new iteration in a day or so after more testing and debugging. 

If you want to review the current design please take a look; otherwise feel free to wait for the new iteration.


lib/common/models/TransactionModel.ts
```diff
@@ -6,4 +6,5 @@ export default interface TransactionModel {
   transactionTime: number;
   transactionTimeHash: string;
   anchorString: string;
+  feePaid: number;
```
mudiali: We also want to add the 'normalizedFee' for this block into this model. The thinking being is that once the TransactionModel is returned to the core, it has all the data it needs to validate the feePaid vs. the normalizedFee (see #322 for details). Any idea on how much change is required for that?

lib/bitcoin/IProofOfFeeConfig.ts
```diff
@@ -0,0 +1,22 @@
+
+/**
+ * Defines configuration of approximate sliding windows used in
+ * proof of fee calculation.
+ */
+export interface ISlidingWindowQuantileConfig {
```
mudiali: 1. For sidetree, only the classic OOP interfaces are prefixed with I. For such data holder objects, you can remove the I prefix.
2. Please create a models folder and put all such models in there.
3. Please create each class in a separate file in the models folder.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -507,18 +548,31 @@ export default class BitcoinProcessor {
...
       // get the output coins in the transaction
-      if (!('vout' in transactions[transactionIndex])) {
+      if (!('vout' in transaction)) {
         // console.debug(`Skipping transaction ${transactionIndex}: no output coins.`);
         continue;
       }
 
-      try {
-        const outputs = transactions[transactionIndex].vout as Array<any>;
+      // Add the transaction to the sampler.  We filter out transactions with unusual
```
mudiali: As per our offline discussion, please filter out the 'sidetree' transactions as well.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -82,6 +86,16 @@ export default class BitcoinProcessor {
...
 
+  /** proof of fee configuration */
+  private readonly proofOfFeeConfig: IProofOfFeeConfig;
+
+  private readonly quantileCalculator: SlidingWindowQuantileCalculator;
+
+  private readonly transactionSampler: ReservoirSampler;
+
+  /** satoshis per bitcoin */
+  private static readonly satoshiPerBTC = 100000000;
```
thehenrytsai: Nit: casing, should be `satoshiPerBtc`, but would actually prefer `satoshisPerBitcoin`.

lib/bitcoin/IProofOfFeeConfig.ts
```diff
@@ -0,0 +1,22 @@
...
+  windowSizeInBatches: number;
```
thehenrytsai: Can we find a different word term than `batch` 'coz every time I see it, I think of Sidetree operation batches, which I think is entirely different here. Perhaps `aggregation` or `grouping` or `set`?

lib/bitcoin/IProofOfFeeConfig.ts
```diff
@@ -0,0 +1,22 @@
...
+  batchSizeInBlocks: number;
+  feeApproximation: number;
```
thehenrytsai: Can we add documentation on what `feeApproximation` mean? Or if it is hard to explain at this level without looking at ValueApproximator, is it really necessary to expose this as a configurable parameter? Is it possible to hardcode to a particular value?

lib/bitcoin/SlidingWindowQuantileCalculator.ts
```diff
@@ -0,0 +1,307 @@
...
+ *
+ * For example, if we use approximation 2, the number 3 and 4 would be normalized to 3,
+ * 5,6,7,8 would be normalized to 4, and so on. Denormalization would denormalize 3 to 4,
+ * 4 to 8, and so on. If we take a number n, the number denormalize(normalize(n)) would be
+ * larger than n, but at most by a factor 2.
+ *
+ * We can replace 2 in the argument above with a value closer to 1 (say 1.1), and all values
+ * will be approximated by a factor at most 1.1.
+ */
+export class ValueApproximator {
```
thehenrytsai: Sorry, I am failing to confidently understand the class by reading the comments above without looking into the implementation. Is this class attempting to deterministically compact a number to its "order/magnitude"? Seems that's not the case, but the example you gave is not enough for me deduct the pattern (sorry for my low IQ).

lib/bitcoin/IProofOfFeeConfig.ts
```diff
@@ -0,0 +1,22 @@
...
+  sampleSize: number;
+  quantile: number;
```
thehenrytsai: Can we add comments on what this means.

lib/bitcoin/IProofOfFeeConfig.ts
```diff
@@ -0,0 +1,22 @@
+
+/**
+ * Defines configuration of approximate sliding windows used in
+ * proof of fee calculation.
+ */
+export interface ISlidingWindowQuantileConfig {
+  windowSizeInBatches: number;
+  batchSizeInBlocks: number;
+  feeApproximation: number;
+  sampleSize: number;
```
thehenrytsai: Can we add comments on what this means.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -528,14 +582,80 @@ export default class BitcoinProcessor {
...
+    }
+
     return blockHash;
   }
 
+  private isBatchBoundary (block: number): boolean {
+    return (block + 1) % this.proofOfFeeConfig.transactionFeeQuantileConfig.batchSizeInBlocks === 0;
+  }
+
+  private getBatchId (block: number): number {
```
thehenrytsai: Design sanity: is this approach resilient to rollbacks? I have a suspicion that rollback needs to be handled. 2 strategies: either rollback when reorg happens; or only store samples that are older "enough" that reorg is mathematically impossible to occur.

lib/bitcoin/PsuedoRandomSampler.ts
```diff
@@ -0,0 +1,232 @@
...
+    // We have a full sampleSize of samples at this point.
+    // This element is picked to be in the sample with probability 1/streamSize
+    this.streamSize++;
+    if (this.psuedoRandomGenerator!.getBernoulliSample(this.sampleSize, this.streamSize) === 1) {
+
+      // If this element is picked, we need to decide which element to evict at random
+      const randIndex = this.psuedoRandomGenerator!.getRandomNumber(this.sampleSize);
+
+      // evict the current element at randIndex and store element instead
+      this.sample[randIndex] = element;
```
thehenrytsai: Design: what is the reason for the random stream (Bernoulli) selection/eviction strategy? Is it to save memory in case of "large sliding window"? If that's the case, I feel storing large number of transaction IDs and perform a static random selection at the end isn't a bad alternative, *if* it leads to easier to understand and/or less code.

lib/bitcoin/SlidingWindowQuantileCalculator.ts
```diff
@@ -0,0 +1,307 @@
...
+
+    // calculate and materialize the quantile as of this batchId
+    const quantile = this.calculateCurrentQuantile();
+    this.historicalQuantiles.set(batchId, quantile);
+
+    // store it into mongo store
+    const quantileInfo = {
+      batchId,
+      quantile,
+      batchFreqVector: runLengthEncode(batchFrequencyVector)
```
thehenrytsai: Design sanity: looks like only the "recent" `batchFreqVector` will be needed (for calculating the next quantile)? If that's the case, it doesn't look like there is code for deleting those no-longer-needed batchFrequency vectors from MongoDB. Which leads to the next question: maybe it is fine having to recompute some quantiles when the service restarts and thus we don't store the `batchFreqVector`?

lib/bitcoin/BitcoinProcessor.ts
arvinda: Good catch. I missed rollbacks. I went with the first approach since the second approach relies on probabilistic assumptions regarding rollbacks. The current code is designed to work for any rollback value.

arvinda: To also simplify the code, I rollback to a "batch" boundary when I can reset state cleanly. The assumption here is that rollbacks are rare events. If we know otherwise, we can avoid rollbacks to batch boundaries for "recent" rollbacks - I would suggest implementing this only if we have a data point suggesting the current solution is expensive.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -2,15 +2,19 @@ import * as httpStatus from 'http-status';
 import MongoDbTransactionStore from '../common/MongoDbTransactionStore';
 import nodeFetch, { FetchError, Response, RequestInit } from 'node-fetch';
 import ErrorCode from '../common/SharedErrorCode';
+import FeeModel from '../common/models/FeeModel';
 import ReadableStream from '../common/ReadableStream';
 import RequestError from './RequestError';
 import ServiceInfo from '../common/ServiceInfoProvider';
 import ServiceVersionModel from '../common/models/ServiceVersionModel';
+import { SlidingWindowQuantileCalculator } from './SlidingWindowQuantileCalculator';
```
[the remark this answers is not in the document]
arvinda: Done. Thx.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -528,14 +582,80 @@ export default class BitcoinProcessor {
...
+    return (inputSatoshiSum - outputSatoshiSum);
+  }
+
   private async addValidSidetreeTransactionsFromVOutsToTransactionStore (
     allVOuts: Array<any>,
     transactionIndex: number,
     transactionBlock: number,
-    transactionHash: any): Promise<void> {
+    transactionHash: any,
+    transactionFee: number): Promise<void> {
```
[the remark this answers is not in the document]
arvinda: Good suggestion. I have pushed the fee calculation to inside the addValidSidetree... function and it happens only if all error cases are ruled out.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -91,6 +105,17 @@ export default class BitcoinProcessor {
     this.bitcoinFee = config.bitcoinFee;
     this.genesisBlockNumber = config.genesisBlockNumber;
     this.transactionStore = new MongoDbTransactionStore(config.mongoDbConnectionString, config.databaseName);
+    this.proofOfFeeConfig = config.proofOfFeeConfig;
```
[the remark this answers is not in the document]
arvinda: Yes. How do you guys do such dependencies? Should I start a new PR for ion?

lib/common/models/TransactionModel.ts
arvinda: Sure, makes sense and it would not be a lot of code. This suggests we can remove the getFee() api change, right?

lib/bitcoin/BitcoinProcessor.ts
arvinda: Done.

lib/bitcoin/IProofOfFeeConfig.ts
```diff
@@ -0,0 +1,22 @@
...
+  quantile: number;
+}
+
+/**
+ * Defines configuration of proof of fee calculation.
+ */
+export interface IProofOfFeeConfig {
```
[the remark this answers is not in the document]
arvinda: Took the suggestion.

lib/bitcoin/SlidingWindowQuantileMongoStore.ts
```diff
@@ -0,0 +1,104 @@
...
+    const collectionNames = collections.map(collection => collection.collectionName);
+
+    if (collectionNames.includes(SlidingWindowQuantileMongoStore.quantileCollectionName)) {
+      console.info('Quantile collection already exists.');
+      return db.collection(SlidingWindowQuantileMongoStore.quantileCollectionName);
+    } else {
+      console.info('Quantile collection does not exists, creating...');
+      const quantileCollection = await db.createCollection(SlidingWindowQuantileMongoStore.quantileCollectionName);
+      await quantileCollection.createIndex({ batchId: 1 }, { unique: true });
+      console.info('Transaction collection created.');
```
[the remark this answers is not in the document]
arvinda: :-) Fixed.

lib/bitcoin/PsuedoRandomSampler.ts
arvinda: We discussed this in our previous meeting. I simplified the reservoir sampling to not use two psuedorandom calls. The sampling code is around 4 lines - I am not sure if the static sampling would be much simpler.

lib/bitcoin/SlidingWindowQuantileCalculator.ts
arvinda: We discussed this at length in the last meeting. You are correct that only the recent "batchFreqVector" values are needed. However, with the run length compression and value approximation, batchFreqVector values are really small strings, so I believe it is not worth the coding complexity to remove them. 

Recomputing quantiles when service restarts is probably not a good idea: we want windows to be very large to keep proof-of-fee values smooth; further, this would complicate the bitcoinprocessor code - the recomputation would be doing everything the regular scan is doing but storing the sidetree transactions in the transaction store.

arvinda: I am going to mark this as resolved - please feel free to revive it if necessary.

lib/bitcoin/PsuedoRandomSampler.ts
```diff
@@ -0,0 +1,232 @@
...
+    if (this.streamSize < this.sampleSize) {
+      this.sample[this.streamSize] = element;
+      this.streamSize++;
+      return;
+    }
+
+    // We have a full sampleSize of samples at this point.
+    // This element is picked to be in the sample with probability 1/streamSize
+    this.streamSize++;
+    if (this.psuedoRandomGenerator!.getBernoulliSample(this.sampleSize, this.streamSize) === 1) {
```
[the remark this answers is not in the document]
arvinda: I have changed the code to remove calls to getBernoulliSample. Please check.

lib/bitcoin/SlidingWindowQuantileMongoStore.ts
```diff
@@ -0,0 +1,104 @@
...
+export interface QuantileInfo {
+  batchId: number;
+  quantile: number;
+  batchFreqVector: number[];
+}
+
+/**
+ * MongoDB store for sliding window quantile information.
+ */
+export class SlidingWindowQuantileMongoStore {
```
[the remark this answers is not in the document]
arvinda: Changed the name to the suggested value.

lib/bitcoin/BitcoinProcessor.ts
arvinda: Went with satoshisPerBitcoin

lib/bitcoin/PsuedoRandomSampler.ts
```diff
@@ -0,0 +1,232 @@
...
+}
+
+/**
+ * Implement the reservoir sampling technique which maintains a uniform random
+ * sample over a "stream" of unknown size (https://en.wikipedia.org/wiki/Reservoir_sampling)
+ *
+ * We can add any number of elements to be considered for sampling and at any point request
+ * a sample of preconfigured size over the elements that have been added so far.
+ */
+export class ReservoirSampler {
```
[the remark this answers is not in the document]
arvinda: Done.

lib/bitcoin/SlidingWindowQuantileCalculator.ts
```diff
@@ -0,0 +1,307 @@
...
+ * Define a sliding window quantile calculator that computes
+ * approximate quantiles over a sliding window. Elements
+ * are added to sliding windows in batches, where each bag
+ * is a collection of values. The sliding window size is
+ * specified in number of batches: when the i'th batch is added,
+ * all the elements added in (i-N)th batch are removed where
+ * N is the size of the sliding window.
+ *
+ */
+export class SlidingWindowQuantileCalculator {
```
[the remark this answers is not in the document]
arvinda: Took the suggestion.

lib/bitcoin/SlidingWindowQuantileCalculator.ts
arvinda: Yes - with the guarantee that we lose a configurable amount in accuracy during this step.

lib/bitcoin/IProofOfFeeConfig.ts
arvinda: went with groupId

lib/bitcoin/IProofOfFeeConfig.ts
arvinda: Implemented these suggestions.

arvinda: Implemented these suggestions.

lib/bitcoin/PsuedoRandomSampler.ts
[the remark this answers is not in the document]
arvinda: Took the suggestion.

lib/bitcoin/SlidingWindowQuantileCalculator.ts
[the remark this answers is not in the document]
arvinda: Took the suggestion.

lib/bitcoin/IProofOfFeeConfig.ts
arvinda: I went with 'group'.

lib/common/models/TransactionModel.ts
arvinda: Added a number, normalizedFee.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -507,18 +548,31 @@ export default class BitcoinProcessor {
...
+      // not worth the cost for an approximate measure.
+      const inputsCount = (transaction.vin as Array<any>).length;
+      if (inputsCount <= this.proofOfFeeConfig.maxTransactionInputCount) {
+        this.transactionSampler.addElement(transaction.txid);
+      }
 
-        await this.addValidSidetreeTransactionsFromVOutsToTransactionStore(outputs, transactionIndex, block, blockHash);
+      try {
+        const outputs = transaction.vout as Array<any>;
+        const transactionFee = await this.getTransactionFeeInSatoshi(transaction.txid);
```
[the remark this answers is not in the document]
arvinda: Took the suggestion (Ali had the same one as well).

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -528,14 +582,80 @@ export default class BitcoinProcessor {
...
+    let inputSatoshiSum = 0;
+    for (let i = 0 ; i < xact.vin.length ; i++) {
+      const xactOutValue = await this.getTransactionOutValueInSatoshi(xact.vin[i].txid, xact.vin[i].vout);
+      inputSatoshiSum += xactOutValue;
+    }
+
+    // transaction outputs in satoshis
+    const xactOuts: number[] = xact.vout.map((v: any) => Math.round((v.value as number) * BitcoinProcessor.satoshiPerBTC));
+
+    const outputSatoshiSum = xactOuts.reduce((s, v) => s + v, 0);
```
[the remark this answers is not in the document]
arvinda: Took the suggestion.

lib/bitcoin/IProofOfFeeConfig.ts
arvinda: Added.

lib/bitcoin/IProofOfFeeConfig.ts
arvinda: Took the suggestion.

lib/bitcoin/IProofOfFeeConfig.ts
```diff
@@ -0,0 +1,22 @@
...
+  transactionFeeQuantileConfig: ISlidingWindowQuantileConfig;
+  quantileScale: number;
```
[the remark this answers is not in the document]
arvinda: Sure, makes sense. I remove this parameter.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -528,14 +582,80 @@ export default class BitcoinProcessor {
...
+    return (block + 1) % this.proofOfFeeConfig.transactionFeeQuantileConfig.batchSizeInBlocks === 0;
+  }
+
+  private getBatchId (block: number): number {
+    return Math.floor(block / this.proofOfFeeConfig.transactionFeeQuantileConfig.batchSizeInBlocks);
+  }
+
+  /** Get the transaction out value in satoshi, for a specified output index */
+  private async getTransactionOutValueInSatoshi (txid: string, outIdx: number) {
+    const xact = await this.rpcCall({
```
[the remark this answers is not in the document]
arvinda: xact is a standard shorthand for transaction in the database community :-) I did expand all names that could find.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -528,14 +582,80 @@ export default class BitcoinProcessor {
...
+      params: [
+        txid,  // transaction id
+        true   // verbose
+      ]
+    });
+
+    // output with the desired index
+    const vout = xact.vout.find((v: any) => v.n === outIdx);
+
+    return Math.round(vout.value * BitcoinProcessor.satoshiPerBTC);
```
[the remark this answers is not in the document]
arvinda: I remember fixing a bug sometime back - for some reason this value turned out to be non-integral - something like x.99999999, perhaps because satoshiperBTC is a really large number.

lib/bitcoin/SlidingWindowQuantileCalculator.ts
```diff
@@ -0,0 +1,307 @@
...
+  public async add (batchId: number, batch: number[]): Promise<void> {
+
+    // Our historical quantiles storage logic relies on batchIds being
+    // consecutive numbers: explicitly check this fact.
+    if (this.prevBatchId) {
+      if (this.prevBatchId + 1 !== batchId) {
+        throw Error(`Quantile calculator batchIds not sequential`);
+      }
+    }
+    this.prevBatchId = batchId;
```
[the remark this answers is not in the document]
arvinda: Sure - I moved this to the very end.

lib/bitcoin/SlidingWindowQuantileCalculator.ts
```diff
@@ -0,0 +1,307 @@
...
+  private mongoStore: SlidingWindowQuantileMongoStore;
+
+  /**
+   * Construct a sliding window quantile computer with specified
+   * approximation paramenters.
+   */
+  public constructor (approximation: number,
+    maxValue: number,
+    private readonly size: number,
+    private readonly quantileParameter: number,
```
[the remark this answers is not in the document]
arvinda: Added - please check.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -507,18 +548,31 @@ export default class BitcoinProcessor {
 
     // console.debug(`Block ${block} contains ${transactions.length} transactions`);
 
+    // reseed source of psuedo-randomness to the blockhash
+    this.transactionSampler.resetPsuedoRandomSeed(blockHash);
```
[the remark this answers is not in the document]
arvinda: Good suggestion. All this code is now refactored into processBlockForPofCalculation().

lib/bitcoin/IProofOfFeeConfig.ts
arvinda: Yes, I have added. I think it is fine to hardcode this - but can we leave this on for now - we need to use this with many different values to get a good set of initial configuration values.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -471,27 +493,37 @@ export default class BitcoinProcessor {
...
-        console.info(`reverted Transactions to block ${firstValidTransaction.transactionTime}`);
-        return firstValidTransaction.transactionTime;
+        console.debug(`Removing transactions since ${TransactionNumber.getBlockNumber(revertToTransactionNumber)}`);
+        await this.transactionStore.removeTransactionsLaterThan(revertToTransactionNumber);
+
+        // Reset transaction sampling
+        this.transactionSampler.clear();
+
+        // Revert the quantile calculator
+        await this.quantileCalculator.removeGroupsGreaterThanOrEqual(revertToBlockNumber + 1);
```
thehenrytsai: For my understanding, any particular reason why have `removeGroupsGreaterThanOREQUAL()` and pass a `+1` instead of just `removeGroupsGreaterThan()` and not pass a `+1`?

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -471,27 +493,37 @@ export default class BitcoinProcessor {
...
-      // Else we repeat the process using the newly reduced list of trasactions.
+      // We did not find a valid transaction - revert as much as the lowest height in the exponentially spaced
+      // transactions and repeat the process with a new reduced list of transactions.
+      const lowestHeight = exponentiallySpacedTransactions[exponentiallySpacedTransactions.length - 1].transactionTime;
+      const revertToTransactionNumber = TransactionNumber.construct(lowestHeight, 0);
+
+      console.debug(`Removing transactions since ${TransactionNumber.getBlockNumber(revertToTransactionNumber)}`);
+      await this.transactionStore.removeTransactionsLaterThan(revertToTransactionNumber);
     }
+
```
thehenrytsai: Sanity: We would have a an infinite loop if the first transaction in DB is invalid.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -524,6 +556,96 @@ export default class BitcoinProcessor {
...
+    // Nothing was found
+    return undefined;
+  }
+
+  private isGroupBoundary (block: number): boolean {
+    return (block + 1) % ProtocolParameters.groupSizeInBlocks === 0;
+  }
+
+  private getgroupId (block: number): number {
```
thehenrytsai: use camelCase

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -458,6 +470,16 @@ export default class BitcoinProcessor {
     };
   }
 
+  /**
+   * For proof of fee calculation, blocks are grouped into fixed sized groups.
+   * This function rounds a block to the first block in its group and returns that
+   * value.
+   */
+  private roundToGroupBoundary (block: number): number {
+    const groupId = Math.floor(block / ProtocolParameters.groupSizeInBlocks);
```
thehenrytsai: Consider renaming to `getFirstBlockInGroup`. Also consider reusing `getgroupId()`.

lib/bitcoin/models/ProtocolParameters.ts
```diff
@@ -0,0 +1,28 @@
...
+  /** Size of the window in number of groups */
+  windowSizeInGroups: number;
+
+  /**
+   * Transaction fees of bitcoin transactions are rounded so that we need to store
+   * a smaller number of distinct values for computing quantiles. This parameter controls
+   * the rounding - higher the value, poorer the rounded value approximates the original fee,
+   * but lesser the space.
+   */
+  feeApproximation: number;
```
thehenrytsai: I don't think this is needed to be expose in v1.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -319,16 +317,30 @@ export default class BitcoinProcessor {
...
+
+    const currentBlockNumber = await this.getCurrentBlockHeight();
+    if (block > currentBlockNumber) {
+      const error = `The input block number must be less than: ${currentBlockNumber}`;
+      console.error(error);
+      throw new RequestError(ResponseStatus.BadRequest, ErrorCode.BlockchainTimeOutOfRange);
+    }
+
+    const blockAfterHistoryOffset = Math.max(block - ProtocolParameters.historicalOffsetInBlocks, 0);
+    const groupId = Math.floor(blockAfterHistoryOffset / ProtocolParameters.groupSizeInBlocks);
```
thehenrytsai: Consider reusing `getgroupId()`.

lib/bitcoin/IProofOfFeeConfig.ts
```diff
@@ -0,0 +1,22 @@
...
+  maxTransactionInputCount: number;
```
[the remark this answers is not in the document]
mudiali: Done.

lib/bitcoin/PsuedoRandomSampler.ts
```diff
@@ -0,0 +1,232 @@
...
+      return;
+    }
+
+    // We have a full sampleSize of samples at this point.
+    // This element is picked to be in the sample with probability 1/streamSize
+    this.streamSize++;
+    if (this.psuedoRandomGenerator!.getBernoulliSample(this.sampleSize, this.streamSize) === 1) {
+
+      // If this element is picked, we need to decide which element to evict at random
+      const randIndex = this.psuedoRandomGenerator!.getRandomNumber(this.sampleSize);
```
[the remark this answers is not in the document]
mudiali: Crypto.getRandomValues doesn't take the seed value so we cannot use it :(

lib/bitcoin/BitcoinProcessor.ts
mudiali: Opened issue #329 to track this issue.

lib/bitcoin/models/ProtocolParameters.ts
mudiali: Will make the change in the subsequent integration changes with the core.

Issue #216, opened 2019-05-22
Design and implement per-op fee calculation algorithm


csuwildcat:

csuwildcat:
@arvinda when you have your write-up ready, can you publish the text and relevant data here in this Issue? I want to make sure the community can view our findings and has an opportunity to review our plan.

csuwildcat:
@arvinda can you post about the status of the per-op fee implemention you've written?

mudiali:
The algorithm part is done. We are going to use the issue #25 to track rest of the changes for the proof of fee.

PR #322, opened 2019-10-30
Add the core side work for the proof of fee work

mudiali:
#216. This PR is to add the sidetree core side changes for the proof of fee work. After this change, we will still need to integrate with the actual fee calculation work in PR #315.

### DIFF
--- a/docs/implementation.md
+++ b/docs/implementation.md
@@ -38,7 +38,7 @@ The anchor string is the data that is stored on the blockchain. The data is stor
 
 WHERE
 
-encoded_number_of_operations: The total number of operations included in the batch file converted to 4 bytes and encoded as Base64 URL string
+encoded_number_of_operations: The total number of operations included in the batch file converted to 4 bytes (in little endian format) and then encoded as Base64 URL string
 hash_of_batch_file: The hash of the batch file
 ```
 
--- a/docs/implementation.md
+++ b/docs/implementation.md
@@ -30,25 +30,6 @@ The primary goals for the _Observer_ are to:
 
 The above goals lead to the design decision of minimal processing of the operations at the time of ingestion, and deferring the heavy processing such as signature validations to the time of DID resolution.
 
-## Anchor String Schema
-The anchor string is the data that is stored on the blockchain. The data is stored in the following format:
-
-```
-[encoded_number_of_operations].[hash_of_batch_file]
-
-WHERE
-
-encoded_number_of_operations: The total number of operations included in the batch file converted to 4 bytes (in little endian format) and then encoded as Base64 URL string
-hash_of_batch_file: The hash of the batch file
-```
-
-### Example
-The following anchor string encodes 10000 operations and the hash of the batch file.
-
-```
-ECcAAA.c8957a1a204074f556a13049b2de287db980640167fb6ebb030766471a4cb33d
-```
-
 ## Versioning
 As the Sidetree protocol evolves, existing nodes executing an earlier version of the protocol need to upgrade to execute the newer version of the protocol while remaining backward compatible to processing of prior transactions and operations.
 
--- a/docs/implementation.md
+++ b/docs/implementation.md
@@ -217,7 +217,7 @@ GET /transactions?since=170&transaction-time-hash=00000000000000000000100158f474
       "transactionTime": "The logical blockchain time this transaction is anchored. Used for protocol version selection.",
       "transactionTimeHash": "The hash associated with the transaction time.",
       "anchorString": "The string written to the blockchain for this transaction.",
-      "feePaid": "A number representing the fee paid for this transaction.",
+      "transactionFeePaid": "A number representing the fee paid for this transaction.",
       "normalizedTransactionFee": "A number representing the normalized transaction fee used for proof-of-fee calculation."
     },
     ...
@@ -237,7 +237,7 @@ HTTP/1.1 200 OK
       "transactionTime": 545236,
       "transactionTimeHash": "0000000000000000002352597f8ec45c56ad19994808e982f5868c5ff6cfef2e",
       "anchorString": "QmWd5PH6vyRH5kMdzZRPBnf952dbR4av3Bd7B2wBqMaAcf",
-      "feePaid": 40000,
+      "transactionFeePaid": 40000,
       "normalizedTransactionFee": 100
     },
     {
@@ -245,7 +245,7 @@ HTTP/1.1 200 OK
       "transactionTime": 545236,
       "transactionTimeHash": "00000000000000000000100158f474719e5a319933856f7f464fcc65a3cb2253",
       "anchorString": "QmbJGU4wNti6vNMGMosXaHbeMHGu9PkAUZtVBb2s2Vyq5d",
-      "feePaid": 600000,
+      "transactionFeePaid": 600000,
       "normalizedTransactionFee": 400
     }
   ]
@@ -286,7 +286,7 @@ POST /transactions/firstValid HTTP/1.1
       "transactionTime": "The logical blockchain time this transaction is anchored. Used for protocol version selection.",
       "transactionTimeHash": "The hash associated with the transaction time.",
       "anchorString": "The string written to the blockchain for this transaction.",
-      "feePaid": "A number representing the fee paid for this transaction.",
+      "transactionFeePaid": "A number representing the fee paid for this transaction.",
       "normalizedTransactionFee": "A number representing the normalized transaction fee used for proof-of-fee calculation."
     },
     ...
@@ -306,7 +306,7 @@ Content-Type: application/json
       "transactionTime": 545236,
       "transactionTimeHash": "0000000000000000002352597f8ec45c56ad19994808e982f5868c5ff6cfef2e",
       "anchorString": "Qm28BKV9iiM1ZNzMsi3HbDRHDPK5U2DEhKpCYhKk83UPEg",
-      "feePaid": 5000,
+      "transactionFeePaid": 5000,
       "normalizedTransactionFee": 100
     },
     {
@@ -314,7 +314,7 @@ Content-Type: application/json
       "transactionTime": 545236,
       "transactionTimeHash": "0000000000000000000054f9719ef6ca646e2503a9c5caac1c6ea95ffb4af587",
       "anchorString": "Qmb2wxUwvEpspKXU4QNxwYQLGS2gfsAuAE9LPcn5LprS1nb",
-      "feePaid": 30,
+      "transactionFeePaid": 30,
       "normalizedTransactionFee": 10
 
     },
@@ -323,7 +323,7 @@ Content-Type: application/json
       "transactionTime": 545200,
       "transactionTimeHash": "0000000000000000000f32c84291a3305ad9e5e162d8cc363420831ecd0e2800",
       "anchorString": "QmbBPdjWSdJoQGHbZDvPqHxWqqeKUdzBwMTMjJGeWyUkEzK",
-      "feePaid": 50000,
+      "transactionFeePaid": 50000,
       "normalizedTransactionFee": 150
     },
     {
@@ -331,7 +331,7 @@ Content-Type: application/json
       "transactionTime": 545003,
       "transactionTimeHash": "0000000000000000001e002080595267fe034d370897b7b506d119ad29da1541",
       "anchorString": "Qmss3gKdm9uU9YLx3MPRHQTcUq1CR1Xv9Zpdu7EBG9Pk9Y",
-      "feePaid": 1000000,
+      "transactionFeePaid": 1000000,
       "normalizedTransactionFee": 200
     },
     {
@@ -339,7 +339,7 @@ Content-Type: application/json
       "transactionTime": 544939,
       "transactionTimeHash": "00000000000000000000100158f474719e5a319933856f7f464fcc65a3cb2253",
       "anchorString": "QmdcDrVPWy3ZXoZcuvFq7fDVqatks22MMqPAxDqXsZzGhy"
-      "feePaid": 100,
+      "transactionFeePaid": 100,
       "normalizedTransactionFee": 50
     }
   ]
@@ -353,7 +353,7 @@ Content-Type: application/json
   "transactionTime": "The logical blockchain time this transaction is anchored. Used for protocol version selection.",
   "transactionTimeHash": "The hash associated with the transaction time.",
   "anchorString": "The string written to the blockchain for this transaction.",
-  "feePaid": "A number representing the fee paid for this transaction.",
+  "transactionFeePaid": "A number representing the fee paid for this transaction.",
   "normalizedTransactionFee": "A number representing the normalized transaction fee used for proof-of-fee calculation."
 }
 ```
@@ -367,7 +367,7 @@ HTTP/1.1 200 OK
   "transactionTime": 545200,
   "transactionTimeHash": "0000000000000000000f32c84291a3305ad9e5e162d8cc363420831ecd0e2800",
   "anchorString": "QmbBPdjWSdJoQGHbZDvPqHxWqqeKUdzBwMTMjJGeWyUkEzK",
-  "feePaid": 50000,
+  "transactionFeePaid": 50000,
   "normalizedTransactionFee": 50
 }
 ```
--- a/docs/implementation.md
+++ b/docs/implementation.md
@@ -217,7 +217,8 @@ GET /transactions?since=170&transaction-time-hash=00000000000000000000100158f474
       "transactionTime": "The logical blockchain time this transaction is anchored. Used for protocol version selection.",
       "transactionTimeHash": "The hash associated with the transaction time.",
       "anchorString": "The string written to the blockchain for this transaction.",
-      "feePaid": "A number representing the fee paid for this transaction."
+      "feePaid": "A number representing the fee paid for this transaction.",
+      "normalizedTransactionFee": "A number representing the normalized transaction fee used for proof-of-fee calculation."
     },
     ...
   ]
@@ -236,14 +237,16 @@ HTTP/1.1 200 OK
       "transactionTime": 545236,
       "transactionTimeHash": "0000000000000000002352597f8ec45c56ad19994808e982f5868c5ff6cfef2e",
       "anchorString": "QmWd5PH6vyRH5kMdzZRPBnf952dbR4av3Bd7B2wBqMaAcf",
-      "feePaid": 40000
+      "feePaid": 40000,
+      "normalizedTransactionFee": 100
     },
     {
       "transactionNumber": 100,
       "transactionTime": 545236,
       "transactionTimeHash": "00000000000000000000100158f474719e5a319933856f7f464fcc65a3cb2253",
       "anchorString": "QmbJGU4wNti6vNMGMosXaHbeMHGu9PkAUZtVBb2s2Vyq5d",
-      "feePaid": 600000
+      "feePaid": 600000,
+      "normalizedTransactionFee": 400
     }
   ]
 }
@@ -283,7 +286,8 @@ POST /transactions/firstValid HTTP/1.1
       "transactionTime": "The logical blockchain time this transaction is anchored. Used for protocol version selection.",
       "transactionTimeHash": "The hash associated with the transaction time.",
       "anchorString": "The string written to the blockchain for this transaction.",
-      "feePaid": "A number representing the fee paid for this transaction."
+      "feePaid": "A number representing the fee paid for this transaction.",
+      "normalizedTransactionFee": "A number representing the normalized transaction fee used for proof-of-fee calculation."
     },
     ...
   ]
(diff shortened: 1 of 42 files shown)


lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -564,7 +563,9 @@ export default class BitcoinProcessor {
           transactionNumber: TransactionNumber.construct(transactionBlock, transactionIndex),
           transactionTime: transactionBlock,
           transactionTimeHash: transactionHash,
-          anchorString: data.slice(this.sidetreePrefix.length)
+          anchorString: data.slice(this.sidetreePrefix.length),
+          feePaid: 100, // issue #216: read actual data from tx
```
thehenrytsai: I don't think #216 will give the fee paid, but we can create another issue to check this in separately if you want.

lib/bitcoin/IBitcoinConfig.ts
```diff
@@ -6,7 +6,6 @@ export interface IBitcoinConfig {
   bitcoinRpcUsername: string | undefined;
   bitcoinRpcPassword: string | undefined;
   bitcoinWalletImportString: string;
-  bitcoinFee: number;
```
thehenrytsai: Need to also remove all other references of `bitcoinFee`, a quick repo search shows at least 2 more location requiring removal (`README.md` & `setup.sh`).

lib/common/MongoDbTransactionStore.ts
```diff
@@ -89,7 +89,9 @@ export default class MongoDbTransactionStore implements ITransactionStore {
         // NOTE: MUST force `transactionNumber` to be Int64 in MondoDB.
         transactionNumber: Long.fromNumber(transaction.transactionNumber),
         transactionTime: transaction.transactionTime,
-        transactionTimeHash: transaction.transactionTimeHash
+        transactionTimeHash: transaction.transactionTimeHash,
```
thehenrytsai: Reminder: Update protocol & implementation documentation.

lib/core/interfaces/IBlockchain.ts
```diff
@@ -34,4 +36,12 @@ export default interface IBlockchain {
    * Useful for cases where high performance is desired and hgih accuracy is not required.
    */
   approximateTime: BlockchainTimeModel;
+
+  /**
+   * Fetches the normalized transaction fee used for proof-of-fee calculation, given the blockchain time.
+   * @param transactionNumber A valid Sidetree transaction number.
+   *
+   * @throws SidetreeError with ErrorCode.InvalidTransactionNumberOrTimeHash if a potential block reorganization is detected.
```
thehenrytsai: Sanity: did you mean `ErrorCode.BlockchainTimeOutOfRange`?

lib/core/Blockchain.ts
```diff
@@ -161,4 +164,33 @@ export default class Blockchain implements IBlockchain {
...
     return responseBody;
   }
+
+  /**
+   * Fetches the normalized transaction fee used for proof-of-fee calculation, given the blockchain time.
+   * @param transactionNumber A valid Sidetree transaction number.
+   *
+   * @throws SidetreeError with ErrorCode.BlockchainTimeOutOfRange if the input transaction number is less
+   * than Sidetree genesis blockchain time or is later than the current blockchain time.
+   */
```
thehenrytsai: Consider only have the interface method documentation at the interface definition, `IBlockchain.ts` in this case.

lib/core/Blockchain.ts
```diff
@@ -161,4 +164,33 @@ export default class Blockchain implements IBlockchain {
...
+
+    if (response.status === HttpStatus.BAD_REQUEST &&
+        responseBody.code === SharedErrorCode.BlockchainTimeOutOfRange) {
+      throw new SidetreeError(SharedErrorCode.BlockchainTimeOutOfRange);
+    }
+
+    if (response.status !== HttpStatus.OK) {
+      console.error(`Blockchain read error response status: ${response.status}`);
+      console.error(`Blockchain read error body: ${responseBodyString}`);
+      throw new SidetreeError(CoreErrorCode.BlockchainReadResponseNotOk);
```
thehenrytsai: Design: Rule of thumb adopted here is to not reuse error code. Consider creating a new error code.

lib/core/Error.ts
```diff
@@ -4,10 +4,20 @@
...
   }
+
+  /**
+   * Returns a new SidetreeError object using the inputs.
+   *
+   * @param code The error code.
+   * @param err The error exception thrown.
+   */
+  public static createFromError (code: string, err: Error): SidetreeError {
+    return new SidetreeError(code, err ? err.message : undefined);
```
thehenrytsai: Sanity: it seems not possible for `err` to be `undefined`?

lib/core/models/Config.ts
```diff
@@ -4,6 +4,7 @@
 export default interface Config {
   batchingIntervalInSeconds: number;
   blockchainServiceUri: string;
+  blockchainFeeMarkupFactor: number;
```
thehenrytsai: Nit:`transactionFeeMarkup...`.

Optional: consider `...FeeMarkupPercentage` as "percentage" is be more easy to understand and there is less interpretation than "factor", obviously you'll need to divide by 100... so I understand if push back.

lib/core/versions/latest/FeeManager.ts
```diff
@@ -0,0 +1,57 @@
...
+   * @throws if the number of operations are <= 0.
+   */
+  public static convertNormalizedFeeToTransactionFee (normalizedFee: number, numberOfOperations: number, feeMarkupFactor: number): number {
+
+    if (numberOfOperations <= 0) {
+      throw new SidetreeError(ErrorCode.OperationCountLessThanZero, `Fee cannot be calculated for the given number of operations: ${numberOfOperations}`);
+    }
+
+    const normalizedFeePerOperation = normalizedFee * ProtocolParameters.normalizedToPerOperationFeeFactor;
+    const perOperationFee = normalizedFeePerOperation * numberOfOperations;
```
thehenrytsai: `perOperationFee` sounds like it is holding the "fee per operation", which is obviously not true.
```suggestion
    const minimumTransactionFee= normalizedFeePerOperation * numberOfOperations;
```

lib/core/versions/latest/models/ProtocolParameters.ts
```diff
@@ -14,4 +14,6 @@ export default interface ProtocolParameters {
   maxOperationsPerBatch: number;
   /** Maximum size of an operation in bytes. */
   maxOperationByteSize: number;
+  /** The factor that converts the normalized fee from blockchain into a 'per operation' fee. */
+  normalizedToPerOperationFeeFactor: number;
```
thehenrytsai: Nit: it's long but consider renaming to `normalizedFeeToPerOperationFeeMultiplier`

lib/core/versions/0.5.0/BatchWriter.ts
```diff
@@ -63,7 +63,13 @@ export default class BatchWriter implements IBatchWriter {
     console.info(`Wrote anchor file ${anchorFileAddress} to content addressable store.`);
 
     // Anchor the 'anchor file hash' on blockchain.
-    await this.blockchain.write(anchorFileAddress);
+    //
+    // NOTE: We are passing 0 for the fee parameter in this version. The fee parameter was
+    // introduced after this version had shipped. This 'write' function is only ever called
+    // for the latest blockchain time which means that this previous version will never be
+    // called and passing a fake value like 0 is ok.
```
thehenrytsai: Nit: just want to call out that technically, the last/previous version of the `BatchWriter` can be active for a period of time before the blockchain time crosses over to the next version.

lib/core/versions/latest/AnchoredDataSerializer.ts
```diff
@@ -0,0 +1,95 @@
...
+import ErrorCode from './ErrorCode';
+import { SidetreeError } from '../../Error';
+
+/**
+ * Encapsulates functionality to serialize/deserialize data that read/write to
+ * the blockchain.
+ */
+export default class AnchoredDataSerializer {
+
+  private static readonly delimeter = '.';
```
thehenrytsai: Typo.
```suggestion
  private static readonly delimiter = '.';
```

lib/core/versions/latest/AnchoredDataSerializer.ts
```diff
@@ -0,0 +1,95 @@
...
+   * @param numberOfOperations The number of operations to write.
+   */
+  public static serialize (dataToBeAnchored: AnchoredData): string {
+
+    // First convert the number of operations input into a 3-byte buffer and then base64 encode it
+    const numberAsBuffer = AnchoredDataSerializer.convertToThreeBytesBuffer(dataToBeAnchored.numberOfOperations);
+    const encodedNumberOfOperations = Encoder.encode(numberAsBuffer);
+
+    // Concatenate the inputs w/ the delimeter and return
+    return `${encodedNumberOfOperations}${AnchoredDataSerializer.delimeter}${dataToBeAnchored.anchorFileHash}`;
```
thehenrytsai: Can we add a section in the `implementation.md` on anchor string schema.

lib/core/versions/latest/AnchoredDataSerializer.ts
```diff
@@ -0,0 +1,95 @@
...
+export default class AnchoredDataSerializer {
+
+  private static readonly delimeter = '.';
+  private static readonly maxInThreeBytes = 0xFFFFFF;
+
+  /**
+   * Converts the given inputs to the string that is to be written to the blockchain.
+   *
+   * @param anchorFileAddress The anchor file address to write.
+   * @param numberOfOperations The number of operations to write.
```
thehenrytsai: Obsolete. Remove.

lib/core/versions/latest/AnchoredDataSerializer.ts
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    // First write the input into a 4 bytes buffer. Big Endian format.
+    // Example
+    //  input: 1000 will be converted to: [0, 0, 3, 232]
+    const byteArrayBuffer = Buffer.alloc(4);
+    byteArrayBuffer.writeUInt32BE(numberOfOperations, 0);
+
+    // Remove the first byte (because of Big Endian) and return the rest
+    return byteArrayBuffer.slice(1, 4);
```
thehenrytsai: Sanity: would allocate only byte size of 3 not yield the same result?

lib/core/versions/latest/AnchoredDataSerializer.ts
```diff
@@ -0,0 +1,95 @@
...
+    return {
+      anchorFileHash: splitData[1],
+      numberOfOperations: numberOfOperations
+    };
+  }
+
+  private static convertToThreeBytesBuffer (numberOfOperations: number): Buffer {
+
+    if (numberOfOperations < 0) {
+      // We are going to write the number as unint32
```
thehenrytsai: Consider removing this line. I think it is misleading as we will write as uint24 (if there is such a thing).

lib/core/versions/latest/AnchoredDataSerializer.ts
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    if (numberOfOperations > this.maxInThreeBytes) {
+      // We are only using 3 bytes to store the number of operations so any number greater than
+      // that is not allowed.
+      throw new SidetreeError(ErrorCode.AnchoredDataNumberOfOperationsGreaterThanMax,
+                              `Number of operations ${numberOfOperations} must be less than equal to ${this.maxInThreeBytes}`);
+    }
+
+    // First write the input into a 4 bytes buffer. Big Endian format.
```
thehenrytsai: Nit: maybe we should consider using little-endian instead because I believe that is more conventional (e.g. Intel and AMD chips).

lib/core/Blockchain.ts
mudiali: Oops .. thought that the code was generic. Good catch.

lib/core/versions/latest/FeeManager.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    // Add some markup to the fee as defined by the caller.
+    return perOperationFee + (feeMarkupFactor * perOperationFee);
```
thehenrytsai: Need to return the normalized fee if the final fee is less than the normalized fee.

tests/JasmineHelper.ts
```diff
@@ -0,0 +1,51 @@
...
+ */
+export default class JasmineHelper {
+
+  /**
+   * Fails the current spec if the execution of the function does not throw the expected SidetreeError.
+   *
+   * @param functionToExcute The function to execute.
+   * @param expectedSidetreeError The expected error.
+   */
+  public static expectSideTreeErrorToBeThrown (functionToExcute: () => any, expectedSidetreeError: SidetreeError): void {
```
thehenrytsai:
```suggestion
  public static expectSidetreeErrorToBeThrown (functionToExcute: () => any, expectedSidetreeError: SidetreeError): void {
```

tests/JasmineHelper.ts
```diff
@@ -0,0 +1,51 @@
...
+    }
+  }
+
+  /**
+   * Fails the current spec if the execution of the function does not throw the expected SidetreeError.
+   *
+   * @param functionToExcute The function to execute.
+   * @param expectedSidetreeError The expected error.
+   */
+  public static async expectSideTreeErrorToBeThrownAsync (functionToExcute: () => Promise<any>, expectedSidetreeError: SidetreeError): Promise<void> {
```
thehenrytsai:
```suggestion
  public static async expectSidetreeErrorToBeThrownAsync (functionToExcute: () => Promise<any>, expectedSidetreeError: SidetreeError): Promise<void> {
```

lib/core/versions/latest/FeeManager.ts
```diff
@@ -0,0 +1,57 @@
...
+   * @param normalizedFee The normalized fee for that transaction.
+   *
+   * @throws if the number of operations is <= 0; if the feepaid is invalid.
+   */
+  public static verifyTransactionFeeAndThrowOnError (feePaid: number, numberOfOperations: number, normalizedFee: number): void {
+
+    // If there are no operations written then someone wrote incorrect data and we are going to throw
+    if (numberOfOperations <= 0) {
+      throw new SidetreeError(ErrorCode.OperationCountLessThanZero, `The number of operations: ${numberOfOperations} must be greater than 0`);
+    }
```
thehenrytsai: Consider moving this check to `AnchorDataSerializer.deserialize()`

tests/bitcoin/BitcoinProcessor.spec.ts
```diff
@@ -85,12 +84,16 @@ describe('BitcoinProcessor', () => {
       height = randomNumber();
     }
     const hash = randomString();
+    const feePaidRandom = randomNumber();
```
thehenrytsai: Design: I do not think `randomNumber()` should have a default max unless the default is MAX_SAFE_INTEGER, not asking to change existing tests, but can we explicitly specify the max for all new `randomNumber()` call?

tests/JasmineHelper.ts
```diff
@@ -0,0 +1,51 @@
+import { SidetreeError } from '../lib/core/Error';
+
+/**
+ * Encapsulates the helper functions for the tests.
+ */
+export default class JasmineHelper {
```
thehenrytsai: Not a fan of the name, perhaps:
```suggestion
export default class SidetreeErrorValidator {
```

lib/core/models/Config.ts
mudiali: Will change this to 'transactionFeeMarkupPercentage'. Also change the 'feePaid' returned by the blockchain service to 'transactionFeePaid.'

lib/core/versions/latest/AnchoredDataSerializer.ts
mudiali: Actually will change it to use the normal 4 bytes.

lib/core/versions/latest/FeeManager.ts
mudiali: Want to keep the serialization and validation logic separate so keeping this the way it is.

lib/core/Error.ts
mudiali: Could be undefined at runtime. For example if the Promise is rejected without an error (which is a bug that I found). So keeping this.

tests/bitcoin/BitcoinProcessor.spec.ts
mudiali: Changing the further usage of the feePaidRandom to not use -1 as to remove any confusion.

tests/JasmineHelper.ts
mudiali: There's still Jasmine related code in here so I'm going to rename but keep Jasmine in the name.

docs/implementation.md
```diff
@@ -30,6 +30,25 @@ The primary goals for the _Observer_ are to:
...
+
+encoded_number_of_operations: The total number of operations included in the batch file converted to 4 bytes (in little endian format) and then encoded as Base64 URL string
+hash_of_batch_file: The hash of the batch file
+```
+
+### Example
+The following anchor string encodes 10000 operations and the hash of the batch file.
+
+```
+ECcAAA.c8957a1a204074f556a13049b2de287db980640167fb6ebb030766471a4cb33d
```
thehenrytsai: Sanity: Isn't the hash portion in Base64 URL encoding also?

lib/core/versions/latest/AnchoredDataSerializer.ts
```diff
@@ -35,61 +34,53 @@ export default class AnchoredDataSerializer {
...
+    const numberOfOperations = AnchoredDataSerializer.convertFromBytesBuffer(decodedNumberOfOperations);
 
     return {
       anchorFileHash: splitData[1],
       numberOfOperations: numberOfOperations
     };
   }
 
-  private static convertToThreeBytesBuffer (numberOfOperations: number): Buffer {
+  private static convertToBytesBuffer (numberOfOperations: number): Buffer {
```
thehenrytsai: Nit: Consider just `convertToBuffer()` or `convertNumberToBuffer()`.

lib/core/versions/latest/AnchoredDataSerializer.ts
```diff
@@ -35,61 +34,53 @@ export default class AnchoredDataSerializer {
...
+  private static convertFromBytesBuffer (bytesBuffer: Buffer): number {
 
-    // Ensure that the input has 3 bytes
-    if (threeBytesBuffer.length !== 3) {
-      throw new SidetreeError(ErrorCode.AnchoredDataNumberOfOperationsNotThreeBytes,
-                              `Input must have 3 bytes but have ${threeBytesBuffer.length} bytes instead.`);
+    // Ensure that the input has 4 bytes
+    if (bytesBuffer.length !== 4) {
+      throw new SidetreeError(ErrorCode.AnchoredDataNumberOfOperationsNotFourBytes,
+                              `Input must have 3 bytes but have ${bytesBuffer.length} bytes instead.`);
```
thehenrytsai: 4 bytes

lib/core/versions/latest/AnchoredDataSerializer.ts
```diff
@@ -35,61 +34,53 @@ export default class AnchoredDataSerializer {
...
-  private static convertFromThreeBytesBuffer (threeBytesBuffer: Buffer): number {
+  private static convertFromBytesBuffer (bytesBuffer: Buffer): number {
 
-    // Ensure that the input has 3 bytes
-    if (threeBytesBuffer.length !== 3) {
-      throw new SidetreeError(ErrorCode.AnchoredDataNumberOfOperationsNotThreeBytes,
-                              `Input must have 3 bytes but have ${threeBytesBuffer.length} bytes instead.`);
+    // Ensure that the input has 4 bytes
+    if (bytesBuffer.length !== 4) {
+      throw new SidetreeError(ErrorCode.AnchoredDataNumberOfOperationsNotFourBytes,
```
thehenrytsai: This line is missing in code coverage, would be nice to add a quick one to make the file 100% once and for all.

lib/core/versions/latest/FeeManager.ts
```diff
@@ -17,41 +17,54 @@ export default class FeeManager {
    *
    * @throws if the number of operations are <= 0.
    */
-  public static convertNormalizedFeeToTransactionFee (normalizedFee: number, numberOfOperations: number, feeMarkupFactor: number): number {
+  public static convertNormalizedFeeToTransactionFee (normalizedFee: number, numberOfOperations: number, feeMarkupPercentage: number): number {
```
thehenrytsai: Optional: consider `computeTransactionFee()` as `normalizedFee` is technically just one of 3 parameters used for the computation.

lib/core/versions/latest/FeeManager.ts
```diff
@@ -17,41 +17,54 @@ export default class FeeManager {
...
 
     if (numberOfOperations <= 0) {
       throw new SidetreeError(ErrorCode.OperationCountLessThanZero, `Fee cannot be calculated for the given number of operations: ${numberOfOperations}`);
     }
 
-    const normalizedFeePerOperation = normalizedFee * ProtocolParameters.normalizedToPerOperationFeeFactor;
-    const perOperationFee = normalizedFeePerOperation * numberOfOperations;
+    const normalizedFeePerOperation = normalizedFee * ProtocolParameters.normalizedFeeToPerOperationFeeMultiplier;
+    const normalizedFeeForAllOperations = normalizedFeePerOperation * numberOfOperations;
```
thehenrytsai: Nit & optional: consider just `feePerOperation` and `feeForAllOperations` as I don't see these fee being "normalized".

docs/implementation.md
mudiali: I don't think so.

PR #323, opened 2019-11-05
#165 Added controller as a property of the publickey 

mudiali:
Added controller as a property of the publickey but only the sidetree node adds an actual value to it. Adding by client is not allowed. #165

### DIFF
--- a/lib/core/versions/latest/Operation.ts
+++ b/lib/core/versions/latest/Operation.ts
@@ -1,5 +1,5 @@
 import Did from './Did';
-import DidPublicKeyModel from '../latest/models/DidPublicKeyModel';
+import DidPublicKeyModel from './models/DidPublicKeyModel';
 import Document from './Document';
 import DocumentModel from './models/DocumentModel';
 import Encoder from './Encoder';
@@ -340,6 +340,10 @@ export default class Operation {
 
       if (typeof publicKey.id !== 'string') {
         throw new SidetreeError(ErrorCode.OperationUpdatePatchPublicKeyIdNotString);
+      }
+
+      if (typeof publicKey.controller === 'string') {
+        throw new SidetreeError(ErrorCode.OperationUpdatePatchPublicKeyControllerNotAllowed);
       }
 
       if (publicKey.type === 'Secp256k1VerificationKey2018') {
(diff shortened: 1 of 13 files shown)


lib/core/versions/latest/Operation.ts
```diff
@@ -342,6 +342,10 @@ export default class Operation {
         throw new SidetreeError(ErrorCode.OperationUpdatePatchPublicKeyIdNotString);
       }
 
+      if (typeof publicKey.controller === 'string') {
```
thehenrytsai:
```suggestion
      if (typeof publicKey.controller !== undefined) {
```

lib/core/versions/latest/OperationProcessor.ts
```diff
@@ -97,6 +97,11 @@ export default class OperationProcessor implements IOperationProcessor {
         }
 
         AnchoredOperation.applyPatchesToDidDocument(didDocument, operation.patches!);
+
+        // Update the document with the actual id .. for example, the patches may add some fields that may not
+        // have the 'id' fields set.
+        Document.updatePlaceholdersInDocumentWithDid(didDocument, didDocument.id);
```
thehenrytsai: Design: I feel it is better to hide/perform this logic behind `applyPatchesToDidDocument()`, and to that point, I think a good place actually do this would be within the `applyPatchToDidDocument()` (note NOT `applyPatchesToDidDocument()`) directly under handling of `add-public-keys`.

lib/core/versions/latest/models/DidPublicKeyModel.ts
```diff
@@ -4,6 +4,7 @@
 export default interface DidPublicKeyModel {
   id: string;
   type: string;
+  controller: string;
```
thehenrytsai: Design: I am contemplating if we should make this property optional so it can be used in both operation request, and resolution response.

lib/core/versions/latest/Operation.ts
mudiali: Lint is pointing out that the suggestion ```typeof publicKey.controller !== undefined``` is always true so I'm going to change it to check whether the 'controller' property is defined in that section at all or not.

lib/core/versions/latest/models/DidPublicKeyModel.ts
mudiali: Leaving this the way it is as there's no need at this moment. The current tests are working because the input is cast'ed at runtime and the controller not present is not a problem.

lib/core/versions/latest/Operation.ts
thehenrytsai: Yeah... that's what I mean, was a typo.

PR #327, opened 2019-11-14
#165 Create should block the controller field in the publickey element.

mudiali:
Fixing the bug where the did document could have the controller field in the publickey element.

### DIFF
--- a/lib/core/versions/latest/Document.ts
+++ b/lib/core/versions/latest/Document.ts
@@ -20,13 +20,13 @@ export default class Document {
     const decodedJsonString = Encoder.decodeAsString(encodedOriginalDidDocument);
     const decodedDidDocument = JSON.parse(decodedJsonString);
 
-    // Replace the placeholder DID with real DID before returning it.
-    Document.updatePlaceholdersInDocumentWithDid(decodedDidDocument, did);
-
     // Return `undefined` if original DID Document is invalid.
     if (!Document.isObjectValidOriginalDocument(decodedDidDocument)) {
       return undefined;
     }
+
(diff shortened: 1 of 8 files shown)


lib/core/versions/latest/Document.ts
```diff
@@ -95,6 +95,11 @@ export default class Document {
       } else {
         keyUsages.set(publicKeyEntry.usage, usageCount + 1);
       }
+
+      // Controller field is not allowed to be filled in by the client
+      if (publicKeyEntry.controller !== undefined) {
```
OR13: I'm not sure if this example is wrong... https://w3c.github.io/did-core/#example-8-various-public-keys

But I recall some discussion about controller flexibility...

https://github.com/w3c/did-core/issues/4

OR13: I don't see the point of having a controller field in the public key if its always the same as the DID Document ID, and if a client can add a public key object to their DID, I believe the client should be able to specify the controller, and its not the responsibility of sidetree to decide if the controller is correct.

OR13: I could be wrong, though... I find this part of the DID Spec very confusing.

mudiali: Agree with you @OR13 that there's no point in having the controller field if it's always the same as the DID Document ID BUT the spec says that this is a mandatory field and so we are not allowing the client to add anything in there. @csuwildcat has commented on the related issue (#165) that the publickey section is really about the resolved documents. I am having a discussion with him there whether we need any updates to the spec or not. In any case if the specs are updated with some other requirement about the controller field (removing it completely for example) then we can update our implementation.

PR #332, opened 2019-11-22
#216 Updating the proof of fee changes to address leftover issues

mudiali:
Following changes are in this PR:
1. The ```feeApproximation``` protocol parameter has been removed. Instead the code now uses the hardcoded value.
1. Renamed the ```lastSeenBlock``` to ```lastProcessedBlock``` in ```BitcoinProcessor.ts```.
1. The ```lastProcessedBlock``` is now only being updated in the ```processTransactions``` function. This ensures that the status of this variable is up to date even during the revert process.
1. The ```getNormalizedFee``` function now checks against the ```lastProcessedBlock``` to ensure that the correct quantile value can be returned.

### DIFF
--- a/lib/bitcoin/BitcoinProcessor.ts
+++ b/lib/bitcoin/BitcoinProcessor.ts
@@ -77,7 +77,7 @@ export default class BitcoinProcessor {
   public lowBalanceNoticeDays: number;
 
   /** Last seen block */
-  private lastSeenBlock: IBlockInfo | undefined;
+  private lastProcessedBlock: IBlockInfo | undefined;
 
   /** Poll timeout identifier */
   private pollTimeoutId: number | undefined;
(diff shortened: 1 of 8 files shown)


lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -325,7 +324,7 @@ export default class BitcoinProcessor {
       throw new RequestError(ResponseStatus.BadRequest, ErrorCode.BlockchainTimeOutOfRange);
     }
 
-    const currentBlockNumber = await this.getCurrentBlockHeight();
+    const currentBlockNumber = this.lastProcessedBlock ? this.lastProcessedBlock.height : await this.getCurrentBlockHeight();
```
thehenrytsai: Consider just assigning to `this.lastProcessedBlock`, since `initialize()` would have already been called.

Issue #256, opened 2019-07-03
Revisit resolution using Initial DID Document


thehenrytsai:
Currently resolution using initial DID Document uses the current blockchain time for deciding hash algorithm to use for generating the DID.

However this implies that either:
* This resolution method is only intended for bootstrapping during the initial registration phase, thus this resolution method cannot be reliably used long term because new hashing algorithms can be introduced; or
* We need to have a reliable method for selecting hashing algorithm for this type of resolution.

csuwildcat:
What about adding leading bytes before the DID Suffix portion of the string?

PR #334, opened 2019-11-23
Issue #256 - Re-implemented long-form DID resolution based on new spec

thehenrytsai:


tests/core/Did.spec.ts
```diff
@@ -1,8 +1,57 @@
...
+    });
+
+    it('should throw if DID given does not match the expected DID method name.', async () => {
+      JasmineSidetreeErrorValidator.expectSidetreeErrorToBeThrown(
+        () => Did.create('did:sidetree:EiAgE-q5cRcn4JHh8ETJGKqaJv1z2OgjmN3N-APx0aAvHg', 'did:sidetree2:'),
+        ErrorCode.DidIncorrectPrefix
+      );
+    });
+
+    it('should throw if DID given does not match the expected DID method name.', async () => {
```
mudiali: Do you mean 'should throw if DID does not have a unique suffix' or something?

tests/core/Did.spec.ts
```diff
@@ -1,8 +1,57 @@
...
+      expect(did.uniqueSuffix).toEqual(uniqueSuffix);
+    });
+
+    it('should throw if DID given does not match the expected DID method name.', async () => {
+      JasmineSidetreeErrorValidator.expectSidetreeErrorToBeThrown(
+        () => Did.create('did:sidetree:EiAgE-q5cRcn4JHh8ETJGKqaJv1z2OgjmN3N-APx0aAvHg', 'did:sidetree2:'),
+        ErrorCode.DidIncorrectPrefix
+      );
+    });
+
```
mudiali: Suggestion: do you think we need a test for when the DID is longform?

lib/core/versions/latest/Document.ts
```diff
@@ -26,7 +28,47 @@ export default class Document {
...
+    let decodedJsonString;
+    try {
+      decodedJsonString = Encoder.decodeAsString(encodedOriginalDidDocument);
+    } catch (error) {
+      throw SidetreeError.createFromError(ErrorCode.DocumentIncorretEncodedFormat, error);
+    }
+
+    let decodedDidDocument;
+    try {
+      decodedDidDocument = JSON.parse(decodedJsonString);
```
mudiali: JsonAsync.Parse?

lib/core/versions/latest/Did.ts
```diff
@@ -1,10 +1,89 @@
...
+      throw new SidetreeError(ErrorCode.DidNoUniqueSuffix);
+    }
+
+    this.shortForm = didMethodName + this.uniqueSuffix;
+
+    // Get the encoded document if it's long-form.
+    if (!this.isShortForm) {
+      const didParameterString = did.substring(indexOfSemiColonChar + 1);
+
+      if (!didParameterString.startsWith('initial-values=')) {
```
csuwildcat: Why is this a `startsWith` check? There's no prohibition on other matrix parameters being included, and they are not ordered such that `initial-values` will always be first.

tests/core/Did.spec.ts
thehenrytsai: Thanks. Good catch!

tests/core/Did.spec.ts
thehenrytsai: Added.

lib/core/versions/latest/Did.ts
thehenrytsai: What you are asking is supporting other DID parameters, which is currently not supported, thus only `initial-values` is allowed. If feature of supporting other DID parameters is a requirement and mainnet blocker, please file a issue.

csuwildcat:
I'm not asking to support other things in our code, but some matrix params
may be passed regardless of whether we care about them, and we should not
just fail based on their presence or order not placing `inital-values` at
the 0 index. Instead, we should parse the params and use the one,
`initial-values`, that we care about.

lib/core/versions/latest/Document.ts
thehenrytsai: Done.

thehenrytsai:
@csuwildcat, I get what you are saying. I am in the camp of "let's be strict about what parameters are "allowed" for Sidetree" for this PR. I am open to discuss what's the right thing to do, but unless you see the discussion on DID parameters is a blocker for this PR, I prefer a separate issue/thread for it.

csuwildcat:
I think this is a blocker because it's noncompliant with the spec. We
should not be introducing order-based requirements, given the DID Doc spec
does not provide for this concept. (not to mention it just makes our code
more brittle for no apparent value)

thehenrytsai:
@csuwildcat, are you saying that this PR must support all generic DID parameters?

csuwildcat:
No, I'm saying it should support parsing the parameters string in a way
that is order agnostic, and still only supporting (aka: doing something
with) the ones we care about. Doesn't the code currently assume no other
parameters will be present in the string and fail if the parameter we care
about isn't the startsWith 0 index?

thehenrytsai:
Yes, the code currently does not allow other parameters because supporting other parameters was not what this PR was about primarily. I would push back writing code parsing parameters if no other parameters is allowed.

csuwildcat:
I'm simply saying the code should split/match iterate on semicolons with a
regex and act on the param we are using (initial-values), regardless of
order or number of total params, of which we'll ignore besides the one we
currently care about.

thehenrytsai:
I don't think you are simply asking me to split on semi-colons because if there is no requirements to support others parameters why do it? The code currently is:

1. Meets all requirements (unless you tell me it doesn't, which is what I've been asking).
1. It is faster than what you are suggesting.
1. It is simpler than what you are suggesting.

So are you suggesting Sidetree must support other generic DID parameters? If so, what are they?

csuwildcat:
There's a big difference between supporting parameters and not having our
code fail because they're simply present in the string or our supported
parameter is at an index other than 0. Because general matrix parameters
could be present outside of the ones we support, you can't/shouldn't assume
they won't be there or at 0 index.

thehenrytsai:
Are you saying that Sidetree must accept other generic parameters even though they are not used? If so, why? Because this is essentially what our disagreement is.

thehenrytsai:
Issue #338 is opened to track the open questions regarding support for generic DID parameters as a mainnet blocking item. For now, only `initial-values` is allowed.

Issue #337, opened 2019-11-26
BitcoinProcessor::processBlock(...) shouldn't suppress all thrown exceptions


mudiali:
https://github.com/decentralized-identity/sidetree/blob/master/lib/bitcoin/BitcoinProcessor.ts#L700

Currently, the processBlock(...) routine suppresses all the exceptions thrown by the ```addValidSidetreeTransactionsFromVOutsToTransactionStore``` function. We should break on exceptions so that we can fix any bugs during that state. Another option could be to add this transaction to a 'retry' block and come back to it when ready. As a mainnet blocker, we should stop suppressing for now.

Issue #25, opened 2018-10-08
Implement Proof of Fee (batch size-relative fee over-payment scheme)


thehenrytsai:
Implement proof-of-work according to Sidetree protocol.

csuwildcat:
I'm seriously leaning towards a basic proof variant that's simply Proof of Bump - a simple algo that transaction anchoring nodes can use to pay more in fees for more ops. This is a far more benefitial strategy than Proof of Burn, and doesn't preclude later additions to operation-level proofing schemes.

OR13:
The first objective of the PoW is to prevent spamming, [here](https://cryptocurrencyhub.io/implementing-a-simple-proof-of-work-algorithm-for-the-blockchain-bdcd50faac18) is a simple article regarding a naive implementation, I think it would be helpful to name a simple naive approach and complete the integration and tests using that algo before attempting a more complicated one.

I don't understand the comment regarding fees. 

As I understand it, PoW in sidetree is only about spam protection, there is no block reward or transaction fees for the miner (batch writer).

Is there a relationship between the blockchain native tokens (BTC, ETH) and sidetree PoW? 

Or is the proof a pure function of `powAlgo`, `lastTransactionTimeHash `, `operationHash`, and `nonce` ?

Will the PoW algorithm be geared towards existing hardware / miners or is it desirable for it to be asic resistant?

See also: https://swende.se/blog/Progpow.html

https://github.com/ifdefelse/ProgPOW

^ this complexity seems like a terrible idea for sidetree.

https://lhartikk.github.io/jekyll/update/2017/07/13/chapter2.html

^ this seems like a much better starting point.

I keep thinking about the case where a client wants to anchor their own operations, it costs them ETH or BTC to do this as well as compute time for PoW, additionally if their operations are invalid this is wasted.

The case is the same for sidetree nodes, but they are handling many operations, and doing much more work, so there is more chance for their work to become invalid while they try to complete a batch.

Can we walk through a simple case with block difficulty?

csuwildcat:
PoW in Sidetree is just for spam, it's not tied to generating value or the
actual target chain's own security system.

The fee-burn idea, which I'm very seriously considering and will spec out
to evaluate, is basically the same as requiring writers of large batches to
engage in Proof of Burn, but instead you overpay txn fees in accordance
with the amount of value a deterministic algo requires you to expend for
the size of batch you're including.

Instead of just burning BTC or ETH into a void, as traditional Proof of
[...]

PR #341, opened 2019-12-04
#25 Integrated bitcoin service with bitcoin testnet

mudiali:
These changes ensure that the bitcoin service runs successfully against testnet after our proof of fee changes. Issue #337 is also being fixed as a result of this work.

### DIFF
--- a/lib/bitcoin/BitcoinProcessor.ts
+++ b/lib/bitcoin/BitcoinProcessor.ts
@@ -800,7 +800,10 @@ export default class BitcoinProcessor {
       const oneSidetreeTxAlreadyFound = (sidetreeTxToAdd !== undefined);
 
       if (isSidetreeTx && oneSidetreeTxAlreadyFound) {
-        throw new Error('The transaction has more then one sidetree anchor strings.');
+        // tslint:disable-next-line: max-line-length
+        const message = `The outputs in block: ${transactionBlock} with transaction id: ${transactionId} has multiple sidetree transactions. So ignoring this transaction.`;
+        console.debug(message);
+        return false;
 
       } else if (isSidetreeTx) {
         // we have found a sidetree transaction
--- a/lib/bitcoin/BitcoinProcessor.ts
+++ b/lib/bitcoin/BitcoinProcessor.ts
@@ -148,6 +148,8 @@ export default class BitcoinProcessor {
   public async initialize () {
     console.debug('Initializing ITransactionStore');
     await this.transactionStore.initialize();
+    await this.quantileCalculator.initialize();
+
     const address = this.privateKey.toAddress();
     console.debug(`Checking if bitcoin contains a wallet for ${address}`);
     if (!await this.walletExists(address.toString())) {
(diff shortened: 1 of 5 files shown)


lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -835,7 +864,7 @@ export default class BitcoinProcessor {
     const requestString = JSON.stringify(request);
     // console.debug(`Fetching ${fullPath}`);
     // console.debug(requestString);
-    console.debug(`Sending jRPC request: id: ${request.id}, method: ${request['method']}`);
+    console.debug(`Sending jRPC request: ${JSON.stringify(request)}`);
```
logangirvin: I'd prefer some smarter logic here so we don't end up logging things that would never really be useful like the signed raw transaction on broadcast, but at the very least could this be:
```suggestion
    console.debug(`Sending jRPC request: ${requestString}`);
```

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -483,10 +474,39 @@ export default class BitcoinProcessor {
...
+    const startingBlockForLastSavedGroupId = this.getStartingBlockFromGroupId(lastSavedGroupId);
+
+    // We want to delete the last group Id because we might not have saved all the transaction(s)
+    // in that group. So removing that ensures that we redo this block which will ensure that
+    // all the transactions are processed again.
+    await this.quantileCalculator.removeGroupsGreaterThanOrEqual(lastSavedGroupId);
+
+    // The starting block is included in the lastGroupId and since we are removing the lastGroupId
+    // we should remove that block from the  the transaction store as well
+    await this.transactionStore.removeTransactionsLaterThan(startingBlockForLastSavedGroupId - 1);
```
thehenrytsai: `removeTransactionsLaterThan` takes in a transaction number, not a block number, so you'll need to construct a transaction number using `TransactionNumber.construct()`

logangirvin:
```suggestion
    await this.transactionStore.removeTransactionsLaterThan(TransactionNumber.construct(startingBlockForLastSavedGroupId - 1, 0));
```

logangirvin:
```suggestion
    await this.transactionStore.removeTransactionsLaterThan(TransactionNumber.construct(startingBlockForLastSavedGroupId, 0) - 1);
```

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -483,10 +474,39 @@ export default class BitcoinProcessor {
...
 
+  private async getStartingBlockForInitialization (): Promise<IBlockInfo | undefined> {
+    // We look at the latest data that we saved in the quantile calculator. The latest
+    // Id in there tells us which group of blocks have been processed.
+    const lastSavedGroupId = await this.quantileCalculator.getLastGroupId();
+
+    if (!lastSavedGroupId) {
+      // We are returning undefined because at the time of this refactor, the calling code
+      // is expecting an undefined for this case and having this code return the genesis
+      // block is going to cause more refactor.
```
thehenrytsai: Consider adding a new issue to track this and referencing it in comment.

lib/bitcoin/BitcoinProcessor.ts
mudiali: Reverting the log message back to the previous one.

lib/bitcoin/BitcoinProcessor.ts
mudiali: Updated code to convert block to transaction number and remove it.

lib/bitcoin/BitcoinProcessor.ts
mudiali: Updated the comment with the new issue number.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -793,7 +800,10 @@ export default class BitcoinProcessor {
       const oneSidetreeTxAlreadyFound = (sidetreeTxToAdd !== undefined);
 
       if (isSidetreeTx && oneSidetreeTxAlreadyFound) {
-        throw new Error('The transaction has more then one sidetree anchor strings.');
+        // tslint:disable-next-line: max-line-length
+        const message = `The outputs in block: ${transactionBlock} with transaction id: ${transactionId} has multiple sidetree transactions. So ignoring this transaction.`;
+        console.debug(message);
+        return false;
```
thehenrytsai: This change looks good, but it raises a question regarding the try-catch block that wraps this method call: should we eat the exception or throw?

mudiali: We shouldn't catch. The catch is there right now only to log debug info.

PR #354, opened 2019-12-19
#303 Refactor lib/bitcoin to limit the bitcoin-core package references to only BitcoinClient

mudiali:


lib/bitcoin/BitcoinClient.ts
```diff
@@ -13,33 +17,91 @@ export default class BitcoinClient implements IBitcoinClient {
   /** Bitcoin peer's RPC basic authorization credentials */
   private readonly bitcoinAuthorization?: string;
 
+  /** Wallet private key */
+  private readonly privateKey: PrivateKey;
+  private readonly privateKeyAddress: Address;
```
thehenrytsai: Consider renaming to just `address` or `walletAddress`, since address is really derived from the "public" portion of the key.

lib/bitcoin/BitcoinClient.ts
```diff
@@ -148,7 +200,7 @@ export default class BitcoinClient implements IBitcoinClient {
     await this.rpcCall(request, false);
   }
 
-  public async walletExists (address: string): Promise<boolean> {
+  private async walletExists (address: string): Promise<boolean> {
```
thehenrytsai: Consider renaming method to `addressAddedToWallet()`.

lib/bitcoin/BitcoinClient.ts
```diff
@@ -108,34 +171,23 @@ export default class BitcoinClient implements IBitcoinClient {
...
+      return { satoshis: unspentOutput.satoshis };
     });
 
-    console.info(`Returning ${unspentTransactions.length} coins`);
-
     return unspentTransactions;
   }
 
-  public async importPublicKey (publicKeyAsHex: string, rescan: boolean): Promise<void> {
+  private async importPublicKey (publicKeyAsHex: string, rescan: boolean): Promise<void> {
```
thehenrytsai: Rename to `addWatchOnlyAddressToWallet()`

For reference see https://bitcoincore.org/en/doc/0.18.0/rpc/wallet/importpubkey/

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -108,48 +101,19 @@ export default class BitcoinProcessor {
...
-        break;
-    }
-    return new PrivateKey(undefined, bitcoreNetwork).toWIF();
-  }
-
   /**
    * Initializes the Bitcoin processor
    */
   public async initialize () {
     console.debug('Initializing ITransactionStore');
```
thehenrytsai: Nit: no longer just initializing transaction store, consider remove comment or update.

lib/bitcoin/BitcoinClient.ts
```diff
@@ -13,33 +17,91 @@ export default class BitcoinClient implements IBitcoinClient {
...
+    } else {
+      console.debug('Wallet found.');
+    }
+  }
+
+  /**
+   * generates a private key in WIF format
+   * @param network Which bitcoin network to generate this key for
+   */
+  public static generatePrivateKey (network: 'mainnet' | 'livenet' | 'testnet' | undefined): string {
```
thehenrytsai: Note that `bitcoin.ts` in ION repo uses this method, but I am actually fine if you simply delete that line there as opposed to expose this class or method in the export just for that purpose.

lib/bitcoin/BitcoinClient.ts
```diff
@@ -162,15 +214,81 @@ export default class BitcoinClient implements IBitcoinClient {
...
     return new Transaction(buffer);
   }
 
   // This function is specifically created to help with unit testing.
-  private static createBlockFromBuffer (buffer: Buffer): Block {
+  private static createBcoreBlockFromBuffer (buffer: Buffer): Block {
     return new Block(buffer);
   }
 
+  private async createBcoreTransaction (transactionData: string, feeInSatoshis: number): Promise<Transaction> {
```
thehenrytsai: Consider changing the method to `createRawTransaction()` and return `transaction.serialize()` instead. This way Bitcore specific logic is not exposed to caller unnecessarily.

lib/bitcoin/BitcoinClient.ts
```diff
@@ -13,33 +17,91 @@ export default class BitcoinClient implements IBitcoinClient {
...
       method: 'sendrawtransaction',
       params: [
         rawTransaction
       ]
     };
+
     const response = await this.rpcCall(request, true);
 
-    return response.length > 0;
+    if (response.length <= 0) {
```
thehenrytsai: Nit: string length won't be less than 0.

lib/bitcoin/BitcoinClient.ts
```diff
@@ -13,33 +17,91 @@ export default class BitcoinClient implements IBitcoinClient {
...
     console.info(`Broadcasting transaction ${transaction.id}`);
+
     const request = {
       method: 'sendrawtransaction',
       params: [
         rawTransaction
       ]
     };
+
     const response = await this.rpcCall(request, true);
```
thehenrytsai: Consider renaming `response` to `transactionHash`.

lib/bitcoin/BitcoinClient.ts
```diff
@@ -1,9 +1,13 @@
 import * as httpStatus from 'http-status';
-import BlockData from './models/BlockData';
+import BitcoinBlockData from './models/BitcoinBlockData';
```
thehenrytsai: Consider renaming to `BitcoinBlockModel` for overall consistency.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -253,10 +217,9 @@ export default class BitcoinProcessor {
...
     const sidetreeTransactionString = `${this.sidetreePrefix}${anchorString}`;
 
-    const address = this.privateKey.toAddress();
-    const unspentOutputs = await this.bitcoinClient.getUnspentCoins(address);
+    const unspentOutputs = await this.bitcoinClient.getUnspentCoins();
 
-    let totalSatoshis = unspentOutputs.reduce((total: number, coin: Transaction.UnspentOutput) => {
+    let totalSatoshis = unspentOutputs.reduce((total: number, coin: BitcoinUnspentCoinsModel) => {
       return total + coin.satoshis;
     }, 0);
```
thehenrytsai: Design: What we really want is an abstraction that simply returns the remaining balance left in the address. So change the interface method from `getUnspentCoins()` to something like `getBalance()`.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -689,23 +635,21 @@ export default class BitcoinProcessor {
...
     // transaction outputs in satoshis
-    const transactionOutputs: number[] = transaction.outputs.map((output) => Math.round(output.satoshis));
+    const transactionOutputs: number[] = transaction.outputs.map((output) => output.satoshis);
 
     const outputSatoshiSum = transactionOutputs.reduce((sum, value) => sum + value, 0);
 
     return (inputSatoshiSum - outputSatoshiSum);
   }
 
   private async addValidSidetreeTransactionsFromVOutsToTransactionStore (
```
thehenrytsai: Design: what if we change this method to be `getSidetreeTransactions(blockData)`, then:

1. Method becomes much more intentional.
1.  Loop logic within in `processBlock` from line 592 to 614 becomes encapsulated in this method, making `processBlock` itself much more concise thus easier to understand (Adding to MongoDB will be done in `processBlock`).

lib/bitcoin/BitcoinClient.ts
mudiali: The address is actually the private key's address so I'll keep the name.

lib/bitcoin/BitcoinClient.ts
mudiali: Hmmm .. didn't realize that. I feel that usage is helpful. Let me think about this one.

lib/bitcoin/BitcoinProcessor.ts
mudiali: Great idea .. will update.

lib/bitcoin/BitcoinProcessor.ts
mudiali: Agree with you 100%. I didn't want to do this change as I thought that this change will make this changelist bigger but looks like that you understand the context here so I can make this change now.

lib/bitcoin/BitcoinClient.ts
mudiali: I'm adding a new public method in the bitcoinprocessor which calls this one.

lib/bitcoin/BitcoinClient.ts
```diff
@@ -96,46 +150,39 @@ export default class BitcoinClient implements IBitcoinClient {
...
+      const transactionOutValue = await this.getTransactionOutValueInSatoshi(currentInput.previousTransactionId, currentInput.outputIndexInPreviousTransaction);
 
-    return unspentTransactions;
+      inputSatoshiSum += transactionOutValue;
+    }
+
+    // transaction outputs in satoshis
+    const transactionOutputs: number[] = transaction.outputs.map((output) => output.satoshis);
+
+    const outputSatoshiSum = transactionOutputs.reduce((sum, value) => sum + value, 0);
```
logangirvin: outputs.map(...).reduce(...); ?

tests/bitcoin/BitcoinClient.spec.ts
```diff
@@ -37,26 +41,62 @@ describe('BitcoinClient', async () => {
     });
   }
 
+  describe('generatePrivateKey', () => {
+    it('should construct a PrivateKey and export its WIF', () => {
+      const privateKey = BitcoinClient.generatePrivateKey('mainnet');
```
logangirvin:
```suggestion
      const privateKey = BitcoinClient.generatePrivateKey('testnet');
```
Not even in tests. Don't try it kids. Satoshi Clients will know.

mudiali: Good catch. Done.

lib/bitcoin/BitcoinClient.ts
mudiali: Feel like this is more readable in separate lines so will keep it.

Issue #293, opened 2019-09-23
Should not allow multiple keys with the same ID in operations


thehenrytsai:
Currently one can submit a create operation with multiple keys with the same ID. This should not be allowed.

isaacJChen:
I'm working on this, can't assign issues to myself RN so leaving this comment here.

PR #355, opened 2019-12-19
#293 make duplicate public key id invalid

isaacJChen:


isaacJChen:
test

tests/core/Observer.spec.ts
```diff
@@ -143,8 +143,8 @@ describe('Observer', async () => {
     // Prepare the mock response from the DownloadManager.
     const didDocumentTemplate = require('../json/didDocumentTemplate.json');
 
-    const [publicKey1, privateKey1] = await Cryptography.generateKeyPairHex('#key1', KeyUsage.recovery);
-    const [publicKey2, privateKey2] = await Cryptography.generateKeyPairHex('#key2', KeyUsage.recovery);
+    const [publicKey1, privateKey1] = await Cryptography.generateKeyPairHex('#extraRecoveryKey1', KeyUsage.recovery);
+    const [publicKey2, privateKey2] = await Cryptography.generateKeyPairHex('#extraRecoveryKey2', KeyUsage.recovery);
```
thehenrytsai: As discussed offline, lets get rid of `generateCreateOperation()` and used the declaritive approach of generating document. ie. `Document.create()`

tests/core/Operation.spec.ts
```diff
@@ -4,6 +4,7 @@ import KeyUsage from '../../lib/core/versions/latest/KeyUsage';
 import Operation from '../../lib/core/versions/latest/Operation';
 import OperationGenerator from '../generators/OperationGenerator';
 import { SidetreeError } from '../../lib/core/Error';
+import DocumentModel from '../../lib/core/versions/latest/models/DocumentModel';
```
thehenrytsai: Nit: we normally sort imports.

tests/core/MongoDbOperationStore.spec.ts
```diff
@@ -36,8 +38,23 @@ async function constructAnchoredCreateOperation (
...
+      'type': 'IdentityHub',
+      'serviceEndpoint': {
+        '@context': 'schema.identity.foundation/hub',
+        '@type': 'UserServiceEndpoint',
+        'instance': [
+          'did:sidetree:value0'
+        ]
+      }
+    }
+  ];
```
thehenrytsai: Can we create a method `OperationGenerator.createIdentityHubServiceEndpoint(instances: string[])`?

Then:

```javascript
const identityHubServiceEndpoint =  OperationGenerator.createIdentityHubServiceEndpoint(['did:sidetree:value0']);
const services = [identityServiceEndpoint];
```

tests/core/Observer.spec.ts
```diff
@@ -141,15 +143,16 @@ describe('Observer', async () => {
...
+    const signingKeys = await Cryptography.generateKeyPairHex('#key2', KeyUsage.signing);
+    const publicKeys = [
+      publicKey1,
+      signingKeys[0]
     ];
-    const operationsBuffer = operations.map((op) => { return Buffer.from(JSON.stringify(op)); });
+    const service = OperationGenerator.createIdentityHubServiceEndpoints(['did:sidetree:value0']);
+    const document = Document.create(publicKeys, service);
+    const operationsBuffer = [await OperationGenerator.createOperationBuffer(OperationType.Create, document, '#key1', privateKey1)];
+
```
mudiali: As per our offline conversation, maybe you can create the following function in OperationGenerator.ts which will be more explicit than the previous one and still be reusable. 

````
async createOperationBuffer(
  recoveryPublicKey: DidPublicKeyModel, 
  recoveryPrivateKey: string, 
  signingPublicKey: DidPublicKeyModel, 
  signingPrivateKey: string, 
  serviceEndpoints: string[]) : Promise<Buffer>{
...
}

File: tests/core/Observer.spec.ts [hunk 5]
```diff
@@ -141,15 +143,16 @@ describe('Observer', async () => {
...
-      await OperationGenerator.generateCreateOperation(didDocumentTemplate, publicKey2, privateKey2)
+    const signingKeys = await Cryptography.generateKeyPairHex('#key2', KeyUsage.signing);
+    const publicKeys = [
+      publicKey1,
+      signingKeys[0]
     ];
-    const operationsBuffer = operations.map((op) => { return Buffer.from(JSON.stringify(op)); });
+    const service = OperationGenerator.createIdentityHubServiceEndpoints(['did:sidetree:value0']);
+    const document = Document.create(publicKeys, service);
+    const operationsBuffer = [await OperationGenerator.createOperationBuffer(OperationType.Create, document, '#key1', privateKey1)];
```
mudiali: Previously, the test was creating 2 operation buffers. Let's keep that logic and create 2 buffers.

tests/generators/OperationGenerator.ts
```diff
@@ -242,4 +216,22 @@ export default class OperationGenerator {
...
+   */
+  public static createIdentityHubServiceEndpoints (instances: string[]): any[] {
+    return [
+      {
+        'id': 'IdentityHub',
+        'type': 'IdentityHub',
+        'serviceEndpoint': {
+          '@context': 'schema.identity.foundation/hub',
+          '@type': 'UserServiceEndpoint',
+          'instance': instances
```
mudiali: Looks like we are only creating the 'UserServiceEndpoint' (see ````'@type': 'UserServiceEndpoint'````), can we change the function name to reflect that please? 

Also, seems like that the 'instance' property is supposed to be single-valued right? If single valued, then let's change the input to ````instance: string````

isaacJChen: I will double check with the rest of the team to make sure if it's supposed to be singular or plural.

isaacJChen: I changed it to instances for now and everything seems to be working fine.

isaacJChen: Changed instances and related validations.

Issue #349, opened 2019-12-14
BitcoinProcessor::processTransactions does not check for any potential forks during block processing


mudiali:
During initialization processTransaction routine processes blocks from the lastProcessedBlock.height to the current block height. There is a bug here that if the underlying blockchain forks in the middle of processing, the processTransaction would not know and can be in an consistent state. 

One potential fix is to keep include the 'just-processed-block'.hash when processing the next block; and check that the nextblock.previousHash === 'just-processed-block'.hash. And if that check fails then throw.

PR #357, opened 2019-12-19
#349 Fixing BitcoinProcessor::ProcessTransactions to detect forks mid-block processing.

logangirvin:
Fixing #349 by introducing "PreviousHash" as a BlockInfo primitive and all the propagation that ensues. The actual check is done by processBlock, and the logic for passing the correct values is done by processTransactions.


lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -355,31 +362,33 @@ export default class BitcoinProcessor {
...
+      lastValidBlock = await this.revertBlockchainCache();
+    }
 
-    return {
-      height: startingBlock,
-      hash: await this.bitcoinClient.getBlockHash(startingBlock)
-    };
+    // Our starting block is the one after the last-valid-block
+    return this.bitcoinClient.getBlockInfoFromHeight(lastValidBlock.height + 1);
   }
```
thehenrytsai: Sanity: this section looks very similar to `getStartingBlockForPeriodicPoll`, no point in consolidating?

mudiali: The two functions are different enough that we can't really merge them. I'll keep them as they are.

PR #366, opened 2020-01-02
#303 Add support to let the admin define a bitcoin fee spending cap.

mudiali:


lib/bitcoin/SpendingMonitor.ts
```diff
@@ -0,0 +1,113 @@
...
+    // the these 2 blocks from the cutoff period. For example:
+    //  - if the cutoff period is 2, then we want transactions from the last-processed-block and the next one.
+    //  - if the cutoff period is 3, then we want transactions from the last-process-block - 1, last-processed-block, and the next one
+    const startingBlockHeight = lastProcessedBlockHeight - this.bitcoinFeeSpendingCutoffPeriodInBlocks - 2;
+
+    // Now get the transactions from the store which are included in the above starting block and higher.
+    const startingBlockFirstTxnNumber = TransactionNumber.construct(startingBlockHeight, 0);
+
+    const allTxnsSinceStartingBlock =
+      await this.transactionStore.getTransactionsLaterThan(startingBlockFirstTxnNumber - 1, SpendingMonitor.maxResultsFromTransactionStore);
```
thehenrytsai: Consider changing the `getTransactionsLaterThan()` interface method so that it can take only the first parameter. This way `maxResultsFromTransactionStore` won't be needed and the spending calculation will not yield false positive allow-writes.

lib/bitcoin/SpendingMonitor.ts
```diff
@@ -0,0 +1,113 @@
...
+  public addTransactionDataBeingWritten (anchorString: string): void {
+    this.anchorStringsWritten.add(anchorString);
+  }
+
+  /**
+   * Calculates whether the specified fee will result in this node going over the spending limit.
+   * @param currentFeeInSatoshis The fee to be added for the next transaction.
+   * @param startingBlockHeight The block height to start the check for the cutoff period.
+   */
+  public async isCurrentFeeOverSpendingLimit (currentFeeInSatoshis: number, lastProcessedBlockHeight: number): Promise<boolean> {
```
thehenrytsai: Style (minor thus optional), consider having returned `true` as the "happy" state as opposed to the "failure" state, so consider `isCurrentFeeWithinSpendingLimit`.

Issue #252, opened 2019-06-29
Add commit/reveal scheme to enable fast next-op proof of legitimacy and prevent DDoR (DID Denial of Resolution) spam attacks


csuwildcat:
Presently, anyone can claim a batch might contain an operation for a given DID, thus a malicious actor could spam anchor file DID Suffix assertions against a target DID, which would amass an unusually large number of batch files to fetch and check at resolution time. Presently, we are mitigating this by only allowing one DID op per DID per batch file, to at least force the attacker to pay for multiple transactions. However, there may be a far better solution - here's how it would work:

1. Alice creates her first operation, and part of that op is a next-op commitment to the hash of a random value that is included in the signed portion of the op payload.
2. Alice is the only one who knows the commitment value, thus when she needs to do her next update op, she creates the op and includes a reveal of the hashed commitment value, which is paired with her DID Suffix claim in the Anchor file. (she also adds a new next-op commitment value to her update op, so she can repeat the process for the next op)
3. Anyone that has processed Alice's first op would know that the next value they are looking for is an Anchor file DID Suffix value that matches the commitment hash she included with her previous op, which only Alice knows and can reveal.

This basically halts the entire class of attack by providing a fast hash/value check nodes can perform against the accrued anchor file DID Suffix claims to chew through any spurious DID op inclusion claims very quickly, before ever incurring a batch download.

csuwildcat:
Here's a visual walk-through of how this would work: 

![next-op-commitments](https://user-images.githubusercontent.com/131786/60448950-7be10b80-9bdb-11e9-827e-6c4af59cae50.png)

OR13:
IMO, this is a critical protocol fix.

This would be an excellent candidate for protocol upgrade integration tests, where we change both the commit algorithm and the operation hash algorithm, and show stability.

csuwildcat:
Additional note about regular next-op commitments vs recovery ops:

There should be a next-op commitment value created for each op at the time of update, and we can debate whether to encrypt that value in payload or not (I would prefer not, as it will eat DID Doc space), but there is one thing that we must enforce: recovery ops should have a different per-recovery next-op commitment value that is encrypted in the DID Doc, to ensure that it is maximally available and persisted.

OR13:
The word encrypted is confusing me here... Isn't it the case that all operation data is plaintext, and a chain of scheme(secretN, nextOpHash) => nextOpCommit, is used to protect nodes from downloading bad data?

[...]

PR #369, opened 2020-01-09
#252 - Added one-time password (commit-reveal) scheme to enable fast processing of invalid operations

thehenrytsai:


lib/core/Resolver.ts
```diff
@@ -38,57 +39,37 @@ export default class Resolver {
       op.type === OperationType.Delete);
 
     // Apply "full" operations first.
-    const [lastFullOperation, lastFullOperationHash] =
-      await this.applyOperations(didDocumentReference, createAndRecoverAndRevokeOperations, undefined, undefined);
+    await this.applyOperations(createAndRecoverAndRevokeOperations, didResolutionModel);
 
-    // If no full operation found at all, the DID is not anchored.
-    if (lastFullOperation === undefined) {
+    // If no valid operation full found at all, the DID is not anchored.
```
OR13: nit: wording a little unclear

lib/core/versions/0.4.0/OperationProcessor.ts
```diff
@@ -20,81 +20,78 @@ export default class OperationProcessor implements IOperationProcessor {
...
+          return { validOperation: false };
         }
 
         const originalDidDocument = Document.from(operation.encodedPayload, this.didMethodName, ProtocolParameters.hashAlgorithmInMultihashCode)!;
 
         const signingKey = Document.getPublicKey(originalDidDocument, operation.signingKeyId);
 
         if (!signingKey) {
-          return { validOperation: false, operationHash };
+          return { validOperation: false };
```
OR13: nit: some amount of code repetition here, may more readable with a const ApplyResult for valid and invalue, as opposed to looking at the boolean value.

tests/core/Operation.spec.ts
```diff
@@ -187,35 +214,120 @@ describe('Operation', async () => {
...
-            publicKeys: [
-              { id: 'aRepeatingId', type: 'thisShouldNotShowUp', usage: 'thisShouldNotShowUp' },
-              { id: 'aNonRepeatingId', type: 'someType', usage: 'some usage' }
-            ]
-          }
-        ];
+  describe('applyPatchesToDidDocument()', async () => {
+    it('should prevent the same id for multiple keys', async () => {
+      const didDocument: DocumentModel = {
+        '@context': 'someContext',
```
OR13: context is not valid, should be a uri or an object.

Latest W3C Context: "https://www.w3.org/ns/did/v1" will work.

OR13: Note any fields defined but not present in the context above will cause errors when processing the DID Document as JSON-LD...

We have defined: https://docs.element-did.com/contexts/sidetree/sidetree-v0.0.jsonld

for "usage", but not updated our context to reflect this yet.

The most correct thing for us to do would be:

"@context": ["https://www.w3.org/ns/did/v1", "https://identity.foundation/contexts/sidetree/sidetree-v0.0.jsonld"]

This is not currently possible, but I want to put it on the radar.

lib/core/versions/0.4.0/OperationProcessor.ts
```diff
@@ -20,81 +20,78 @@ export default class OperationProcessor implements IOperationProcessor {
...
-
       const operation = AnchoredOperation.createAnchoredOperation(anchoredOperationModel);
-      operationHash = operation.operationHash;
 
       if (operation.type === OperationType.Create) {
 
         // If we have seen a previous create operation.
         if (didDocumentReference.didDocument) {
-          return { validOperation: false, operationHash };
+          return { validOperation: false };
```
gjgd: Why are we modifying files from previous versions?

isaacJChen: I have the same question.

lib/core/Resolver.ts
thehenrytsai: Good pick up, fixed.

lib/core/versions/0.5.0/OperationProcessor.ts
```diff
@@ -20,81 +20,77 @@ export default class OperationProcessor implements IOperationProcessor {
...
 
         // If we have not seen a valid create operation yet.
         if (didDocument === undefined) {
-          return { validOperation: false, operationHash };
+          return { validOperation: false };
         }
 
         // The current did document should contain the public key mentioned in the operation ...
         const publicKey = Document.getPublicKey(didDocument, operation.signingKeyId);
         if (!publicKey) {
```
isaacJChen: nit: `if (!publickey || !(await operation.verifySignature(publicKey)))` ? Same with other places that used the same structure.

lib/core/versions/0.5.0/OperationProcessor.ts
```diff
@@ -20,81 +20,77 @@ export default class OperationProcessor implements IOperationProcessor {
...
         didDocumentReference.didDocument = undefined;
-        return { validOperation: true, operationHash };
+        return { validOperation: true };
       } else {
         // Update operation
 
         // If we have not seen a valid create operation yet.
         if (didDocument === undefined) {
-          return { validOperation: false, operationHash };
+          return { validOperation: false };
```
isaacJChen: These are repeated a lot. Would it make sense to have a default return object and just return it on these conditions? Then in the validOperation = true case, manipulate the default object so we can change all of them by changing just one place.

lib/core/versions/0.4.0/OperationProcessor.ts
thehenrytsai: Fair point, but will won't fix this because:

1. This is old versioned code, so don't intend to do make any changes unless really needed.
1. We are planning to reset the genesis block in the testnet in the next ION release, at which point this code will be removed forever.

tests/core/Operation.spec.ts
thehenrytsai: Changed to `https://www.w3.org/ns/did/v1`

lib/core/versions/0.4.0/OperationProcessor.ts
thehenrytsai: We normally should avoid modifying code in previous versions as much as possible. But since this change involves an interface change (which in theory should really be minimized, but here we are), so changing old code to match the interface.

lib/core/versions/0.5.0/OperationProcessor.ts
OR13: I thought the same, a constant for each case would be more readable: 

```
const InvalidOperationResponse = { validOperation: false };
...
return InvalidOperationResponse;
```

readability isn't that bad now, but if the response type starts to get more complicated I could see this helping...

thehenrytsai: I think this is the same point that Orie pointed out earlier, which is totally fair. But as explained earlier I intend to won't fix this because:

1. This is old versioned code, so don't intend to do make any changes unless really needed. Latest code (under `latest` already made improvements on this old code, feel free to check it out).
1. We are planning to reset the genesis block in the testnet in the next ION release, at which point this code will be removed forever.

lib/core/versions/0.5.0/OperationProcessor.ts
thehenrytsai: Same comment as other versioned code.

csuwildcat:
Just want to thank everyone who participated in helping to craft this scheme, and obviously most of all to the devs who actually implemented it. This was one of the more significant outstanding items to add before mainnet, and I am really excited that we're one step closer!

PR #371, opened 2020-01-13
Ische/implement operation rate limiting

isaacJChen:


lib/core/Observer.ts
```diff
@@ -1,12 +1,15 @@
 import IBlockchain from './interfaces/IBlockchain';
+import IOperationRateLimiter from './interfaces/IOperationRateLimiter';
 import IOperationStore from './interfaces/IOperationStore';
 import ITransactionProcessor from './interfaces/ITransactionProcessor';
 import ITransactionStore from './interfaces/ITransactionStore';
 import IUnresolvableTransactionStore from './interfaces/IUnresolvableTransactionStore';
 import IVersionManager from './interfaces/IVersionManager';
+import OperationRateLimiter from './versions/latest/OperationRateLimiter';
```
thehenrytsai: Design: I have yet to read through the code. But it seems undesirable that this layer references any versioned code.

lib/core/Observer.ts
```diff
@@ -1,12 +1,15 @@
...
 import SharedErrorCode from '../common/SharedErrorCode';
 import timeSpan = require('time-span');
 import TransactionModel from '../common/models/TransactionModel';
+import TransactionNumber from '../bitcoin/TransactionNumber';
```
thehenrytsai: Design: Similar to the previous comment, I have yet to see why this is needed, but block height is intentionally hidden from the core up until this point. If this class must be used, at very lease it needs to be moved to `common` folder.

lib/core/Observer.ts
isaacJChen: O operation rate limiter shouldn't be versioned anymore. I will move it out.

thehenrytsai: As discussed offline, we'll need to version rate limiting algorithm.

lib/core/Observer.ts
```diff
@@ -102,8 +112,9 @@ export default class Observer {
         let transactions = readResult ? readResult.transactions : [];
         moreTransactions = readResult ? readResult.moreTransactions : false;
 
+        const transactionsToProcess = this.operationRateLimiter.getHighestFeeTransactionsPerBlock(transactions);
```
thehenrytsai: Design: an alternative approach would be to make Observer aware of of different versions of rate limiting by:

1. (Optional first step) Ensuring the contract for fetching transactions from bitcoin service always returns complete set in same blockchain time/height.
2. Group the given transactions by their blockchain time/height.
3. Feed each transaction group to the correct version of rate limiter to select the qualified transactions.
4. Finally process the select transactions.

This logic probably involves 1 loop nested in another.

package.json
```diff
@@ -46,7 +47,7 @@
   "scripts": {
     "precommit": "npm run lint",
     "build": "tsc && copyfiles \"lib/**/*.json\" dist && copyfiles \"tests/**/*.js*\" dist && copyfiles \"package.json\" dist",
-    "test": "npm run build && nyc jasmine --config=./tests/jasmine.json",
+    "test": "npm run build && node node_modules/nyc/bin/nyc.js node_modules/jasmine/bin/jasmine.js --config=./tests/jasmine.json",
```
thehenrytsai: Undo this.

package.json
```diff
@@ -16,6 +16,7 @@
     "mongodb": "3.2.3",
     "multihashes": "0.4.14",
     "node-fetch": "2.6.0",
+    "priorityqueue": "^1.0.0",
```
thehenrytsai: Remove `^`. We force version selection.

lib/core/Observer.ts
```diff
@@ -39,6 +43,7 @@ export default class Observer {
     private transactionStore: ITransactionStore,
     private unresolvableTransactionStore: IUnresolvableTransactionStore,
     private observingIntervalInSeconds: number) {
+    this.operationRateLimiter = new OperationRateLimiter(10000); // temp number TBD
```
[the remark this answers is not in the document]
thehenrytsai: Agree, more importantly, input of this number needs to come from config file.

lib/core/interfaces/IOperationRateLimiter.ts
```diff
@@ -0,0 +1,19 @@
...
+ * Rate limits operations. only allow n number of highest fee transactions to pass.
+ */
+export default interface IOperationRateLimiter {
+
+  /**
+   * Given an array of transactions, return a subset of them which have high enough fee within their block to be processed. In the returned object,
+   * also contains transactions that belong to an incomplete block at the end of the array
+   * @param orderedTransactions An array transactions which are ordered by transactionTime in ascending order
+   */
+  getHighestFeeTransactionsPerBlock (orderedTransactions: TransactionModel[]): TransactionModel[];
```
thehenrytsai: Make this method name generic, since selection criteria can vary completely between versions. Perhaps `selectQualifiedTransactions()`.

lib/core/interfaces/IOperationRateLimiter.ts
```diff
@@ -0,0 +1,19 @@
+import TransactionModel from '../../common/models/TransactionModel';
+
+/**
+ * Rate limits operations. only allow n number of highest fee transactions to pass.
+ */
+export default interface IOperationRateLimiter {
```
thehenrytsai: Discussion: I wonder if we should generalize this name to something like `IThroughputLimiter` or we only foresee rate limiting by operation forever.

lib/core/interfaces/IOperationRateLimiter.ts
```diff
@@ -0,0 +1,19 @@
...
+
+  /**
+   * Given an array of transactions, return a subset of them which have high enough fee within their block to be processed. In the returned object,
+   * also contains transactions that belong to an incomplete block at the end of the array
```
thehenrytsai: Sanity: seems that this part of the comment is not true? Doesn't look like that's what's been implemented.

lib/core/interfaces/IOperationRateLimiter.ts
isaacJChen: Good idea. It doesn't have to be operation.

lib/core/Observer.ts
isaacJChen: Actually now that you drew my attention to this, is this calculation specific to bitcoin? If so, we shouldn't use it.

PR #372, opened 2020-01-14
#252 - Added documentation on operation OTP

thehenrytsai:


docs/protocol.md
```diff
@@ -180,6 +179,18 @@ Sidetree protocol defines the following two mechanisms to enable scaling, while
...
+### One Operation per DID per Batch
+  Only one operation per DID per batch is allowed, this prevents the operation chain of any DID from growing at an intractable rate.
+
+#### One-Time Password (OTP) for Operations
+  Upon DID creation, the create operation payload must include:
+  1. The hash of a _one-time password_ (OTP) for the next recovery operation.
+  1. The hash of a _one-time password_ (OTP) for the next update operation.
+
+  The DID owner must reproduce and present the correct OTP in the subsequent operation for the operation to be considered valid. In addition, each subsequent operation must also include the hash of the new OTP(s) for the next operation. This scheme enables efficient dismissal of counterfeit operations without needing to evaluate signatures.
+
```
mudiali: Do you think that it is worth explaining that the for the update operation, only the updateOTP is required and the recoveryOTP-hash doesn't need to be updated/included and vice versa for the recovery op?

Issue #324, opened 2019-11-05
Consider moving/copying `operation` property in operation header into the payload.


thehenrytsai:

OR13:
@thehenrytsai can this be closed now?

OR13:
@thehenrytsai I'm marking this as pending close. there is not enough information on the issue, and it appears to be addressed.

PR #374, opened 2020-01-16
#324 - Moved operation/type property in JWS header into payload.

thehenrytsai:


OR13:
I'd be interested to know the context for this change.

I think it makes sense though, we should avoid polluting the JWS header with anything that does not help with JOSE verification.

thehenrytsai:
> I'd be interested to know the context for this change.
> 
> I think it makes sense though, we should avoid polluting the JWS header with anything that does not help with JOSE verification.

Hey @OR13, you are right, this is an attempt to remove any polluting to the JWS header. Since we have started adding a bunch of properties into the payload itself, the operation type should really be in payload also for consistency.

lib/core/versions/latest/Operation.ts
```diff
@@ -282,6 +272,16 @@ export default class Operation {
     const decodedPayloadJson = Encoder.decodeAsString(operation.payload);
     const decodedPayload = JSON.parse(decodedPayloadJson);
 
+    // Get the operation type.
+    const operationType = decodedPayload.type;
+
+    // 'type' property must be one of the allowed strings.
+    const allowedOperations = new Set([OperationType.Create, OperationType.Update, OperationType.Delete, OperationType.Recover]);
```
isaacJChen: nit: I'm pretty sure this can just be `new Set(Object.values(OperationType))` this way we don't need to hard code the values and add them in the future.

lib/core/versions/latest/ErrorCode.ts
```diff
@@ -32,11 +32,14 @@ export default {
   OperationCreateInvalidDidDocument: 'operation_create_invalid_did_document',
   OperationCountLessThanZero: 'operation_count_less_than_zero',
   OperationExceedsMaximumSize: 'operation_exceeds_maximum_size',
-  OperationHeaderMissingKid: 'operation_header_missing_kid',
   OperationHeaderMissingOrIncorrectAlg: 'operation_header_missing_or_incorrect_alg',
-  OperationHeaderMissingOrIncorrectOperation: 'operation_header_missing_or_incorrect_operation',
-  OperationMissingOrIncorrectPayload: 'operation_missing_or_incorrect_payload',
+  OperationHeaderMissingOrIncorrectKid: 'operation_header_missing_kid',
```
gjgd: nit inconsistent name with OperationHeaderMissingOrIncorrectAlg above
```suggestion
  OperationHeaderMissingOrIncorrectKid: 'operation_header_missing_or_incorrect_kid',
```

thehenrytsai: Thanks for observing it.

lib/core/versions/latest/Operation.ts
thehenrytsai: This is great suggestion. Took it already!

PR #375, opened 2020-01-17
Ische/throughputLimiter

isaacJChen:


lib/core/interfaces/IThroughputLimiter.ts
```diff
@@ -0,0 +1,13 @@
+import TransactionModel from '../../common/models/TransactionModel';
+
+/**
+ * Rate limits operations. only allow n number of highest fee transactions to pass.
```
thehenrytsai: Remove the part regarding highest fee, that's implementation specific.

lib/core/models/Config.ts
```diff
@@ -10,4 +10,6 @@ export default interface Config {
   maxConcurrentDownloads: number;
   observingIntervalInSeconds: number;
   mongoDbConnectionString: string;
+  maxNumberOfOperationsPerBlock: number;
+  maxNumberOfTransactionsPerBlock: number;
```
thehenrytsai: Design: these two protocol config parameters definitely should not come from server config file. Move them to ProtocolParameters.ts

tests/core/Observer.spec.ts
```diff
@@ -50,9 +51,11 @@ describe('Observer', async () => {
     downloadManager.start();
 
     const transactionProcessor = new TransactionProcessor(downloadManager, operationStore);
+    const throughputLimiter = new ThroughputLimiter(10, 25, transactionStore);
```
thehenrytsai: For my understanding, any particular reason for choosing 10,  and 25? Doesn't look like it is required for the test to pass.

tests/core/OperationThroughputLimiter.spec.ts
```diff
@@ -0,0 +1,199 @@
...
+        }),
+        transactionFeePaid: 14,
+        normalizedTransactionFee: 1
+      }
+    ];
+  }
+
+  beforeEach(() => {
+    transactionStore = new MockTransactionStore();
+    throughputLimiter = new ThroughputLimiter(10, 25, transactionStore);
```
thehenrytsai: Nit: For readability, can we assign const variables to 10 and 25 before passing so reader knows which is which.

tests/core/OperationThroughputLimiter.spec.ts
```diff
@@ -0,0 +1,199 @@
+import AnchoredDataSerializer from '../../lib/core/versions/latest/AnchoredDataSerializer';
```
thehenrytsai: Let's name the file `ThroughputLimiter.spec.ts` for consistency.

lib/core/versions/latest/ThroughputLimiter.ts
```diff
@@ -0,0 +1,90 @@
...
+        throw new SidetreeError(ErrorCode.TransactionsNotInSameBlock, 'transaction must be in the same block to perform rate limiting');
+      }
+      this.transactionsPriorityQueue.push(transaction);
+    }
+
+    const [numberOfOperations, numberOfTransactions] = await this.getNumberOfOperationsAndTransactionsAlreadyInBlock(currentBlockHeight);
+    let numberOfOperationsToQualify = this.maxNumberOfOperationsPerBlock - numberOfOperations;
+    let numberOfTransactionsToQualify = this.maxNumberOfTransactionsPerBlock - numberOfTransactions;
+    const transactionsToReturn = this.getHighestFeeTransactionsFromCurrentTransactionTime(numberOfOperationsToQualify, numberOfTransactionsToQualify);
+    this.transactionsPriorityQueue.clear();
```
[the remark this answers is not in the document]
mudiali: My vote is for #2 as well. It's cleaner.

lib/core/Observer.ts
```diff
@@ -104,13 +111,26 @@ export default class Observer {
...
         // Queue parallel downloading and processing of batch files.
         for (const transaction of transactions) {
-          const awaitingTransaction = {
-            transaction: transaction,
-            processingStatus: TransactionProcessingStatus.Pending
-          };
-          this.transactionsUnderProcessing.push(awaitingTransaction);
-          // Intentionally not awaiting on downloading and processing each operation batch.
-          void this.processTransaction(transaction, awaitingTransaction);
+          if (transaction.transactionTime === this.currentBlockHeight) {
```
mudiali: Another design thought: How about we make another class which does the 'filter-by-rate-limiting' outside of the observer. Then the code in this class changes to (I didn't put too much thought on the naming):
````
constructor (...) {
  ...
  this.operationRateLimitManager = new OperationRateLimitManager(this.versionManager);
  ...
}

private async processTransactions(...) {
  ...
  // This is the new line that rate-limit the existing transactions and the rest of the code remains the same.
  qualifiedTransactions = this.operationRateLimitManager.selectQualifiedTransactions(transactions);
  
  for (const transaction of qualifiedTransactions) {
  ...
}
````

Pros:
1. The new variables + functionality introduced here moves to that new class. 
1. Existing tests of this class don't need to change too much. They just need to spy on this new member ````this.operationLimitManager````.
1. Testing the new class becomes easier as creating test data will be easier.

Comment (isaacJChen) [reply lib/core/Observer.ts]: Good idea.

File: tests/core/Observer.spec.ts (hunk 3 above)
Comment (isaacJChen) [reply tests/core/Observer.spec.ts]: It's just some value that also allows testing for transactions rate limiting and operation rate limiting.

File: lib/core/versions/0.4.0/ThroughputLimiter.ts [hunk 8]
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * dummy throughput limiter
+ */
+export default class ThroughputLimiter implements IThroughputLimiter {
+
+  /**
+   * Return what is passed in
+   */
+  selectQualifiedTransactions (orderedTransactions: TransactionModel[]): Promise<TransactionModel[]> {
```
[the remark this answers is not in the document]
isaacJChen: I don't think so? We didn't have throughput limiter in the old blocks so this should just accept whatever is given and spit it back out.

lib/core/versions/latest/ThroughputLimiter.ts
```diff
@@ -0,0 +1,90 @@
...
+    if (!orderedTransactions.length) {
+      return [];
+    }
+
+    const currentBlockHeight = orderedTransactions[0].transactionTime;
+    for (const transaction of orderedTransactions) {
+      if (transaction.transactionTime !== currentBlockHeight) {
+        this.transactionsPriorityQueue.clear();
+        throw new SidetreeError(ErrorCode.TransactionsNotInSameBlock, 'transaction must be in the same block to perform rate limiting');
+      }
```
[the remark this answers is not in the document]
isaacJChen: You're right. I added it just in case we make a mistake in observer, we will know from this error. Also I think it doesn't hurt to be safe. What do you think?

lib/core/versions/0.4.0/ThroughputLimiter.ts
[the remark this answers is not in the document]
thehenrytsai: Sorry I wasn't clear, I was more referring to the interface as opposed to the implementation. I think if the input transactions does not need to be "ordered", then just call it `transactions` in the interface.

lib/core/versions/latest/ThroughputLimiter.ts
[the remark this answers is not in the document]
thehenrytsai: I don't feel too strongly about it, so we can keep if you'd like. I agree with defensive programming in principle, but in this context, there are many "checks" that we could have done in the current codebase but have not, so I ask myself what makes this check special and others not?

lib/core/Observer.ts
```diff
@@ -104,13 +111,26 @@ export default class Observer {
...
+            this.transactionsInCurrentBlock.push(transaction);
+          } else {
+            if (this.currentBlockHeight !== undefined) {
+              const throughputLimiter = this.versionManager.getThroughputLimiter(this.currentBlockHeight);
+              const qualifiedTransactionsInCurrentBlock = await throughputLimiter.selectQualifiedTransactions(this.transactionsInCurrentBlock);
```
thehenrytsai: We also need to update `protocol.md` describing rate/throughput limiting as one of the  anti-spam mechanism, under DDoS Attack & Mitigation section, and I am fine if it comes in as a separate PR.

lib/core/versions/latest/ThroughputLimiter.ts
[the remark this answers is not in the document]
isaacJChen: I feel like the check should be there because the class and function shouldn't know how it's being used, so it shouldn't assume that the array is ordered. I will leave it for now. It's a quick change to remove it if you think otherwise.

docs/protocol.md
```diff
@@ -163,6 +163,16 @@ Without any mitigation strategy, malicious but protocol adherent nodes can creat
...
 
+#### Rate limiting
+   
+   To prevent spam attack causing transaction and operation stores to grow at an unhealthy rate, rate limiting is put in place. 
+
+   The current implementation caps the number of operations allowed to be observed and processed to 1000 a second, and the cap for transactions is set to 300 per transaction time. The selection logic when the caps are exceeded is the following:
+   
+   higher fee per transaction comes first, if transaction fee is the same, lower transaction number comes first.
+   
+   By picking the transactions with higher transaction fee, it encourages batching, while allowing small transactions to have the opportunity to also be included if they are willing to pay a premium of a couple USD. Alternatives to this approach are highest fee per operation and first comes first serve, but they don't encourage batching and discourage writing near the end of a transaction time.
```
mudiali: Suggestion to change the " ... pay a premium of a couple of USD." to something like "... pay a higher transaction fee."

tests/core/Observer.spec.ts
```diff
@@ -329,23 +408,43 @@ describe('Observer', async () => {
...
+          }),
           'transactionFeePaid': 1,
           'normalizedTransactionFee': 1
         },
         {
           'transactionNumber': 4,
           'transactionTime': 4000,
           'transactionTimeHash': '4000',
-          'anchorString': '4thTransaction',
+          'anchorString': AnchoredDataSerializer.serialize({
```
mudiali: These changes are not really needed anymore right? I mean the observer tests now only need to spy on the througputLimiter call it seems.

isaacJChen: Do you not think it's useful to test end to end?

lib/core/versions/latest/TransactionSelector.ts
```diff
@@ -0,0 +1,103 @@
...
+      }
+    }
+    const numberOfTransactions = transactions ? transactions.length : 0;
+    return [numberOfOperations, numberOfTransactions];
+  }
+
+  /**
+   * Given transactions within a block, return the ones that should be processed.
+   */
+  private getHighestFeeTransactionsFromCurrentTransactionTime (
```
thehenrytsai: Nit: static.

lib/core/versions/latest/TransactionSelector.ts
```diff
@@ -0,0 +1,103 @@
...
+      const numOfOperationsInCurrentTransaction = AnchoredDataSerializer.deserialize(currentTransaction.anchorString).numberOfOperations;
+      numberOfOperationsSeen += numOfOperationsInCurrentTransaction;
+
+      if (numberOfOperationsSeen <= numberOfOperationsToQualify) {
+        transactionsToReturn.push(currentTransaction);
+      }
+    }
+
+    // sort based on transaction number ascending
+    return transactionsToReturn.sort((a, b) => { return a.transactionNumber - b.transactionNumber; });
```
thehenrytsai: Should we move this line back to the observer or throughput limiter, since this requirement is none of the selector's concern.

lib/core/ThroughputLimiter.ts
```diff
@@ -0,0 +1,48 @@
...
+   * given a an array of transactions, return an array of qualified transactions based on current state
+   * @param transactions array of transactions to filter for
+   */
+  public async getQualifiedTransactions (transactions: TransactionModel[]) {
+    let qualifiedTransactions: TransactionModel[] = [];
+    for (const transaction of transactions) {
+      if (transaction.transactionTime === this.currentBlockHeight) {
+        this.transactionsInCurrentBlock.push(transaction);
+      } else {
+        if (this.currentBlockHeight !== undefined) {
```
thehenrytsai: Design: it just occurred to me that it is quite undesirable for transactions to be held indefinitely in the priority queue, until if/when another transaction is written in a later block... this seems to be a real issue at least during the foreseeable future. The only way I can think of to remedy this is to require a logical interface change to blockchain `read()` to always return complete block data. Unfortunately I think it is the right thing to do, what do you guys think?

lib/core/versions/latest/TransactionSelector.ts
isaacJChen: I agree.

lib/core/ThroughputLimiter.ts
isaacJChen: Or maybe we can put them in something like MongoDB until they are needed? I assume memory is your concern right?

tests/core/Observer.spec.ts
mudiali: E2E tests are definitely useful but we are not doing end 2 end in these tests are we?

My concern is that if we change some logic in the throughputLimiter, we would have to come back and change these tests. IMO, moving that logic out of that class (or even into a new function) has the additional advantage of making our testing easier. A new test can be added which tests whether the througputLimiter is being called or not and only the txns returned by that function are being processed.

Given above, if you still want to keep these here then I'm fine as you have already done the hard work and written the tests.

lib/core/ThroughputLimiter.ts
isaacJChen: Will address in another issue and pr

lib/core/versions/0.4.0/TransactionSelector.ts
```diff
@@ -0,0 +1,15 @@
+import ITransactionSelector from '../../interfaces/ITransactionSelector';
+import TransactionModel from '../../../common/models/TransactionModel';
+
+/**
+ * dummy transaction selector
```
OR13: is this accurate documentation?

isaacJChen: Yes, they are not supposed to do anything but return what you pass in.

PR #380, opened 2020-01-27
Add Prettier

OR13:
When editing files, VS Code is thrashing whitespace.

This will cause any small change to also include lots of formatting issue.

Its possible there is a better way to configure VS Code not to do this, but my desired behavior is as follows:

```
git clone git@github.com:decentralized-identity/sidetree.git
cd sidetree
npm i
code .
# edit file and save
npm run lint
# no lint changes.
```


OR13:
I like that we are using githooks and `tslint --fix --project .`... there is never a point in running a linter without the fix option IMO.

thehenrytsai:
@OR13, thanks for the PR. In principle I am all for automated standardization on formatting style of code. I personally felt TSLint does a sufficient job at providing rules and we also have a `precommit` script that auto fixes the violations upon commit.

Is there's a reason why TSLint and/or this process isn't sufficient? If you feel strongly about `prettier`, pending inputs from others, I would still at very least push back on some of the rules currently used at a glance:

1. We follow JavaScript Standard Style, which adds a space before a function declaration's parentheses, the config in prettier currently violates that rule.
2. We feel enforcing each function parameter to be in their individual line is really an overkill, and make code "uglier" in most cases in our opinion.
3. Similar feeling on forcing new lines on chained functions.

OR13:
@thehenrytsai When you edit files in VSCode, does the edited version (after save) match what the linter will format it to with lint fix?

Thats what I am trying to figure out how to do.

I suppose I can live with seeing the file change a bunch, and then waiting for the linter to fix it.

I'm just trying to figure out how to make VSCode auto format to the same result as lint fix.

OR13:
Right now when I edit a file, it changes quote marks when I save...

then after one line change the linter will correct all those issues before the commit happens:

```
› git commit -m "test"

Warning: Setting pre-commit script in package.json > scripts will be deprecated
Please move it to husky.hooks in package.json, a .huskyrc file, or a husky.config.js file
Or run ./node_modules/.bin/husky-upgrade for automatic update

See https://github.com/typicode/husky for usage

husky > pre-commit (node v10.15.3)

> @decentralized-identity/sidetree@0.5.0 lint /Users/orie/Desktop/DIF/sidetree
> tslint --fix --project .

Fixed 7 error(s) in /Users/orie/Desktop/DIF/sidetree/lib/index.ts


[feat/test-vscode f70a62e] test
 1 file changed, 10 insertions(+), 13 deletions(-)
```

OR13:
I'm gonna close this, I will reopen if I can't figure out how to get VSCode to be stable, I think the issue is prettier needs to be disabled for this project. its probably in my user settings.

PR #382, opened 2020-01-28
Prevent Prettier from fighting TSLint

OR13:
disables prettier in vscode settings.


OR13:
This seems to have done the trick... sorry for wasting your time with that trash PR :)

Issue #271, opened 2019-07-25
Implement economic locking as a prerequisite for writing larger batches


csuwildcat:
Once the Proof of Fee algos/code are in place (which shouldn't be too long now), there will be a network-determined cost for individual operations set across block-windows (based on a deterministic calculation nodes can all use to get the right answer) which will force writers to pay a txn fee in accordance with the size of the op batch they seek to write.

When a fee variable exists in the system, we may have another interesting tool at our disposal: a form of staking - but to be clear, not Proof of Stake or some consensus scheme of that name. Imagine that the ability to write very large batches and have them be seen as valid required you to 'stake' a much larger sum locked in an on-chain txn and held for some time before you could use it to write one batch per block from the locked sum. The larger the batches you sought to write, the longer you would need to leave that 'staked' value untouched before you could actually use it to write batches, and even then, just one per block at the size-level you staked to achieve. This would follow a curve, so that people writing very small batches may not stake anything or wait at all, whereas the staked cost/time would increase a ton as you ascended into larger batch sizes that approached the maximum cap.

Here's how it would look:

Let's say the per-op fee variable is the equivalent of $0.001 for the current block-window

- If I wanted to write under 100 ops (just an example number), I wouldn't have to stake and wait at all.
- If I wanted to write a 10,000 op batch, I might need to stake 1 million times the current Proof of Fee rate ($1000), and wait 10 blocks to start using that wallet of op funding value.
- If I wanted to write a 50,000 op batch, I might need to stake 50 million times the current Proof of Fee rate ($50,000), and wait 500 blocks to start using the locked funds for batch writing.

After the lock-up period is over, you can still only write 1 batch per block from that funding txn. Additionally, if the locked funds are moved before or any time after the locking period for anything other than to write a batch, the lock is broken and everything is reset. Any txn that breaks the rules would have its batches ignored. 

The two key outcomes should be:

1. Spammers have to publicly lock up exponentially more $ than it actually costs to generate a massive spike of large batches, which is a huge logistical and feasibility hurdle for the vast majority of attackers.
2. You can 'see spam coming', because the larger the batches you wish to write, the longer the lock time - which could be literally days before you can start writing.

OR13:
I'm a bit concerned about on chain transaction type extensions,  I think the simpler the ledger side of sidetree the better. 

[...]

PR #384, opened 2020-01-28
#271 Base functions to create freeze, refreeze, and return-to-wallet transactions

mudiali:


lib/bitcoin/BitcoinClient.ts
```diff
@@ -313,6 +325,128 @@ export default class BitcoinClient {
     return transaction;
   }
 
+  // @ts-ignore
```
isaacJChen: I assume these ts-ignores are here because they are unused right now but will be used in the future?

lib/bitcoin/BitcoinClient.ts
```diff
@@ -313,6 +325,156 @@ export default class BitcoinClient {
...
+   */
+  private async calculateTransactionFee (transaction: Transaction): Promise<number> {
+    // Get esimtated fee from RPC and add some percentage to it.
+    const estimatedFeeInKb = await this.getCurrentEstimatedFeeInSatoshisPerKb();
+    const estimatedFeeInKbWithPercentage = estimatedFeeInKb + (estimatedFeeInKb * .2);
+
+    // Estimate the size of the transaction and add some percentage to it.
+    const estimatedSizeInBytes = (transaction.inputs.length * 150) + (transaction.outputs.length * 50);
+    const estimatedSizeInBytesWithPercentage = estimatedSizeInBytes + (estimatedSizeInBytes * .2);
+    const estimatedSizeInKb = estimatedSizeInBytesWithPercentage / 1000;
```
thehenrytsai: Consider just apply one percentage at the very end, as opposed to applying percentage twice in line 337 and 341.

lib/bitcoin/BitcoinClient.ts
```diff
@@ -313,6 +325,156 @@ export default class BitcoinClient {
...
+      scriptPubKey: previousFreezeRedeemScriptHash,
+      satoshis: previousFreezeAmountInSatoshis
+    });
+
+    return frozenOutputAsUnspentOutput;
+  }
+
+  private createFreezeBitcoreScript (freezeUntilBlock: number): Script {
+    const lockBuffer = (crypto.BN as any).fromNumber(freezeUntilBlock).toScriptNumBuffer();
+    const publicKeyHashOut = Script.buildPublicKeyHashOut(this.privateKeyAddress);
```
thehenrytsai: Can we rename `privateKeyAddress` to 'walletAddress`? I got confused for a bit thinking private key is being used here somehow.

lib/bitcoin/BitcoinClient.ts
```diff
@@ -313,6 +325,156 @@ export default class BitcoinClient {
...
+      txid: previousFreezeTransaction.id,
+      vout: 0,
+      scriptPubKey: previousFreezeRedeemScriptHash,
+      satoshis: previousFreezeAmountInSatoshis
+    });
+
+    return frozenOutputAsUnspentOutput;
+  }
+
+  private createFreezeBitcoreScript (freezeUntilBlock: number): Script {
```
thehenrytsai: Consider making this a static utility method by passing in the address, as it is immediately trivial to me looking at the method why address is "not required".

lib/bitcoin/BitcoinClient.ts
```diff
@@ -313,6 +325,156 @@ export default class BitcoinClient {
...
+    previousFreezeUntilBlock: number): Transaction.UnspentOutput {
+
+    const previousFreezeAmountInSatoshis = previousFreezeTransaction.outputs[0].satoshis;
+    const previousFreezeRedeemScript = this.createFreezeBitcoreScript(previousFreezeUntilBlock);
+    const previousFreezeRedeemScriptHash = Script.buildScriptHashOut(previousFreezeRedeemScript);
+
+    const frozenOutputAsUnspentOutput = Transaction.UnspentOutput.fromObject({
+      txid: previousFreezeTransaction.id,
+      vout: 0,
+      scriptPubKey: previousFreezeRedeemScriptHash,
```
thehenrytsai: Sanity: howcome the redeem script is not part of the unspent output returned anywhere?

lib/bitcoin/BitcoinClient.ts
```diff
@@ -313,6 +325,156 @@ export default class BitcoinClient {
...
+      satoshis: previousFreezeAmountInSatoshis
```
thehenrytsai: For my understanding: I was under the impression prior to this review that there is no such thing as spending a fraction of an unspent transaction output, thus surprise that you can actually specify satoshi amount? What happens if one specify amount less than total available?

lib/bitcoin/BitcoinClient.ts
mudiali: Correct.

lib/bitcoin/BitcoinClient.ts
```diff
@@ -313,6 +325,156 @@ export default class BitcoinClient {
...
+    // First create an input from the previous frozen transaction output.
+    const frozenOutputAsInput = this.createBitcoreUnspentOutputFromFrozenTransaction(previousFreezeTransaction, previousFreezeUntilBlock);
+    const previousFreezeAmountInSatoshis = frozenOutputAsInput.satoshis;
+
+    // Now create a spend transaction using the frozen output. Create the transaction with all
+    // inputs and outputs as they are needed to calculate the fee.
+    const spendTransaction = new Transaction()
+                                   .from([frozenOutputAsInput])
+                                   .to(paytoAddress, previousFreezeAmountInSatoshis)
+                                   .lockUntilBlockHeight(previousFreezeUntilBlock);
```
thehenrytsai: For my understanding: I don't quite get this line, what does it mean to lock a transaction?

lib/bitcoin/BitcoinClient.ts
mudiali: This output is passed in as input to the spend txn. Then the input is modified to add signature and the script itself. I wasn't able to create an unspentoutput for a script and following the freeze demo was the only way I could do it.

lib/bitcoin/BitcoinClient.ts
mudiali: Lock == don't add this transaction to the blockchain until the specified block height. The txn will remain in the mempool. This is to make sure that we are not trying to spend it before we are at or past the freeze block.

lib/bitcoin/BitcoinClient.ts
```diff
@@ -313,6 +325,156 @@ export default class BitcoinClient {
...
+    // and add another one with the correct amount.
+    spendTransaction.outputs.shift();
+    spendTransaction.to(paytoAddress, previousFreezeAmountInSatoshis - transactionFee)
+                    .fee(transactionFee);
+
+    // Now update the first input to include it's signature.
+    // Create an input in the format expected for a spend-from-freeze input and set it on the
+    // transaction.
+    const previousFreezeScript = this.createFreezeBitcoreScript(previousFreezeUntilBlock);
+    const signature = (Transaction as any).sighash.sign(spendTransaction, this.privateKey, 0x1, 0, previousFreezeScript);
```
thehenrytsai: Consider declaring const variable for `0x1` and `0` before passing to give indication of what they are.

lib/bitcoin/BitcoinClient.ts
```diff
@@ -313,6 +325,156 @@ export default class BitcoinClient {
...
+    const previousFreezeAmountInSatoshis = frozenOutputAsInput.satoshis;
+
+    // Now create a spend transaction using the frozen output. Create the transaction with all
+    // inputs and outputs as they are needed to calculate the fee.
+    const spendTransaction = new Transaction()
+                                   .from([frozenOutputAsInput])
+                                   .to(paytoAddress, previousFreezeAmountInSatoshis)
+                                   .lockUntilBlockHeight(previousFreezeUntilBlock);
+
+    const transactionFee = await this.calculateTransactionFee(spendTransaction);
```
thehenrytsai: Is it possible to calculate the fee after script is added to the transaction for a better estimate.

lib/bitcoin/BitcoinClient.ts
thehenrytsai:
```suggestion
                                   .lockUntilBlockHeight(previousFreezeUntilBlock); // Transaction remains in mempool until specified block height.

```

lib/bitcoin/BitcoinClient.ts
mudiali: The math is not the same if we just multiply w/ .4 at the very end. I have moved the percentage additions to the end.

lib/bitcoin/BitcoinClient.ts
mudiali: If you specify the amount less than the total then the rest goes to the miner. I think that the library requires this so that it can do validations on the input/output/fee calculations.

lib/bitcoin/BitcoinClient.ts
mudiali: The script size does not effect the estimate. We estimate using the number of inputs & outputs.

lib/bitcoin/BitcoinClient.ts
mudiali: I updated the variable names and added comments to clarify that we are creating a transaction output and that is why we have these variable names.

PR #385, opened 2020-01-28
Ische/make bitcoin processor return complete blocks

isaacJChen:


docs/implementation.md
```diff
@@ -208,6 +208,7 @@ GET /transactions?since=170&transaction-time-hash=00000000000000000000100158f474
 ```
 
 #### Response body schema
+> Note: The transactions array will always end with a complete block of data, but can start in the middle of a block.
```
thehenrytsai:
```suggestion
The transactions array must always end with a complete block of data, but can start in the middle of a block if `since` query parameter is provided.
```

isaacJChen:
Just commenting that we need the pageSize to be greater than transactions per block for this to work. So the bitcoin config pageSize now has to be 3000+

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -175,22 +175,54 @@ export default class BitcoinProcessor {
       }
     }
 
+    // wait until at least one block is processed
+    if (this.lastProcessedBlock === undefined) {
+      return {
+        transactions: [],
+        moreTransactions: true
```
thehenrytsai: Sanity:
```suggestion
        moreTransactions: false
```

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -175,8 +174,22 @@ export default class BitcoinProcessor {
       }
     }
 
-    console.info(`Returning transactions since ${since ? 'block ' + TransactionNumber.getBlockNumber(since) : 'begining'}...`);
-    let transactions = await this.transactionStore.getTransactionsLaterThan(since, this.pageSize);
+    // wait until at least one block is processed
```
mudiali: This check should not be required. The service doesn't start even listening until it has caught up to the latest block. See the ````processTransactions(...)```` call in the ````initialize()```` function. Also, the ion\bitcoin.ts where the initialize function is called.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -99,7 +99,6 @@ export default class BitcoinProcessor {
       mongoQuantileStore);
     this.transactionSampler = new ReservoirSampler(ProtocolParameters.sampleSizePerGroup);
```
mudiali: We need a change in the ion repo as ````bitcoin-config.json```` to remove this parameter.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -704,4 +717,51 @@ export default class BitcoinProcessor {
...
+      let transactions: TransactionModel[] = await this.transactionStore.getTransactionsByTransactionTime(beginTransactionNumber, endTransactionNumber);
+
+      transactions = transactions.filter((transaction) => {
+        // filter anything greater than the last processed block because they are not complete
+        return transaction.transactionTime <= this.lastProcessedBlock!.height &&
+          // if there is a since, filter transactions that are less than or equal to since (the first block will have undesired transactions)
+          (since === undefined || transaction.transactionNumber > since);
+      });
+
+      const transactionsGroupedByTransactionTime: TransactionModel[][] = BitcoinProcessor.groupTransactionsByTransactionTime(transactions);
```
mudiali: Clarifying question: since we are flattening all the transactions eventually, why do we need to group? Maybe we can do a sort-by-transaction-time at the end to achieve the same effect?

Or is it because we are using the 'grouping' to figure out the number-of-blocks that we have gotten? In that case can we use a Set/HashSet to figure out the distinct blocks and keep a running count or something?

I have a feeling that having a flat array will be more maintainable and order wouldn't change.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -704,4 +717,51 @@ export default class BitcoinProcessor {
...
     return sidetreeTxToAdd;
   }
+
+  /**
+   * Return transactions grouped by block height. EX: [[transactions in block1], [transactions in block2], [transactions in block3]]
+   * @param since Transaction number to query since
+   * @param numOfBlocks number of blocks worth of transactions to get at least
+   */
+  private async getAtLeastPageSizeBlocksWorthOfTransactionsSinceTransactionNumber (since: number | undefined): Promise<TransactionModel[][]> {
+    let beginTransactionNumber = since === undefined ? 0 : TransactionNumber.getBlockNumber(since);
```
mudiali: This line is initializing the ````beginTransactionBlock```` or ````beginTransactionTime```` right? If yes, then the starting block should be ````this.genesisBlockNumber````. Also, if yes, then the variable names for the corresponding end also needs to change.

lib/common/MongoDbTransactionStore.ts
```diff
@@ -158,11 +158,30 @@ export default class MongoDbTransactionStore implements ITransactionStore {
...
   /**
-   * Gets a list of transactions that are in the desired transaction time
-   * @param transactionTime The transaction time to query for
+   * Gets a list of transactions between the bounds of transaction time.
+   * @param beginTransactionTime The first transaction time to query for (inclusive)
+   * @param endTransactionTime The transaction time to stop querying for (exclusive). Optional. If not provided, then only query for beginTransactionTime
    */
-  public async getTransactionsByTransactionTime (transactionTime: number): Promise<TransactionModel[]> {
-    const transactions = await this.transactionCollection!.find({ transactionTime: { $eq: Long.fromNumber(transactionTime) } }).toArray();
+  public async getTransactionsByTransactionTime (beginTransactionTime: number, endTransactionTime?: number): Promise<TransactionModel[]> {
```
mudiali: I suggest not having the ````endTransactionTime```` be optional. The function becomes much more simple.
````
start = Math.min(beginTransactionTime, endTransactionTime);
end = Math.max(beginTransactionTime, endTransactionTime);
...
````

File: lib/bitcoin/BitcoinProcessor.ts (hunk 3 above)
Comment (isaacJChen) [reply lib/bitcoin/BitcoinProcessor.ts]: Good point.

File: lib/bitcoin/BitcoinProcessor.ts (hunk 5 above)
Comment (isaacJChen) [reply lib/bitcoin/BitcoinProcessor.ts]: I am using grouping to figure out the number of blocks that we have gotten. A flat array and keeping a count for number of blocks works too. I just thought this makes the most sense conceptually, but I'm open to change it.

File: lib/bitcoin/BitcoinProcessor.ts (hunk 6 above)
Comment (isaacJChen) [reply lib/bitcoin/BitcoinProcessor.ts]: You are correct, idk why I named it transactionNumber. Good catch.

File: lib/common/MongoDbTransactionStore.ts (hunk 7 above)
Comment (isaacJChen) [reply lib/common/MongoDbTransactionStore.ts]: Using min max is a lot cleaner but it makes inclusive more arbitrary (the smaller value will always be inclusive and the larger will be exclusive), where as the current approach, start will always be inclusive and end will be exclusive, which makes it possible to have the behavior of something like "get from 10 (exclusive) to 100 (inclusive)"

Comment (isaacJChen) [reply lib/common/MongoDbTransactionStore.ts]: Gave it some thought and I think it's ok to use min max as currently we don't have a use case for looking back yet so the behavior doesn't really matter.

File: lib/common/MongoDbTransactionStore.ts [hunk 8]
```diff
@@ -158,11 +158,25 @@ export default class MongoDbTransactionStore implements ITransactionStore {
...
   /**
-   * Gets a list of transactions that are in the desired transaction time
-   * @param transactionTime The transaction time to query for
+   * Gets a list of transactions between the bounds of transaction time. The smaller value will be inclusive while the bigger be exclusive
+   * @param beginTransactionTime The first transaction time to begin querying for
+   * @param endTransactionTime The transaction time to stop querying for
    */
-  public async getTransactionsByTransactionTime (transactionTime: number): Promise<TransactionModel[]> {
-    const transactions = await this.transactionCollection!.find({ transactionTime: { $eq: Long.fromNumber(transactionTime) } }).toArray();
+  public async getTransactionsByTransactionTime (beginTransactionTime: number, endTransactionTime: number): Promise<TransactionModel[]> {
```
mudiali: Should we add ````.sort(transactionTime)```` to sort the DB return values?

tests/bitcoin/BitcoinProcessor.spec.ts
```diff
@@ -246,23 +242,135 @@ describe('BitcoinProcessor', () => {
...
+      const verifyMock = spyOn(bitcoinProcessor, 'verifyBlock' as any);
+      const lastProcessedBlockHeight = bitcoinProcessor['genesisBlockNumber'];
+      const transactions = createTransactions(BitcoinProcessor['pageSizeInBlocks'], lastProcessedBlockHeight + 1, false);
+      // make the last transaction time genesis so the transactions will all be out of bound
+      bitcoinProcessor['lastProcessedBlock'] = {
+        height: lastProcessedBlockHeight,
+        hash: 'some hash',
+        previousHash: 'previous hash'
+      };
+      const laterThanMock = spyOn(bitcoinProcessor['transactionStore'], 'getTransactionsByTransactionTime').and.callFake((() => {
```
mudiali: Just an FYI: you can do the following too instead of ````.callFake(...)````
````
spyOn(....).and.returnValue(Promise.resolve([]));
````

File: lib/bitcoin/BitcoinProcessor.ts [hunk 10]
```diff
@@ -704,4 +705,52 @@ export default class BitcoinProcessor {
...
+
+    const transactionsToReturn: TransactionModel[] = [];
+
+    // while need more blocks and have not reached the processed block
+    while (numOfBlocksAcquired < BitcoinProcessor.pageSizeInBlocks && beginTransactionTime <= this.lastProcessedBlock!.height) {
+      let transactions: TransactionModel[] = await this.transactionStore.getTransactionsByTransactionTime(beginTransactionTime, endTransactionTime);
+
+      transactions = transactions.filter((transaction) => {
+        // filter anything greater than the last processed block because they are not complete
+        return transaction.transactionTime <= this.lastProcessedBlock!.height &&
```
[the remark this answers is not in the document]
isaacJChen: <= because lastProcessedBlock is updated when a block is completely processed, which means it is good to go.

lib/common/MongoDbTransactionStore.ts
isaacJChen: I think sorting by transaction number is what we should do.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -175,9 +174,11 @@ export default class BitcoinProcessor {
       }
     }
 
-    console.info(`Returning transactions since ${since ? 'block ' + TransactionNumber.getBlockNumber(since) : 'begining'}...`);
-    let transactions = await this.transactionStore.getTransactionsLaterThan(since, this.pageSize);
```
thehenrytsai: Design: What do you think if we remove `getTransactionLaterThan()` method completely from the interface?

lib/common/MongoDbTransactionStore.ts
```diff
@@ -158,11 +158,25 @@ export default class MongoDbTransactionStore implements ITransactionStore {
...
+    let transactions: TransactionModel[] = [];
+    if (beginTransactionTime === endTransactionTime) {
+      // if begin === end or end is not provided, query for 1 transaction time
```
thehenrytsai: Sanity: currently end is always provided.

lib/common/MongoDbTransactionStore.ts
```diff
@@ -158,11 +158,25 @@ export default class MongoDbTransactionStore implements ITransactionStore {
...
+      transactions = await this.transactionCollection!.find({ transactionTime: { $eq: Long.fromNumber(beginTransactionTime) } }).toArray();
+    } else {
+      const start = Math.min(beginTransactionTime, endTransactionTime);
+      const end = Math.max(beginTransactionTime, endTransactionTime);
```
thehenrytsai: I am not sure I get the purpose of this.

lib/common/MongoDbTransactionStore.ts
thehenrytsai:
```suggestion
  public async getTransactionsStartingFrom (inclusiveBeginTransactionTime: number, exclusiveEndTransactionTime: number): Promise<TransactionModel[]> {
```

You'll need to update references if you take this change.

lib/bitcoin/BitcoinProcessor.ts
isaacJChen: Yes I like that. Right now only SpendingMonitor uses it, and it is grabbing by blocks too. I will create a new issue and separate pr for it.

lib/common/MongoDbTransactionStore.ts
isaacJChen: In case begin is greater than end. I like your comment on the function name and parameter change though. It removes the need for this min max calculation and implies that end will always be lt and begin will always be gte.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -704,4 +705,52 @@ export default class BitcoinProcessor {
     // non sidetree transaction
     return sidetreeTxToAdd;
   }
+
+  /**
+   * Return transactions since transaction number and number of blocks acquired.
+   * @param since Transaction number to query since
+   * @param numOfBlocks number of blocks worth of transactions to get at least
```
thehenrytsai: Doesn't exist.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -704,4 +705,52 @@ export default class BitcoinProcessor {
...
+   * @returns a tuple of [transactions, numberOfBlocksContainedInTransactions]
+   */
+  private async getAtLeastPageSizeBlocksWorthOfTransactionsSinceTransactionNumber (since: number | undefined): Promise<[TransactionModel[], number]> {
```
```suggestion
  private async getTransactionsSince (numberOfBlocks: number, since: number | undefined): Promise<[TransactionModel[], number]> {
```

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -704,4 +705,52 @@ export default class BitcoinProcessor {
...
+    let beginTransactionTime = since === undefined ? this.genesisBlockNumber : TransactionNumber.getBlockNumber(since);
+    let endTransactionTime = beginTransactionTime + BitcoinProcessor.pageSizeInBlocks;
+    let numOfBlocksAcquired = 0;
+
+    const transactionsToReturn: TransactionModel[] = [];
+
+    // while need more blocks and have not reached the processed block
+    while (numOfBlocksAcquired < BitcoinProcessor.pageSizeInBlocks && beginTransactionTime <= this.lastProcessedBlock!.height) {
+      let transactions: TransactionModel[] = await this.transactionStore.getTransactionsByTransactionTime(beginTransactionTime, endTransactionTime);
```
thehenrytsai: I am a bit concerned that with an `await` in the while loop, `this.lastProcessedBlock` would change as the loops goes on.

lib/common/MongoDbTransactionStore.ts
```diff
@@ -158,11 +158,23 @@ export default class MongoDbTransactionStore implements ITransactionStore {
...
+    let transactions: TransactionModel[] = [];
+    if (inclusiveBeginTransactionTime === exclusiveEndTransactionTime) {
+      // if begin === end, query for 1 transaction time
+      transactions = await this.transactionCollection!.find({ transactionTime: { $eq: Long.fromNumber(inclusiveBeginTransactionTime) } })
+      .sort({ transactionNumber: 1 }).toArray();
+    } else {
+      transactions = await this.transactionCollection!.find({ $and: [
+        { transactionTime: { $gte: Long.fromNumber(inclusiveBeginTransactionTime) } },
+        { transactionTime: { $lt: Long.fromNumber(exclusiveEndTransactionTime) } }
+      ] }).sort({ transactionNumber: 1 }).toArray();
```
mudiali: Nit: maybe move the .sort(...).toArray() outside of the if-else block so that it is only done once.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -174,23 +174,19 @@ export default class BitcoinProcessor {
...
-    const groupedTransactions = await this.getAtLeastPageSizeBlocksWorthOfTransactionsSinceTransactionNumber(since);
     // if not enough blocks to fill the page then there are no more transactions
-    const moreTransactions = groupedTransactions.length >= BitcoinProcessor.pageSizeInBlocks;
-    let transactions: TransactionModel[] = [];
-    // flatten out transactions into array of transactions instead of array of array
-    transactions = transactions.concat(...groupedTransactions);
-    // filter the results to only return transactions, and not internal data
+    const moreTransactions = numOfBlocksAcquired >= BitcoinProcessor.pageSizeInBlocks;
     transactions = transactions.map((transaction) => {
       return {
```
thehenrytsai: Sanity: Sorry I must missed something, why the `map`?

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -174,23 +174,19 @@ export default class BitcoinProcessor {
...
-    if (this.lastProcessedBlock === undefined) {
-      return {
-        transactions: [],
-        moreTransactions: false
-      };
-    }
-
     console.info(`Returning transactions since ${since ? 'block ' + TransactionNumber.getBlockNumber(since) : 'beginning'}...`);
+    const currentLastProcessedBlock = this.lastProcessedBlock!;
+    let [transactions, numOfBlocksAcquired] = await this.getTransactionsSince(since, currentLastProcessedBlock);
```
thehenrytsai: Design: I am not digging the `lastProcessedBlock` is passed in for no obvious reason externally. Is it possible to let the method assign a local variable and return it out as part of a "tripple" value instead of "tuple"?

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -719,49 +722,43 @@ export default class BitcoinProcessor {
   }
 
   /**
-   * Return transactions grouped by block height. EX: [[transactions in block1], [transactions in block2], [transactions in block3]]
+   * Return transactions since transaction number and number of blocks acquired (Will get at least pageSizeInBlocks)
    * @param since Transaction number to query since
-   * @param numOfBlocks number of blocks worth of transactions to get at least
+   * @param currentLastProcessedBlock The current last processed block. Not using this.lastProcessedBlock to avoid value changing during processing
```
thehenrytsai: Rename to something like `inclusiveLastBlock`

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -719,49 +722,43 @@ export default class BitcoinProcessor {
...
+    let numOfBlocksAcquired = 0;
 
-    const groupedTransactionsToReturn: TransactionModel[][] = [];
+    const transactionsToReturn: TransactionModel[] = [];
 
     // while need more blocks and have not reached the processed block
-    while (groupedTransactionsToReturn.length < BitcoinProcessor.pageSizeInBlocks && beginTransactionNumber <= this.lastProcessedBlock!.height) {
-      let transactions: TransactionModel[] = await this.transactionStore.getTransactionsByTransactionTime(beginTransactionNumber, endTransactionNumber);
+    while (numOfBlocksAcquired < BitcoinProcessor.pageSizeInBlocks && beginTransactionTime <= currentLastProcessedBlock.height) {
+      const endTransactionTime = beginTransactionTime + BitcoinProcessor.pageSizeInBlocks;
```
thehenrytsai: Can we rename `endTransactionTime` to `exlusiveEndTransactionTime` (and optionally `beginTransactionTime` to `inclusiveBeginTransactionTime)

lib/common/MongoDbTransactionStore.ts
```diff
@@ -158,28 +158,21 @@ export default class MongoDbTransactionStore implements ITransactionStore {
   }
 
   /**
-   * Gets a list of transactions between the bounds of transaction time.
-   * @param beginTransactionTime The first transaction time to query for (inclusive)
-   * @param endTransactionTime The transaction time to stop querying for (exclusive). Optional. If not provided, then only query for beginTransactionTime
+   * Gets a list of transactions between the bounds of transaction time. The smaller value will be inclusive while the bigger be exclusive
+   * @param beginTransactionTime The first transaction time to begin querying for
+   * @param endTransactionTime The transaction time to stop querying for
```
thehenrytsai: Param names to be updated.

tests/mocks/MockTransactionStore.ts
```diff
@@ -159,7 +159,14 @@ export default class MockTransactionStore implements ITransactionStore, IUnresol
    * Gets a list of transactions that are in the desired transaction time
    * @param transactionTime The transaction time to query for
    */
```
thehenrytsai: Remove comments.

lib/bitcoin/BitcoinProcessor.ts
isaacJChen: Using Ali's model. Pass in as the stopping point

PR #387, opened 2020-02-03
Added Map File to file structure.

thehenrytsai:
1. Added Map File to file structure.
1. Increased code coverage of `TransactionProcessor` to 100%.
1. Some code refactoring.


tests/core/AnchorFile.spec.ts
```diff
@@ -12,140 +13,146 @@ describe('AnchorFile', async () => {
...
 
-      await expectAsync(AnchorFile.parseAndValidate(anchorFileCompressed, 1))
+      // Overreid the default max operation count per batch.
+      const originalMaxOperationsPerBatch = ProtocolParameters.maxOperationsPerBatch;
+      ProtocolParameters.maxOperationsPerBatch = 1;
+
+      await expectAsync(AnchorFile.parseAndValidate(anchorFileCompressed))
         .toBeRejectedWith(new SidetreeError(ErrorCode.AnchorFileExceededMaxOperationCount));
+
+      ProtocolParameters.maxOperationsPerBatch = originalMaxOperationsPerBatch;
```
mudiali: Sanity check: should this be in 'try-finally'? Not sure if the expectAsync above fails then whether this line is executed or not.

tests/core/MapFile.spec.ts
```diff
@@ -0,0 +1,52 @@
+import Compressor from '../../lib/core/versions/latest/util/Compressor';
+import ErrorCode from '../../lib/core/versions/latest/ErrorCode';
+import JasmineSidetreeErrorValidator from '../JasmineSidetreeErrorValidator';
+import MapFile from '../../lib/core/versions/latest/MapFile';
+import MapFileModel from '../../lib/core/versions/latest/models/MapFileModel';
+import SidetreeError from '../../lib/core/SidetreeError';
+
+describe('MapFile', async () => {
```
mudiali: Sanity: there's no positive test case ... shouldn't we add one?

tests/core/AnchorFile.spec.ts
thehenrytsai: Thanks, good call!

tests/core/MapFile.spec.ts
thehenrytsai: Discuss offline, `Observer` "E2E" test already covers this path, so will skip adding for this case.

PR #392, opened 2020-02-07
Fixes for 0.6.1

isaacJChen:


lib/core/versions/latest/Operation.ts
```diff
@@ -283,6 +283,7 @@ export default class Operation {
     // Verify operation specific payload schema.
     switch (operationType) {
       case OperationType.Create:
+        decodedPayload.didDocument = JSON.parse(Encoder.decodeAsString(decodedPayload.didDocument));
```
OR13: May want to add a comment about what happens when JSON.parse fails... we should expect encoding attacks to be among the most common out of the gate.

isaacJChen: Good point, I will make sure we do that. This pr is for visibility into the changes we are making after integration test. Not final yet so feel free to comment on anything you see should change.

isaacJChen: We handle bad requests in the request handler so bad json will not cause anything dangerous. The request will be rejected.

docs/protocol.md
```diff
@@ -322,21 +322,36 @@ POST / HTTP/1.1
 #### Original DID Document example
 ```json
 {
+  "id": "someIdasdf",
   "@context": "https://w3id.org/did/v1",
```
isaacJChen: This fix the outdated documentation. We now require some fields that were not required.

lib/core/versions/latest/Operation.ts
```diff
@@ -280,22 +243,49 @@ export default class Operation {
...
+    this.encodedProtectedHeader = operation.protected;
+    this.encodedPayload = operation.payload;
+    this.signature = operation.signature;
+    this.operationHash = Operation.computeHash(this.encodedPayload);
+
+    // Verify operation specific payload schema and further decode if needed, then initialize.
+    switch (this.type) {
       case OperationType.Create:
+        // additional parsing required because did doc is nest base64url encoded
+        decodedPayload.didDocument = JSON.parse(Encoder.decodeAsString(decodedPayload.didDocument));
```
isaacJChen: This is a fix for did doc parsing as we added another layer of encoding and wasn't handling it.

lib/core/versions/latest/TransactionSelector.ts
```diff
@@ -66,8 +66,12 @@ export default class TransactionSelector implements ITransactionSelector {
     let numberOfOperations = 0;
     if (transactions) {
       for (const transaction of transactions) {
-        const numOfOperationsInCurrentTransaction = AnchoredDataSerializer.deserialize(transaction.anchorString).numberOfOperations;
-        numberOfOperations += numOfOperationsInCurrentTransaction;
+        try {
```
isaacJChen: The try catch blocks prevent invalid anchorstring to break throw error.

lib/core/versions/latest/TransactionSelector.ts
```diff
@@ -66,8 +66,12 @@ export default class TransactionSelector implements ITransactionSelector {
...
+          const numOfOperationsInCurrentTransaction = AnchoredDataSerializer.deserialize(transaction.anchorString).numberOfOperations;
+          numberOfOperations += numOfOperationsInCurrentTransaction;
+        } catch {
+          console.debug(`Unable to deserialize ${transaction.anchorString}, transaction not considered as selected.`);
```
mudiali: Please include the exception details as it might be good to know what happened. You can capture the relevant details of the exception ````e```` with: ````JSON.stringify(e, Object.getOwnPropertyNames(e))````

lib/core/versions/latest/TransactionSelector.ts
```diff
@@ -89,11 +93,14 @@ export default class TransactionSelector implements ITransactionSelector {
...
-      if (numberOfOperationsSeen <= numberOfOperationsToQualify) {
-        transactionsToReturn.push(currentTransaction);
+      try {
+        const numOfOperationsInCurrentTransaction = AnchoredDataSerializer.deserialize(currentTransaction.anchorString).numberOfOperations;
+        numberOfOperationsSeen += numOfOperationsInCurrentTransaction;
+        if (numberOfOperationsSeen <= numberOfOperationsToQualify) {
+          transactionsToReturn.push(currentTransaction);
+        }
+      } catch {
+        console.debug(`Unable to deserialize ${currentTransaction.anchorString}, transaction not selected`);
```
mudiali: 1. Please include the exception details (same as above).
1. Maybe we should change the message to be more generic as we are doing more things than just deserializing within try-catch. For example: An unexpected error occurred during processing; transaction is not selected. Error: ...

tests/core/TransactionSelector.spec.ts
```diff
@@ -205,6 +205,28 @@ describe('TransactionSelector', () => {
       ];
       expect(result).toEqual(expected);
     });
+
```
mudiali: Shouldn't we have new tests for the changes in the functions ````getNumberOfOperationsAndTransactionsAlreadyInBlock ```` and ````getHighestFeeTransactionsFromCurrentTransactionTime ```` and not this one? The following test will exercise the error in the first function but not the 2nd one right?

isaacJChen: It will go into both errors because the transactions passed in as argument will trigger one, and the transactions read from the DB will trigger another. I didn't add tests for the individual functions because they are private and I am able to cover the branches by testing the public function so I did it that way. I can add tests for the private functions too though.

mudiali: Personally I like testing the smaller functions as it is easier to add variations to the test. If there's a change in the private function, we will have to change the inputs here to match the both the new variation and the other private function as well.

Also since our unit tests are white box, it is ok to test the private functions separately.

But you have already written the tests to test all branches I am ok with this change.

PR #393, opened 2020-02-07
#271 Implement lock verification functionality

mudiali:
Second set of changes for the value locking issue. Changes include:
1. Lock resolver to verify and fetch lock information
1. Lock identifier model and its serialization support
1. Some changes are for a 'lock monitor' coming next
1. Renaming BitcoinLockTransaction* to SavedLockTransaction* to more accurately describe the usage of these classes.


lib/bitcoin/BitcoinClient.ts
```diff
@@ -89,14 +92,44 @@ export default class BitcoinClient {
...
+  }
 
-    return this.rpcCall(request, true);
+  /**
+   * Broadcasts the specified lock transaction.
+   *
+   * @param bitcoinLockTransaction The transaction object.
+   */
+  public async broadcastLockTransaction (bitcoinLockTransaction: BitcoinLockTransactionModel): Promise<string> {
+    if (!(bitcoinLockTransaction.transactionObject instanceof Transaction)) {
```
isaacJChen: Why not just force it to be a Transaction object instead of Any?

mudiali: Don't want to expose the bitcore-lib object outside of this class really.

lib/bitcoin/BitcoinClient.ts
```diff
@@ -233,6 +266,13 @@ export default class BitcoinClient {
     return (inputSatoshiSum - outputSatoshiSum);
   }
 
+  /**
+   * Gets the public address for the linked wallet.
+   */
+  public getWalletAddressAsBuffer (): Buffer {
```
thehenrytsai: Sanity: doesn't look like used? Remove if that's the case.

lib/bitcoin/BitcoinClient.ts
```diff
@@ -89,14 +92,44 @@ export default class BitcoinClient {
...
 
-    return this.rpcCall(request, true);
+  /**
+   * Broadcasts the specified lock transaction.
+   *
+   * @param bitcoinLockTransaction The transaction object.
+   */
+  public async broadcastLockTransaction (bitcoinLockTransaction: BitcoinLockTransactionModel): Promise<string> {
+    if (!(bitcoinLockTransaction.transactionObject instanceof Transaction)) {
+      throw new Error('Transaction property is not of bitcore-lib.Transaction');
```
thehenrytsai: Discussion: should we start considering throwing specific typed errors (e.g. SidetreeError), or their is no point in this layer?

lib/bitcoin/BitcoinClient.ts
```diff
@@ -89,14 +92,44 @@ export default class BitcoinClient {
...
+   * Creates (and NOT broadcast) a lock transaction using the funds for the linked wallet.
+   *
+   * NOTE: if the linked wallet outputs are spent then this transaction cannot be broadcasted. So broadcast
+   * this transaction before spending from the wallet.
+   *
+   * @param lockAmountInSatoshis The amount to lock.
+   * @param lockUntilBlock  The block until the amount to lock to; the amount becomes spendable AT this block.
+   */
+  public async createLockTransaction (lockAmountInSatoshis: number, lockUntilBlock: number): Promise<BitcoinLockTransactionModel> {
+    const unspentCoins = await this.getUnspentOutputs(this.walletAddress);
```
thehenrytsai: Consider merging this function with `createFreezeTransaction()` method.

lib/bitcoin/enums/SavedLockTransactionType.ts
```diff
@@ -0,0 +1,7 @@
+enum SavedLockTransactionType {
```
thehenrytsai: Nit: maybe `LockTransactionType` is sufficient?

lib/bitcoin/lock/LockResolver.ts
```diff
@@ -0,0 +1,139 @@
...
+      amountLocked: lockTransaction.outputs[0].satoshis,
+      lockEndTransactionTime: lockUntilBlock,
+      linkedWalletAddress: walletAddressObj.toString()
+    };
+  }
+
+  /**
+   * Checks whether the redeem script is indeed a lock script.
+   * @param redeemScript The script to check.
+   * @returns A touple where [0] == true if the script is a lock script; false otherwise, [1] == if [0] is true then the lockUntilBlock.
```
thehenrytsai:
```suggestion
   * @returns A touple where [0] = true and [1] = lock-until-block if the script is a lock script; [0] = false and [1] = 0 otherwise.
```

lib/bitcoin/lock/LockResolver.ts
```diff
@@ -0,0 +1,139 @@
...
+    const scriptAsmParts = redeemScript.toASM().split(' ');
+
+    // Verify different parts
+    const isScriptValid =
+      scriptAsmParts.length === 8 &&
+      scriptAsmParts[1] === 'OP_NOP2' &&
+      scriptAsmParts[2] === 'OP_DROP' &&
+      scriptAsmParts[3] === 'OP_DUP' &&
+      scriptAsmParts[4] === 'OP_HASH160' &&
+      scriptAsmParts[6] === 'OP_EQUALVERIFY' &&
```
thehenrytsai: Nit: consider adding a comment on `scriptAsParts[5]` being verified later.

lib/bitcoin/lock/LockResolver.ts
```diff
@@ -0,0 +1,139 @@
...
+
+  /**
+   * Checks whether the specified redeem script is paying to the target wallet address.
+   * @param redeemScript The redeem script representing a lock script.
+   * @param targetWalletAddress The wallet address to be verified.
+   */
+  private static isRedeemScriptPayingToTargetWallet (redeemScript: Script, targetWalletAddress: Address): boolean {
+    // Convert the wallet address into the standard "publickeyhash" bitcoin output
+    const publicKeyHashOutput = Script.buildPublicKeyHashOut(targetWalletAddress);
+    const publicKeyHashOutputAsHex = publicKeyHashOutput.toHex();
```
thehenrytsai: Nit^10: Consider adding "Script" to the variable names. e.g. publicKeyHashOutputScript.

lib/bitcoin/lock/LockResolver.ts
```diff
@@ -0,0 +1,139 @@
...
+
+    // If the above output is the suffix in the redeem script then it is paying to the target wallet
+    return redeemScript.toHex().endsWith(publicKeyHashOutputAsHex);
+  }
+
+  private static createScriptFromHexInput (redeemScriptAsHex: string): Script {
```
thehenrytsai: Nit: "Input" sounds like transaction input.

```suggestion
  private static createScript (redeemScriptAsHex: string): Script {

```

lib/bitcoin/lock/MongoDbLockTransactionStore.ts
```diff
@@ -1,4 +1,4 @@
-import BitcoinLockTransactionModel from './models/BitcoinLockTransactionModel';
+import BitcoinLockTransactionModel from '../models/SavedLockTransactionModel';
```
thehenrytsai: Maybe just `LockTransactionModel`?

lib/bitcoin/lock/LockResolver.ts
```diff
@@ -0,0 +1,139 @@
...
+   * @returns The blockchain lock model if the specified identifier is verified; throws if verification fails.
+   */
+  public async resolveLockIdentifierAndThrowOnError (lockIdentifier: LockIdentifier): Promise<BlockchainLockModel> {
+
+    // The verifictation of a lock-identifier has the following steps:
+    //   (A). The redeem script in the lock-identifier is actually a 'locking' script
+    //   (B). The redeem script in the lock-identifier is paying to the wallet in the lock-identifier
+    //   (C). The transaction in the lock-identifier is paying to the redeem script in the lock-identifier
+    //
+    // With above, we can verify that the amount is/was locked for the specified wallet in
```
isaacJChen: This function is not accessing the amount. It is only verifying A, B and C, how do we verify that the amount is correct and if the lock is still active or not?

lib/common/models/BlockchainLockModel.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * Encapsulates data about locks created and retrieved by the blockchain service.
+ */
+export default interface BlockchainLockModel {
```
thehenrytsai:
```suggestion
export default interface TimeLockModel {

```

lib/bitcoin/lock/LockResolver.ts
isaacJChen: Or is it core's job to do that?

lib/common/models/BlockchainLockModel.ts
```diff
@@ -0,0 +1,16 @@
...
+ */
+export default interface BlockchainLockModel {
+  /** Uniquely identifies a lock */
+  identifier: string;
+
+  /** The amount that is locked */
+  amountLocked: number;
+
+  /** At this transaction time the lock is no longer valid */
+  lockEndTransactionTime: number;
```
thehenrytsai: `lockEndTransactionTime: lockUntilBlock` in `LockResolver` is a bit unclear because `lockUntilBlock` seems to be inclusive. 

Consider `unlockTransactionTime` here and `unlockAtBlock` in `LockResolver`.

lib/common/models/BlockchainLockModel.ts
```diff
@@ -0,0 +1,16 @@
...
+
+  /** The wallet where the amount goes upon unlock */
+  linkedWalletAddress: string;
```
thehenrytsai: If this is contract with core, then consider naming `owner`.

tests/JasmineSidetreeErrorValidator.ts
```diff
@@ -50,4 +51,50 @@ export default class JasmineSidetreeErrorValidator {
...
     }
   }
+
+  /**
+   * Fails the current spec if the execution of the function does not throw the expected BitcoinError.
+   *
+   * @param functionToExcute The function to execute.
+   * @param expectedBitcoinErrorCode The expected Bitcoin error code.
+   */
+  public static expectBitcoinErrorToBeThrown (functionToExcute: () => any, expectedBitcoinErrorCode: string): void {
```
thehenrytsai: Would Jasmine's expect(...).toThrow(...) be sufficient?

e.g.
```javascript
expect(() => { Operation.validateUpdatePayload(updatePayload); }).toThrow(new SidetreeError(ErrorCode.OperationUpdatePayloadMissingOrUnknownProperty));
```

tests/bitcoin/BitcoinClient.spec.ts
```diff
@@ -302,6 +362,37 @@ describe('BitcoinClient', async () => {
...
+    });
+
+    it('should throw if the RPC call fails.', async (done) => {
+      const mockRawTransaction = 'mocked-raw-transaction';
+      const mockRpcOutput = 'mockRpcOutput';
+
+      const spy = mockRpcCall('sendrawtransaction', [mockRawTransaction], mockRpcOutput);
+      spy.and.throwError('test');
+      try {
+        await await bitcoinClient['broadcastTransactionRpc'](mockRawTransaction);
```
thehenrytsai: Sanity: `await await` intended?

lib/bitcoin/lock/LockResolver.ts
mudiali: @isaacJChen, you are right that the core will do that.

lib/bitcoin/BitcoinClient.ts
mudiali: This specific instance is being removed in the next step as we want to save the fully serialized txn so this check becomes obsolete.

tests/bitcoin/BitcoinClient.spec.ts
mudiali: Oops .. not even sure what it means :)

lib/bitcoin/lock/LockResolver.ts
mudiali: Code is being removed.

lib/bitcoin/BitcoinClient.ts
mudiali: Removing this as we are no longer going to be using the wallet address as the 'writer'. Instead it will be the publickeyhash (bitcoin's Hash160) which I will add in the future.

tests/JasmineSidetreeErrorValidator.ts
mudiali: Jasmine's .toThrow(...) does a toString() to the error for comparison. And our errors sometimes also have a message appended which causes the comparison to fail. That was the original reason behind adding this utility class.

lib/bitcoin/BitcoinClient.ts
mudiali: Will address it in the next iteration.

lib/common/models/BlockchainLockModel.ts
```diff
@@ -9,8 +9,8 @@ export default interface BlockchainLockModel {
   amountLocked: number;
 
   /** At this transaction time the lock is no longer valid */
-  lockEndTransactionTime: number;
+  unlockTransactionTime: number;
 
-  /** The wallet where the amount goes upon unlock */
-  linkedWalletAddress: string;
+  /** The destination where the amount goes upon unlock */
```
thehenrytsai: Consider to describe it generally without referring to implementation detail.
```suggestion
  /** The owner of the lock. */

```

lib/bitcoin/lock/LockResolver.ts
```diff
@@ -3,9 +3,19 @@ import BitcoinError from '../BitcoinError';
...
 import LockIdentifierSerializer from './LockIdentifierSerializer';
-import { Script, Address } from 'bitcore-lib';
+import { Script } from 'bitcore-lib';
+
+/** Structure (internal for this class) to hold the redeem script verification results */
+interface LockScriptVerifyResult {
+  /** whether or not the script was valid */
+  isScriptValid: boolean;
+  /** the target address when the script unlcoks; undefined if the script is not valid. */
+  publicKeyHashOut: string | undefined;
```
thehenrytsai: Nit: I confused this as a script as I thought you mentioned that xxxOut typically refers to the "output script"? Such as `buildPublicKeyHashOut` and `targetScriptHashOut`.

```suggestion
  publicKeyHash: string | undefined;

```

lib/bitcoin/lock/LockResolver.ts
```diff
@@ -3,9 +3,19 @@ import BitcoinError from '../BitcoinError';
...
+import LockIdentifierModel from '../models/LockIdentifierModel';
 import LockIdentifierSerializer from './LockIdentifierSerializer';
-import { Script, Address } from 'bitcore-lib';
+import { Script } from 'bitcore-lib';
+
+/** Structure (internal for this class) to hold the redeem script verification results */
+interface LockScriptVerifyResult {
+  /** whether or not the script was valid */
+  isScriptValid: boolean;
+  /** the target address when the script unlcoks; undefined if the script is not valid. */
```
thehenrytsai:
```suggestion
  /** the public key hash of the target address when the script unlcoks; undefined if the script is not valid. */

```

lib/common/models/BlockchainLockModel.ts
mudiali: Changing it to ValueTimeModel.

PR #401, opened 2020-02-15
Updated long-form DID construction.

thehenrytsai:


lib/core/versions/latest/Did.ts
```diff
@@ -53,48 +55,45 @@ export default class Did {
     }
 
     this.shortForm = didMethodName + this.uniqueSuffix;
+  }
+
+  /**
+   * Parses the input string as Sidetree DID.
+   * @param did Short or long-form DID string.
```
isaacJChen: Param doc is not matching the actual function parameters.

lib/core/versions/latest/CreateOperation.ts
```diff
@@ -34,15 +34,25 @@ export default class CreateOperation implements OperationModel {
...
 
+  /** Encoded string of the suffix data. */
+  public readonly encodedSuffixData: string;
+
   /** Operation data. */
   public readonly operationData: OperationDataModel;
 
   /**
    * Constructs an Operation if the operation buffer passes schema validation, throws error otherwise.
+   * NOTE: should only be used by `parse()` and `parseObject()`. because
```
isaacJChen: Because... what? haha

lib/core/versions/latest/RequestHandler.ts
```diff
@@ -189,11 +180,26 @@ export default class RequestHandler implements IRequestHandler {
 
     // The code reaches here if this DID is not registered on the ledger.
 
-    didDocument = await Document.fromLongFormDid(did);
+    const document = await this.applyCreateOperation(did.createOperation!);
```
isaacJChen: This is not a question specific to this pr, but this pr got me thinking how I'd use long form, so I thought of this. 

What is preventing me from just always use long form so I don't have to pay the fee to put my document onto a ledger? So say I make a valid operationRequest paying the bare minimum fee to avoid cost, that will allow me to keep using my long form and never having to worry if my document is actually anchored.

csuwildcat: This is correct - you can instantly use the DIDs you generate via their long form ID variant, and continue doing so until you need to perform a DID Doc modification (roll keys, change endpoints, etc.). Instant use and registration load reduction were the two main reasons we created long form ID variants in the first place. IDs for some types of things (i.e. inanimate objects, IOT devices) may never be registered at all, and go their entire service lifetimes using the long form ID variant.

lib/core/versions/latest/Did.ts
thehenrytsai: Thanks, fixed.

lib/core/versions/latest/Did.ts
```diff
@@ -53,48 +55,45 @@ export default class Did {
...
 
       if (!didParameterString.startsWith(Did.initialValuesParameterPrefix)) {
         throw new SidetreeError(ErrorCode.DidLongFormOnlyInitialValuesParameterIsAllowed);
       }
 
-      // Trim the `initial-values=` string to get the full initial DID DOcument.
-      this.encodedDidDocument = didParameterString.substring(Did.initialValuesParameterPrefix.length);
+      // Trim the `initial-values=` string to get the encoded create operation request body.
+      const encodedCreateRequest = didParameterString.substring(Did.initialValuesParameterPrefix.length);
+      const createRequesBuffer = Encoder.decodeAsBuffer(encodedCreateRequest);
```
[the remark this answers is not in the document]
thehenrytsai: Fixed.

tests/core/CreateOperation.spec.ts
```diff
@@ -0,0 +1,123 @@
+import ErrorCode from '../../lib/core/versions/latest/ErrorCode';
```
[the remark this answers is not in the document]
thehenrytsai: Thanks for noticing, fixed.

lib/core/versions/latest/CreateOperation.ts
```diff
@@ -34,15 +34,25 @@ export default class CreateOperation implements OperationModel {
...
    */
-  private constructor (operationBuffer: Buffer, didUniqueSuffix: string, suffixData: SuffixDataModel, operationData: OperationDataModel) {
+  private constructor (
+    operationBuffer: Buffer,
+    didUniqueSuffix: string,
+    encodedSuffixData: string,
+    suffixData: SuffixDataModel,
+    operationData: OperationDataModel) {
     this.type = OperationType.Create;
```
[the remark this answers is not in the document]
thehenrytsai: Yes, it is defined in the `OperationModel`.

PR #402, opened 2020-02-15
#271 Adding functions to the bitcoin client needed for the lockmonitor.

mudiali:
Added following functionality to the BitcoinClient for the lock monitor changes coming next:
1. Functions to create lock, relock, and releaselock transactions.
1. Function to broadcast the lock transactions.
1. Function to return the public key hash of the wallet.

Also, added a couple of fields to the BitcoinLockTransactionModel object.

### DIFF
--- a/lib/bitcoin/BitcoinClient.ts
+++ b/lib/bitcoin/BitcoinClient.ts
@@ -467,12 +467,14 @@ export default class BitcoinClient {
     const payToScriptHashOutput = Script.buildScriptHashOut(freezeScript);
     const payToScriptAddress = new Address(payToScriptHashOutput);
 
-    const freezeTransaction = await this.createSpendTransactionFromFrozenTransaction(
+    // We are creating a spend transaction and are paying to another freeze script.
+    // So essentially we are re-freezing ...
+    const reFreezeTransaction = await this.createSpendTransactionFromFrozenTransaction(
       previousFreezeTransaction,
       previousFreezeUntilBlock,
       payToScriptAddress);
 
-    return [freezeTransaction, freezeScript.toHex()];
+    return [reFreezeTransaction, freezeScript.toHex()];
   }
 
   private async createSpendToWalletTransaction (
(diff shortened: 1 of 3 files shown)


lib/bitcoin/BitcoinClient.ts
```diff
@@ -99,18 +107,12 @@ export default class BitcoinClient {
...
   public async broadcastLockTransaction (bitcoinLockTransaction: BitcoinLockTransactionModel): Promise<string> {
-    if (!(bitcoinLockTransaction.transactionObject instanceof Transaction)) {
-      throw new Error('Transaction property is not of bitcore-lib.Transaction');
-    }
-
-    const transaction: Transaction = bitcoinLockTransaction.transactionObject;
-    const rawTransaction = transaction.serialize();
 
-    return this.broadcastTransactionRpc(rawTransaction);
+    return this.broadcastTransactionRpc(bitcoinLockTransaction.serializedTransactionObject);
```
isaacJChen: Why do we need this? To hide the BitcoinLockTransactionModel's implementation detail?

mudiali: We want to rebroadcast the full transaction for the following scenario:
1. The service created a lock txn.
1. It saved the info to the DB
1. The service crashed before it could broadcast.
1. The service comes back up, sees the txn in the DB but not on the blockchain. So it just rebroadcast.

We don't want to broadcast-first-and-then-save. Because if the service crashes before saving then we don't know that there's a lock and may lose that info.

lib/bitcoin/BitcoinClient.ts
```diff
@@ -410,23 +458,23 @@ export default class BitcoinClient {
...
     previousFreezeUntilBlock: number,
-    freezeUntilBlock: number): Promise<Transaction> {
+    freezeUntilBlock: number): Promise<[Transaction, string]> {
 
     const freezeScript = BitcoinClient.createFreezeScript(freezeUntilBlock, this.walletAddress);
     const payToScriptHashOutput = Script.buildScriptHashOut(freezeScript);
     const payToScriptAddress = new Address(payToScriptHashOutput);
 
-    return this.createSpendTransactionFromFrozenTransaction(
+    const freezeTransaction = await this.createSpendTransactionFromFrozenTransaction(
```
isaacJChen: This is confusing. We are creating a spendTransaction but assign it to const freezeTransaction.

mudiali: Good point .. will update.

Issue #223, opened 2019-05-23
Implement better method of fetching IPFS content size without downloading it


thehenrytsai:
Currently `cumulativeSize` returned by `IPFS.object.stat()` is used but it is not precise, replace the implementation.

csuwildcat:
I have a request for feedback in with my buddy at Protocol Labs (devs of IPFS)

thehenrytsai:
https://github.com/libp2p/interface-peer-discovery#modules-that-implement-the-interface
https://github.com/ipfs/js-ipfs-unixfs-exporter#example
https://github.com/ipfs/js-ipfs/blob/master/src/core/components/cat.js#L28

csuwildcat:
I just want to make sure we remain aware that any metadata convenience methods IPFS provides to report the size before download are not trustable, as the file creator/sender can misreport the values. The only thing that it will catch is files that honestly state they are too large, so all the code to check the real size is still required as we do the actual download.

isaacJChen:
@csuwildcat Yes. We will check the reported size: if exceeds limit, skips download || if within limit, start downloading by chunks, for each chunk, check current cumulative size and stops if size exceeded.

PR #404, opened 2020-02-19
Ische/kill ipfs download on large content size

isaacJChen:
- ipfs stops downloading if file is too big
- ipfsStorage test coverage  100%

https://github.com/decentralized-identity/sidetree/issues/223


lib/ipfs/RequestHandler.ts
```diff
@@ -18,13 +18,23 @@ export default class RequestHandler {
 
   private serviceInfo: ServiceInfo;
 
+  /**
+   * initializes an instance of request handler
+   * @param fetchTimeoutInSeconds Timeout for fetch request. Fetch request will return `not-found` when timed-out.
+   * @param repo Optional IPFS datastore implementation.
+   */
+  public static async initialize (fetchTimeoutInSeconds: number, repo?: any) {
```
mudiali: Isn't ````create```` a better name for this function? 

Also, the ion repo needs to call this new function now right?

lib/ipfs/@types/ipfs/index.d.ts
```diff
@@ -5,188 +5,173 @@ declare module 'ipfs' {
...
-    once(event: string, callback: () => void): IPFS;
- 
-    createAddStream(options: any, callback: Callback<any>): void;
-    createAddStream(callback: Callback<any>): void;
+  on (event: string, callback: () => void): IPFS;
+  on (event: 'error', callback: (error: { message: any }) => void): IPFS;
+  once (event: string, callback: () => void): IPFS;
 
-    createPullStream(options: any): any;
+  add (data: IPFS.FileContent, options?: any): AsyncIterator<any>;
```
mudiali: sanity: wondering why does ````add```` return an AsyncIterable<...>? Sounds like it should return a ````Promise<void>```` maybe?

isaacJChen: It is an iterator because add can return multiple results. It happens when you are writing a nested file structure like `dir1/dir2/file.txt`. It will first return the add result for dir1, then dir2, then file.txt

lib/ipfs/RequestHandler.ts
isaacJChen: Yes ION will have to call this instead of instantiation. Is create the verb? I see both create and initialize in the code base. Not sure if we designate them to certain behaviors.

lib/ipfs/IpfsStorage.ts
```diff
@@ -83,65 +71,29 @@ export default class IpfsStorage {
...
-    });
-
-    await fetchContent;
-
-    if (fetchResult.code === FetchResultCode.Success) {
-      fetchResult.content = Buffer.concat(bufferChunks);
+      }
+    } catch (e) {
+      // when an error is thrown, that means the hash points to something that is not a file
+      console.log(`Error thrown while downloading content from IPFS: ${e}`);
```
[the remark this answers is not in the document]
isaacJChen: 1. There's only error message so we'd have to compare the message if we want to catch specific error. I've looked at the code and tested it, and as far as I know, the only possible errors are caused by not a file.

2. Good idea!

lib/ipfs/IpfsStorage.ts
```diff
@@ -45,32 +45,20 @@ export default class IpfsStorage {
...
-
-    // NOTE: IPFS API does not have an API for finding out only the content size,
-    // IPFS API only provides the "cumulative size" that includes the additional metadata on how the Merkle DAG is formed.
-    // So here we account for the additional space used by performing a more lenient max size check by 10%,
-    // we will track the exact conent size of content later when we fetch the content if the content passes this size check.
-    const adjustedMaxSize = maxSizeInBytes * 1.1;
-    if (contentMetadata.CumulativeSize > adjustedMaxSize) {
-      console.info(`Cumulative size of ${contentMetadata.CumulativeSize} bytes is greater than the ${adjustedMaxSize} cumulative size limit.`);
+    if (contentMetadata.DataSize > maxSizeInBytes) {
+      console.info(`Size of ${contentMetadata.DataSize} bytes is greater than the ${maxSizeInBytes} data size limit.`);
```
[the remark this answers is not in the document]
isaacJChen: From what I can tell reading the tests for stat, it will always return a value, but I could be wrong. I will add back the check just to be safe. I think it is good to have and not assume.

lib/ipfs/IpfsStorage.ts
```diff
@@ -15,21 +15,21 @@ export default class IpfsStorage {
   /**
    * Static method to have a single instance of class and mock in unit tests
    */
-  public static create (repo?: any): IpfsStorage {
+  public static async create (repo?: any): Promise<IpfsStorage> {
```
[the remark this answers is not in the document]
isaacJChen: Good idea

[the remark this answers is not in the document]
isaacJChen: Hmm I think we should treat it as a singleton. When I was writing tests, I accidentally created multiple instances of IPFS and the library didn't like that. I can implement something like

`If instance exists, log message and print out current repo, then return existing instance`

lib/ipfs/@types/ipfs/index.d.ts
mudiali: Ahh .. got it.

lib/ipfs/RequestHandler.ts
mudiali: In my mind, ````initialize```` is for when the objected is already 'created' and needs to be 'initialized'. Every call to ````initialize```` is like calling a function on an existing object. Whereas ````create```` returns a new instance of the object which the caller needs to hold and dispose off as needed. Every call to ````create```` returns a new instance of an object (which is what we are doing here.

Another thought actually, looks like the constructor doesn't need to be public right? Small nit though so up to you.

lib/ipfs/IpfsStorage.ts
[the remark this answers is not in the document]
mudiali: If we want this to be an instance then we should remove the parameter right? Otherwise, for each parameter input, we need to create and save that instance (which doesn't make sense to me but maybe it does .. ?)

lib/common/AsyncTimeout.ts
```diff
@@ -0,0 +1,46 @@
+/**
```
mudiali: Few suggestions:
1. The class name to something like: ````AsyncExecutor````?
1. The function name to something like: ````ExecuteWithTimeout````?
1. The signature: ````public static async ExecuteWithTimeout<T> (asyncCall: Promise<T>, timeoutInMilliseconds: number): Promise<T>````
1. Upon reaching timeout, the function just throws ````AsyncTimeoutError````. Otherwise, T is returned. Or I guess we can return the ````AsyncExecuteWithTimeoutResult```` if we want to stick with your current model.
1. For any other errors, an ````IpfsError```` is thrown?
1. All the related classes for this functionality is under ````async```` folder.

lib/ipfs/RequestHandler.ts
```diff
@@ -18,13 +18,28 @@ export default class RequestHandler {
...
+   * create an instance of request handler
+   * @param fetchTimeoutInSeconds Timeout for fetch request. Fetch request will return `not-found` when timed-out.
+   * @param repo Optional IPFS datastore implementation.
+   */
+  public static async create (fetchTimeoutInSeconds: number, repo?: any) {
+    let ipfsStorage: IpfsStorage;
+    try {
+      ipfsStorage = await IpfsStorage.createSingleton(repo);
+    } catch {
+      ipfsStorage = IpfsStorage.getSingleton();
```
mudiali: We should throw a specific exception-error-code in createSingleton and only call getSingleton() if that exception is thrown. Other exceptions should be just bubbled up.

lib/common/AsyncTimeout.ts
```diff
@@ -0,0 +1,46 @@
...
+    return Promise.race([startTiming, asyncCall])
+      .then((res) => {
+        // if resolved, clear the timeout and return
+        clearTimeout(timeoutReference);
+        return { timedOut: false, result: res };
+      }).catch((e) => {
+        clearTimeout(timeoutReference);
+        // if thrown, determine if it's a timeout error or async call threw and error
+        if (e instanceof TimeoutError) {
+          return { timedOut: true, result: undefined };
```
mudiali: Do we know if the ````asyncCall```` promise is rejected when we return here?

isaacJChen: Async.race throws away the promise. Whether or not ipfs will have background processes running is TBD until we get a response from the IPFS folks.

lib/ipfs/IpfsStorage.ts
```diff
@@ -83,64 +92,43 @@ export default class IpfsStorage {
...
-        if (currentContentSize > maxSizeInBytes) {
-          console.info(`Content stream reached ${currentContentSize} bytes which is greater than the ${maxSizeInBytes} bytes limit.`);
-          fetchResult.code = FetchResultCode.MaxSizeExceeded;
+    let iterator: AsyncIterator<Buffer>;
+    try {
+      iterator = this.node.cat(hash);
+    } catch (e) {
+      // when an error is thrown, that means the hash points to something that is not a file
+      console.error(`Error thrown while downloading content from IPFS: ${JSON.stringify(e, Object.getOwnPropertyNames(e))}`);
+      return { code: FetchResultCode.NotAFile };
```
mudiali: An uber point about these helper function such as this ````fetchContent(...)````. Should we throw IpfsError w/ specific error codes here? For example, we could throw ````new IpfsError.createFromError(ErrorCode.IpfsStorageFetchFailed, e));```` and the outer layer will catch and return the appropriate ````FetchResult````. It seems a little weird that these inner helper functions are returning the FetchResult. 

This is just a thought btw and I'm open for discussion.

lib/ipfs/IpfsStorage.ts
```diff
@@ -83,64 +92,43 @@ export default class IpfsStorage {
...
+        result = await AsyncExecutor.executeWithTimeout(iterator.next(), IpfsStorage.timeoutDuration);
+      } catch (e) {
+        if (e instanceof AsyncTimeoutError) {
+          return { code: FetchResultCode.NotFound };
+        }
+        // this should never happen
+        console.error(`unexpected error thrown, please investigate and fix: ${e}`);
+        throw e;
+      }
+      if (result.value instanceof Buffer) {
```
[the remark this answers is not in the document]
isaacJChen: No, because it could be [buffer, buffer, undefined]

lib/ipfs/IpfsStorage.ts
```diff
@@ -83,64 +92,43 @@ export default class IpfsStorage {
...
 
-          readableStream.destroy();
-          return;
+    let result: IteratorResult<Buffer>;
+    do {
+      try {
+        result = await AsyncExecutor.executeWithTimeout(iterator.next(), IpfsStorage.timeoutDuration);
+      } catch (e) {
+        if (e instanceof AsyncTimeoutError) {
+          return { code: FetchResultCode.NotFound };
```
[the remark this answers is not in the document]
isaacJChen: Good idea, I think I want to add it in the AsyncExecutor itself though, what do you think?

lib/ipfs/IpfsStorage.ts
isaacJChen: https://github.com/decentralized-identity/sidetree/issues/415

Created issue for future work.

lib/ipfs/IpfsStorage.ts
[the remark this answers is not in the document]
isaacJChen: Decided to add outside as discussed.

lib/ipfs/ErrorCode.ts
```diff
@@ -0,0 +1,4 @@
+export default {
+  ipfsStorageInstanceCanOnlyBeCreatedOnce: 'ipfs_storage_instance_can_only_be_created_once',
```
thehenrytsai: Nit: other Error code files uses PascalCasing.

lib/ipfs/IpfsError.ts
```diff
@@ -0,0 +1,22 @@
+/**
+ * Standardized error class for throwing generic errors internal to this project.
+ */
+export default class IpfsError extends Error {
```
thehenrytsai: Consider creating a bug to consolidating `IpfsError`, `BitcionError` and `SidetreeError` into one single class under `common`.

lib/ipfs/IpfsStorage.ts
```diff
@@ -45,32 +61,24 @@ export default class IpfsStorage {
...
 
-    // NOTE: IPFS API does not have an API for finding out only the content size,
-    // IPFS API only provides the "cumulative size" that includes the additional metadata on how the Merkle DAG is formed.
-    // So here we account for the additional space used by performing a more lenient max size check by 10%,
-    // we will track the exact conent size of content later when we fetch the content if the content passes this size check.
-    const adjustedMaxSize = maxSizeInBytes * 1.1;
-    if (contentMetadata.CumulativeSize > adjustedMaxSize) {
-      console.info(`Cumulative size of ${contentMetadata.CumulativeSize} bytes is greater than the ${adjustedMaxSize} cumulative size limit.`);
+    if (contentMetadata.DataSize > maxSizeInBytes) {
+      console.info(`Size of ${contentMetadata.DataSize} bytes is greater than the ${maxSizeInBytes} data size limit for CID ${hash}.`);
```
thehenrytsai: I think this can cause legitimate files to be wrongly considered as bad. Let's file a bug to optimize this part, I suspect that we will still need to use the same/similar cumulative size check for optimization.

lib/ipfs/IpfsError.ts
isaacJChen: https://github.com/decentralized-identity/sidetree/issues/417

lib/ipfs/IpfsStorage.ts
isaacJChen: Will remove the check completely, and create an optimization task.

lib/ipfs/IpfsStorage.ts
```diff
@@ -151,14 +138,15 @@ export default class IpfsStorage {
    * @returns The multihash content identifier of the stored content.
    */
   public async write (content: Buffer): Promise<string> {
-    const files = await this.node.add(content);
-    return files[0].hash;
+    const file = await this.node.add(content).next();
```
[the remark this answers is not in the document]
isaacJChen: It returns the next element in the response. So if you add a file, the write info (CID, and other metadata) is returned. It is the same as calling index 0 like what we were doing, but in an async iterator now.

lib/ipfs/IpfsStorage.ts
```diff
@@ -83,65 +91,44 @@ export default class IpfsStorage {
...
-        if (currentContentSize > maxSizeInBytes) {
-          console.info(`Content stream reached ${currentContentSize} bytes which is greater than the ${maxSizeInBytes} bytes limit.`);
-          fetchResult.code = FetchResultCode.MaxSizeExceeded;
+    let iterator: AsyncIterator<Buffer>;
+    try {
+      iterator = this.node.cat(hash);
+    } catch (e) {
+      // when an error is thrown, that means the hash points to something that is not a file
+      console.debug(`Error thrown while downloading content from IPFS for CID ${hash}: ${JSON.stringify(e, Object.getOwnPropertyNames(e))}`);
+      return { code: FetchResultCode.NotAFile };
```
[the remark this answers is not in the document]
isaacJChen: These are the errors it can throw so yes. If the CID is a directory, it is not a file, and if it has no content, it also is not a file

`    if (file.unixfs && file.unixfs.type.includes('dir')) {
      throw new Error('this dag node is a directory')
    }

    if (!file.content) {
      throw new Error('this dag node has no content')
    }`

lib/ipfs/IpfsStorage.ts
```diff
@@ -83,65 +91,44 @@ export default class IpfsStorage {
...
-      file.content.on('end', () => {
-        resolveFunction();
-      });
-
-      file.content.resume();
-    });
-
-    await fetchContent;
+      }
+    } while (!result.done);
```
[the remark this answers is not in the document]
isaacJChen: done is always a true or false, the type `IteratorResult` has a field `done`, and it has to be true or false.

lib/ipfs/IpfsStorage.ts
```diff
@@ -83,65 +91,44 @@ export default class IpfsStorage {
...
+      // when an error is thrown, that means the hash points to something that is not a file
+      console.debug(`Error thrown while downloading content from IPFS for CID ${hash}: ${JSON.stringify(e, Object.getOwnPropertyNames(e))}`);
+      return { code: FetchResultCode.NotAFile };
+    }
 
-          readableStream.destroy();
-          return;
+    let result: IteratorResult<Buffer>;
+    do {
+      try {
```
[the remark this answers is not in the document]
isaacJChen: Sure, it can be. Will change.

lib/ipfs/IpfsStorage.ts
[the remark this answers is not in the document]
isaacJChen: add comment

PR #408, opened 2020-02-26
estimated transaction fee

isaacJChen:
Note: This change will make the previous versions (currently just 0.6.1 since we deleted everything else) break. Will require another genesis reset.

@isaacJChen update documentation.


lib/bitcoin/BitcoinClient.ts
```diff
@@ -8,6 +8,7 @@ import nodeFetch, { FetchError, Response, RequestInit } from 'node-fetch';
 import ReadableStream from '../common/ReadableStream';
 import { Address, crypto, Networks, PrivateKey, Script, Transaction, Unit } from 'bitcore-lib';
 import { IBlockInfo } from './BitcoinProcessor';
+import BitcoinDataTransactionModel from './models/BitcoinDataTransactionModel';
```
mudiali: Sort please.

lib/bitcoin/BitcoinClient.ts
```diff
@@ -28,7 +29,8 @@ export default class BitcoinClient {
     bitcoinRpcPassword: string | undefined,
     bitcoinWalletImportString: string,
     private requestTimeout: number,
-    private requestMaxRetries: number) {
+    private requestMaxRetries: number,
+    private transactionFeeMarkupPercentage: number) {
```
mudiali: One concern that I have is that this percentage is only applied to the 'sidetree' transactions. Any other transactions such as lock/relock etc don't honor this percentage. So we need to decide between:
1. Either every transaction has this markup.
1. Or rename this variable to denote that this markup is ONLY for the sidetree transactions.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -214,40 +215,38 @@ export default class BitcoinProcessor {
...
+   * @param minFee The minimum fee to be paid for this transaction.
    */
-  public async writeTransaction (anchorString: string, fee: number) {
-    console.info(`Fee: ${fee}. Anchoring string ${anchorString}`);
-    // ----
-    // Issue #347 opened to track the investigation for this hardcoded value.
-    // Ensure that we are always paying this minimum fee.
-    fee = Math.max(fee, 1000);
-    // ----
+  public async writeTransaction (anchorString: string, minFee: number) {
```
mudiali: This changes the contract between core and bitcoin API and needs documentation updates. https://github.com/decentralized-identity/sidetree/blob/master/docs/implementation.md#write-a-sidetree-transaction

lib/bitcoin/IBitcoinConfig.ts
```diff
@@ -16,4 +16,5 @@ export interface IBitcoinConfig {
   requestTimeoutInMilliseconds: number | undefined;
   requestMaxRetries: number | undefined;
   transactionPollPeriodInSeconds: number | undefined;
+  transactionFeeMarkupPercentage: number | undefined;
```
mudiali: 1. We need a corresponding change in the ion repo to include this new parameter.
1. Can we keep this variable as required? And by default set it to 0?

lib/bitcoin/BitcoinClient.ts
isaacJChen: Will rename to only apply to sidetree transactions as discusses yesterday.

docs/implementation.md
```diff
@@ -402,7 +402,7 @@ POST /transactions
 #### Request body schema
 ```json
 {
-  "fee": "A number representing the transaction fee to be paid to write this transaction to the blockchain.",
+  "minFee": "A number representing the minimum transaction fee to be paid to write this transaction to the blockchain. The actual fee paid can fluctuate deterministically.",
```
mudiali: How about ````minimumFree````

mudiali: Just a reminder that this requires a corresponding change in ion repo as well.

mudiali:
```suggestion
  "minFee": "A number representing the minimum transaction fee to be paid to write this transaction to the blockchain. The actual fee is dynamically calculated and can be higher than this amount (but not lower).",
```

isaacJChen: Good call. I created an issue so we remember what we've done when we update ion

https://github.com/decentralized-identity/ion/issues/56

isaacJChen: Is it free though? JK, I like it, will change.

lib/bitcoin/BitcoinClient.ts
```diff
@@ -1,5 +1,6 @@
 import * as httpStatus from 'http-status';
 import BitcoinBlockModel from './models/BitcoinBlockModel';
+import BitcoinDataTransactionModel from './models/BitcoinDataTransactionModel';
```
thehenrytsai: Makes more sense to me if it was `BitcoinTransactionDataModel`.

lib/bitcoin/IBitcoinConfig.ts
```diff
@@ -16,4 +16,5 @@ export interface IBitcoinConfig {
   requestTimeoutInMilliseconds: number | undefined;
   requestMaxRetries: number | undefined;
   transactionPollPeriodInSeconds: number | undefined;
+  sidetreeTransactionFeeMarkupPercentage: number;
```
thehenrytsai: Consider renaming to just `transactionFeeMarkupPercentage`.

lib/core/models/Config.ts
```diff
@@ -4,7 +4,7 @@
 export default interface Config {
   batchingIntervalInSeconds: number;
   blockchainServiceUri: string;
-  transactionFeeMarkupPercentage: number;
+  sidetreeTransactionFeeMarkupPercentage: number;
```
thehenrytsai: Sanity: not needed?

lib/core/versions/latest/FeeManager.ts
```diff
@@ -17,7 +17,7 @@ export default class FeeManager {
    *
    * @throws if the number of operations are <= 0.
    */
-  public static computeTransactionFee (normalizedFee: number, numberOfOperations: number, feeMarkupPercentage: number): number {
+  public static computeTransactionFee (normalizedFee: number, numberOfOperations: number): number {
```
thehenrytsai: Consider:
```suggestion
  public static computeMinimalTransactionFee (normalizedFee: number, numberOfOperations: number): number {

```

tests/json/config-test.json
```diff
@@ -6,6 +6,5 @@
   "batchingIntervalInSeconds": 10,
   "observingIntervalInSeconds": 10,
   "maxConcurrentDownloads": 20,
-  "mongoDbConnectionString": "mongodb://localhost:27017/",
-  "transactionFeeMarkupPercentage": 10
```
thehenrytsai: Sanity: don't forget to remove this from ION (and internal ION) repo also.

isaacJChen: Yes, I have an issue for that

lib/core/models/Config.ts
isaacJChen: yes

PR #409, opened 2020-02-27
#271 Add lock-monitor to the bitcoin processor

mudiali:
1. Add support to get the current valueTimeLock to BitcoinProcessor.
1. Add support to search for any valueTimeLock by identifier to BitcoinProcessor.
1. Add lock monitor to the BitcoinProcessor.
1. Update lock monitor to handle if the current lock is not yet written to the bitcoin yet.

The REST API support is not there yet.


lib/bitcoin/lock/MongoDbLockTransactionStore.ts
```diff
@@ -16,7 +18,8 @@ export default class MongoDbLockTransactionStore {
    * @param serverUrl The target server url.
    * @param databaseName The database name where the collection should be saved.
    */
-  public constructor (private serverUrl: string, private databaseName: string) {
+  public constructor (private serverUrl: string, private databaseName?: string) {
```
isaacJChen: Since the defaultDatabaseName is not used anywhere else, couldn't the constructor just be

`public constructor (private serverUrl: string, private databaseName: string = 'sidetree')`?

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -110,6 +122,22 @@ export default class BitcoinProcessor {
...
+
+    const valueTimeLockTransactionFeesInBtc = config.valueTimeLockTransactionFeesAmountInBitcoins || 0.25;
+    const numberOfBlocksInOneMonth = BitcoinClient.estimatedNumberOfBlocksInOneHour * 24 * 30;
+
+    this.lockMonitor =
+      new LockMonitor(
+        this.bitcoinClient,
+        this.mongoDbLockTransactionStore,
+        config.valueTimeLockPollPeriodInSeconds || 10 * 60,
+        config.valueTimeLockAmountInBitcoins * BitcoinProcessor.satoshiPerBitcoin, // Desired lock amount in satoshis
```
isaacJChen: We should make sure this number comes out to be a whole number

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -110,6 +122,22 @@ export default class BitcoinProcessor {
         config.bitcoinWalletImportString,
         config.requestTimeoutInMilliseconds || 300,
         config.requestMaxRetries || 3);
+
+    this.lockResolver = new LockResolver(this.bitcoinClient);
+
+    this.mongoDbLockTransactionStore = new MongoDbLockTransactionStore(config.mongoDbConnectionString, config.databaseName);
+
+    const valueTimeLockTransactionFeesInBtc = config.valueTimeLockTransactionFeesAmountInBitcoins || 0.25;
```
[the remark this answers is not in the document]
mudiali: Great catch. will fix.

lib/bitcoin/BitcoinProcessor.ts
mudiali: Will add checks inside the LockMonitor constructor.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -110,6 +122,22 @@ export default class BitcoinProcessor {
...
+    const numberOfBlocksInOneMonth = BitcoinClient.estimatedNumberOfBlocksInOneHour * 24 * 30;
+
+    this.lockMonitor =
+      new LockMonitor(
+        this.bitcoinClient,
+        this.mongoDbLockTransactionStore,
+        config.valueTimeLockPollPeriodInSeconds || 10 * 60,
+        config.valueTimeLockAmountInBitcoins * BitcoinProcessor.satoshiPerBitcoin, // Desired lock amount in satoshis
+        numberOfBlocksInOneMonth,                                                  // Desired lock duration in blocks
+        valueTimeLockTransactionFeesInBtc * BitcoinProcessor.satoshiPerBitcoin);   // Txn Fees amoount in satoshis
```
thehenrytsai: Any reason why we cannot make internal code and config both use satoshis? Also, would be nice if the lockTransactionsFee is right after lockAmount.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -283,6 +313,47 @@ export default class BitcoinProcessor {
     return this.serviceInfoProvider.getServiceVersion();
   }
 
+  /**
+   * Gets the lock information for the specified identifier (if specified); if nothing is passed in then
+   * it returns the current lock information (if one exist).
+   *
+   * @param lockIdentifier The identifier of the lock to look up; should be null for the current node's lock information.
+   */
+  public async getValueTimeLock (lockIdentifier: string): Promise<ValueTimeLockModel> {
```
thehenrytsai: Following [TypeScript guidelines](https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines#null-and-undefined), We should avoid `null` in TypeScript whenever, so remove the mentioning of `null` in line 320; does this mean `lockIdnetifier` should be `string | undefined`?

lib/bitcoin/lock/LockMonitor.ts
```diff
@@ -51,8 +58,30 @@ export default class LockMonitor {
...
+   * if the lock information is not confirmed on the blockchain.
+   */
+  public getCurrentValueTimeLock (): ValueTimeLockModel | undefined {
+
+    // Make a copy of the state so in case it gets changed between now and the function return
+    const currentLockState = Object.assign({}, this.currentLockState!);
+
+    // If there's no lock then return undefined
+    if (currentLockState.status === LockStatus.None) {
+      return undefined;
```
thehenrytsai: Would it make sense to throw an error here instead of undefined like line 78?

lib/bitcoin/lock/LockMonitor.ts
```diff
@@ -108,7 +144,8 @@ export default class LockMonitor {
 
     const currentLockState: LockState = {
       currentValueTimeLock: undefined,
-      latestSavedLockInfo: undefined
+      latestSavedLockInfo: undefined,
+      status: LockStatus.None
```
thehenrytsai: Consider initializing `this.currentLockState` with this so `currentLockState` and just be type `LockState` so you don't have to use the `!` operater in line 104 and 114.

lib/bitcoin/lock/LockMonitor.ts
```diff
@@ -127,6 +172,8 @@ export default class LockMonitor {
     if (!(await this.isTransactionWrittenOnBitcoin(lastSavedLock.transactionId))) {
 
       await this.rebroadcastTransaction(lastSavedLock);
+      currentLockState.status = LockStatus.Pending;
+
       return currentLockState;
```
thehenrytsai: Consider constructing and return the full state object within these short-circuit return blocks for readability.

lib/bitcoin/lock/LockMonitor.ts
```diff
@@ -143,6 +190,7 @@ export default class LockMonitor {
     };
 
     currentLockState.currentValueTimeLock = await this.lockResolver.resolveLockIdentifierAndThrowOnError(lastLockIdentifier);
+    currentLockState.status = LockStatus.Confirmed;
```
thehenrytsai: Maybe I misunderstood, the comment above about "which we still need to resolve." is confusing to me, I was under the impression that we don't need to resolve, but we are doing it out of convenience?

lib/bitcoin/BitcoinProcessor.ts
mudiali: Refactoring it to 2 different methods.

lib/bitcoin/lock/LockMonitor.ts
mudiali: You are right ... throwing it upon refactor.

lib/bitcoin/lock/LockMonitor.ts
mudiali: Updated.

lib/bitcoin/lock/LockMonitor.ts
mudiali: Good idea .. done.

lib/bitcoin/lock/LockMonitor.ts
mudiali: Updating the comment.

lib/bitcoin/BitcoinProcessor.ts
mudiali: I think it's easier for the admin to think in BTC instead of figuring out the BTC and then multiplying w/ the right value to convert to satoshis etc.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -283,6 +315,47 @@ export default class BitcoinProcessor {
     return this.serviceInfoProvider.getServiceVersion();
   }
 
+  /**
+   * Gets the lock information for the specified identifier (if specified); if nothing is passed in then
+   * it returns the current lock information (if one exist).
+   *
+   * @param lockIdentifier The identifier of the lock to look up; should be null for the current node's lock information.
```
```suggestion
   * @param lockIdentifier The identifier of the lock to look up.
```

Issue #266, opened 2019-07-17
Allow clearing of dead ops to avoid negatively impacting resolution time, storage/bandwidth, and new node boot time


csuwildcat:

csuwildcat:
Along-time goal of the protocol has been to devise a means for large scale pruning of historical data to reduce the total amount of data a full node stores. The following proposal aims to achieve that goal via a set of changes to existing data structures and protocol rules. The changes are likely to result in a significant reduction of past op data from the network.

The first step in in this effort is retooling the Anchor/Batch file approach to include another level: a Map File that captures everything necessary to securely, accurately derive the lineage of an ID far into the future, without retention of verbose op data from batch files.

### New three-tiered file structure:

The protocol will be modified to separate ops by type within the Anchor file, linking to their own, distinct Map and Batch files. The number of files will be increased from the current two (Anchor and Batch files) to the following six:

1. Anchor File
2. Update op Map File
3. Recovery/Checkpoint op Map File
4. Create op Batch file
5. Update op Batch file
6. Recovery/Checkpoint op Batch file

#### Create Operations

**Anchor**

Single values:
- Hash of Create op Batch File: 32 bytes

Per DID:
- Initial DID Doc hash: 32 bytes
- Recovery commit hash: 32 bytes
- Recovery key: 33 bytes

**Map**
_NO MAP FILE_

**Batch**

Per DID:
- 1 Kilobyte of op data - _must include update commit hash_

#### Update Operations

**Anchor**

Single values:
- Map File hash: 32 bytes

**Map**
Single values:
- Hash of Update op Batch File: 32 bytes

Per DID:
- Operation hash: 32 bytes
- Update reveal value: 32 bytes
- Signature over operation hash commit hash: 71 bytes

**Batch**
 - 1 Kilobyte of op data - _must include next Update commit hash_

#### Recovery/Checkpoint Operations

**Anchor**

Single values:
- Hash of Recovery/Checkpoint op Map file: 32 bytes

Per DID:
- DID Unique Suffix: 32 bytes
- Reveal Value: 32 bytes

**Map**

Single values:
- Hash of Recovery/Checkpoint op Batch file: 32 bytes

Per DID:
- Operation hash: 32 bytes
- Commit hash: 32 bytes
- Recovery key (optional): 33 bytes
- Signature over DDO hash, commit hash, and optional recovery key: 71 bytes

**Batch**
- 1 Kilobyte of op data - _must include next Update commit hash_

[...]

PR #411, opened 2020-02-27
#266 - Updated update operation request schema

thehenrytsai:


lib/core/versions/latest/UpdateOperation.ts
```diff
@@ -0,0 +1,122 @@
...
+    this.operationData = operationData;
+  }
+
+  /**
+   * Parses the given buffer as a `UpdateOperation`.
+   */
+  public static async parse (operationBuffer: Buffer): Promise<UpdateOperation> {
+    const operationJsonString = operationBuffer.toString();
+    const operationObject = await JsonAsync.parse(operationJsonString);
+    const createOperation = await UpdateOperation.parseObject(operationObject, operationBuffer);
```
isaacJChen: updateOperation = ...

lib/core/versions/latest/UpdateOperation.ts
```diff
@@ -0,0 +1,122 @@
...
+    return createOperation;
+  }
+
+  /**
+   * Parses the given operation object as a `CreateOperation`.
```
isaacJChen: as UpdateOperation

lib/core/versions/latest/util/Jws.ts
```diff
@@ -22,14 +78,55 @@ export default class Jws {
...
 
+  /**
+   * Signs the given protected header and payload as a JWS.
+   *
+   * @param payload If the given payload is of string type, it is assumed to be encoded string;
+   *                else the object will be stringified and encoded.
+   */
+  public static async sign (
+    protectedHeader: any,
+    payload: any | string,
```
isaacJChen: Any covers string, unless you think it's better to be explicit.

tests/core/util/Jws.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+      const [, signingPrivateKey] = await Cryptography.generateKeyPairHex(signingKeyId, KeyUsage.signing);
+
+      const protectedHeader = {
+        kid: signingKeyId,
+        alg: true // Incorect type.
+      };
+
+      const payload = { anyProperty: 'anyValue' };
+
+      const jws = await Jws.sign(protectedHeader, payload, signingPrivateKey);
```
isaacJChen: We should probably have a test for when payload is a string if we want to support it.

lib/core/versions/latest/DocumnetComposer.ts
```diff
@@ -0,0 +1,34 @@
+import UpdateOperation from './UpdateOperation';
```
mudiali: Typo in the filename.

lib/core/versions/latest/OperationProcessor.ts
```diff
@@ -112,30 +121,25 @@ export default class OperationProcessor implements IOperationProcessor {
...
 
-    // The operation passes all checks, apply the patches.
-    AnchoredOperation.applyPatchesToDidDocument(didDocument, operation.patches!);
+    const resultingDocument = await this.documentComposer.applyUpdateOperation(operation, didResolutionModel.didDocument);
 
-    didResolutionModel.metadata!.lastOperationTransactionNumber = operation.transactionNumber;
-    didResolutionModel.metadata!.nextUpdateOtpHash = operation.nextUpdateOtpHash!;
+    didResolutionModel.didDocument = resultingDocument;
+    didResolutionModel.metadata!.lastOperationTransactionNumber = namedAnchoredOperationModel.transactionNumber;
+    didResolutionModel.metadata!.nextUpdateOtpHash = operation.operationData.nextUpdateOtpHash;
```
mudiali: I'm a bit confused as to how are you able to access the `operationData.nextUpdateOtpHash` without having to import it? Also, having the same name for the internal class (in both CreateOperation.ts and UpdateOperation.ts) was going to clash right?

lib/core/versions/latest/UpdateOperation.ts
thehenrytsai: Fixed.

lib/core/versions/latest/util/Jws.ts
thehenrytsai: Removed.

lib/core/versions/latest/UpdateOperation.ts
thehenrytsai: Fixed!

lib/core/versions/latest/DocumnetComposer.ts
thehenrytsai: Thanks. Fixed.

tests/core/util/Jws.spec.ts
thehenrytsai: Added a note saying that this is mainly created for tests to create valid data, and the path is hit many times already.

lib/core/versions/latest/OperationProcessor.ts
thehenrytsai: Short answer is that this is something that TypeScript handles gracefully.

PR #422, opened 2020-03-04
#271 Add lock start time to the ValueTimeLockModel

mudiali:
Had to do some integration testing while trying to figure this out and ended up fixing a couple of bugs as well.
1. We were not checking the `confirmations` property on the transaction to see whether the transaction is confirmed or not. Now we are.
1. The unfreeze/spend transaction cannot be broadcasted until the unlock block is reached. Before we were trying to broadcast 1 block before the unlock block.
1. The release lock logic should also check whether we have reached the unlock block or not.


lib/bitcoin/lock/LockMonitor.ts
```diff
@@ -277,12 +296,8 @@ export default class LockMonitor {
     latestSavedLockInfo: SavedLockModel,
     desiredLockAmountInSatoshis: number): Promise<boolean> {
 
-    const currentBlockTime = await this.bitcoinClient.getCurrentBlockHeight();
-
-    console.info(`Current block: ${currentBlockTime}; Current lock's unlock block: ${currentValueTimeLock.unlockTransactionTime}`);
-
     // Just return if we're not close to expiry
```
isaacJChen: The old logic was testing if we are close, now we are testing if we are exactly at release time right?

lib/bitcoin/lock/LockMonitor.ts
```diff
@@ -277,12 +296,8 @@ export default class LockMonitor {
     latestSavedLockInfo: SavedLockModel,
     desiredLockAmountInSatoshis: number): Promise<boolean> {
 
-    const currentBlockTime = await this.bitcoinClient.getCurrentBlockHeight();
-
-    console.info(`Current block: ${currentBlockTime}; Current lock's unlock block: ${currentValueTimeLock.unlockTransactionTime}`);
-
     // Just return if we're not close to expiry
-    if (currentValueTimeLock.unlockTransactionTime - currentBlockTime > 1) {
+    if (! (await this.isUnlockTimeReached(currentValueTimeLock.unlockTransactionTime))) {
```
isaacJChen: I feel like we discussed this already, but I can't remember what the conclusion was. Between the time a new lock is broadcasted and when it is actually on the chain, during that time, the node will lose its ability to write large volume right?

lib/bitcoin/lock/LockMonitor.ts
mudiali: Good point. Will change.

lib/bitcoin/lock/LockMonitor.ts
mudiali: Yes you are right. There may be a block (or two) when this node will not have any lock.

lib/bitcoin/lock/LockMonitor.ts
```diff
@@ -375,4 +408,12 @@ export default class LockMonitor {
 
     return lockInfoToSave;
   }
+
+  private async isUnlockTimeReached (unlockTransactionTime: number): Promise<boolean> {
+    const currentBlockTime = await this.bitcoinClient.getCurrentBlockHeight();
+
+    console.info(`Current block: ${currentBlockTime}; Current lock's unlock block: ${unlockTransactionTime}`);
+
+    return currentBlockTime - unlockTransactionTime >= 0;
```
thehenrytsai: Nit:
```suggestion
    return currentBlockTime >= unlockTransactionTime;

```

lib/bitcoin/lock/LockMonitor.ts
```diff
@@ -277,12 +296,8 @@ export default class LockMonitor {
     latestSavedLockInfo: SavedLockModel,
     desiredLockAmountInSatoshis: number): Promise<boolean> {
```
thehenrytsai: Would be nice to know the returned `boolean` means here without looking at the caller pattern.

lib/bitcoin/lock/LockMonitor.ts
mudiali: THANK you  ... I was completely blanking out for some reason and made it so much harder than it has to be :) I actually spent like 10/15 mins on this statement alone.

lib/bitcoin/lock/LockMonitor.ts
```diff
@@ -316,6 +331,24 @@ export default class LockMonitor {
     return true;
   }
 
+  /**
+   * Performs the release lock routine. Returns true if there were changes to the lock; false otherwise.
```
thehenrytsai: Super nit: consider using `@returns` for describing return values. I was looking for it when reading code, didn't realized it was actually described it in summary.

lib/bitcoin/lock/LockResolver.ts
```diff
@@ -151,4 +153,14 @@ export default class LockResolver {
       throw BitcoinError.createFromError(ErrorCode.LockResolverTransactionNotFound, e);
     }
   }
+
+  private async calculateLockStartingBlock (transaction: BitcoinTransactionModel): Promise<number> {
+    if (transaction.confirmations === 0 || transaction.confirmations < 0) {
```
thehenrytsai: Sanity: Maybe I am misreading this, can confirmations be less than zero?
```suggestion
    if (transaction.confirmations <= 0) {

```

mudiali: They cannot be but just for sanity doing the `<` check.

During integration, I saw that the `confirmations` was undefined for the broadcasted (and not yet written) transactions. So I had the `===` check. I can change it to the `<=` because I did update the `bitcoinclient` to make sure that `confirmations` is never undefined.

lib/bitcoin/lock/LockMonitor.ts
```diff
@@ -192,22 +196,37 @@ export default class LockMonitor {
...
+      };
+
+    } catch (e) {
+
+      if (e instanceof BitcoinError && e.code === ErrorCode.LockResolverTransactionNotConfirmed) {
+        // This means that the transaction was broadcasted but hasn't been written on the blockchain yet.
+        return {
+          currentValueTimeLock: undefined,
+          latestSavedLockInfo: lastSavedLock,
+          status: LockStatus.Pending
```
thehenrytsai: No action required, more of a discussion: `currentValueTimeLock` being `undefined` when `LockStatus` is `Pending` is throwing me off, I see the logic being the lock is not "current" unless it is "confirmed and not expired", in which case we can also consider naming it `activeValueTimeLock`.

lib/bitcoin/lock/LockMonitor.ts
mudiali: Sure will do.

lib/bitcoin/lock/LockMonitor.ts
mudiali: I was thinking on the same lines actually. Let me change the name and add some comments around the field to be more clear about the usage.

Issue #425, opened 2020-03-04
Make ipfs request handler follow constructor/initialize pattern


isaacJChen:

tplooker:
Outdated, we are going to move go implementation for IPFS node

PR #427, opened 2020-03-06
#425 - Make ipfs request handler follow constructor/initialize pattern

sambarnes:
In response to #425 

New to js/ts so please correct me if I misused the definite assignment assertion operator on the `RequestHandler.ipfsStorage` property. Wasn't sure if `!` should be used or if it should have just been declared as `ipfsStorage: IpfsStorage | undefined` to force the caller to make sure `intialize()` was invoked.

### DIFF
--- a/lib/ipfs/ErrorCode.ts
+++ b/lib/ipfs/ErrorCode.ts
@@ -1,4 +1,5 @@
 export default {
   IpfsStorageInstanceCanOnlyBeCreatedOnce: 'ipfs_storage_instance_can_only_be_created_once',
-  IpfsStorageInstanceGetHasToBeCalledAfterCreate: 'ipfs_storage_instance_get_has_to_be_called_after_create'
+  IpfsStorageInstanceGetHasToBeCalledAfterCreate: 'ipfs_storage_instance_get_has_to_be_called_after_create',
+  RequestHandlerIpfsStorageInstanceGetHasToBeCalledAfterInitialize: 'request_handler_ipfs_storage_instance_get_has_to_be_called_after_initialize'
 };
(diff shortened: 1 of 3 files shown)


JaceHensley:
I think it'd be better to define it as `ipfsStorage: IpfsStorage | undefined` and then check that `ipfsStorage` is set before using it. And then throw an error saying "You must call initialize first" or something like that

sambarnes:
@JaceHensley  I agree that is the safer solution. I have changed the behavior to throw when trying to call functions that use `ipfsStorage` on an uninitialized `RequestHandler`.

The only thing is that now the tests are failing to build unless I suppress the following:
```
tests/ipfs/RequestHandler.spec.ts:57:39 - error TS2345: Argument of type '"read"' is not assignable to parameter of type 'never'.
57     spyOn(requestHandler.ipfsStorage, 'read').and.returnValue(Promise.resolve({ code: FetchResultCode.Success, content: Buffer.from('dummyContent') }));

tests/ipfs/RequestHandler.spec.ts:71:39 - error TS2345: Argument of type '"write"' is not assignable to parameter of type 'never'.
71     spyOn(requestHandler.ipfsStorage, 'write').and.returnValue(Promise.resolve('QmYtUc4iTCbbfVSDNKvtQqrfyezPPnFvE33wFmutw9PBBk'));
```

I am unfamiliar with the usage of `spyOn` so the reason for failing build is not immediately apparent to me. I'll see what I can come up with though

JaceHensley:
Hmm, that's probably because it could be undefined. What if instead of a field ipfsStorage was a setter/getter and the getter could be guarded.

```ts
private _ipfsStorage: IpfsStorage | undefined

get ipfsStorage(): IpfsStorage {
  if (!this._ipfsStorage) throw new Error()

  return this._ipfsStorage
}

set ipfsStorage(value: IpfsStorage) {
  this._ipfsStorage = value
}
```

OR13:
any reason not to merge this?

OR13:
@sambarnes @JaceHensley @thehenrytsai This PR has been open for a really long time. I'm going to close it in 24 hours without comments / approvals / feedback.

PR #428, opened 2020-03-06
#271 Parse and return the sidetree transaction writer (public key hash) 

mudiali:
1. While reading the sidetree transaction, also read the writer.
1. Return the writer info for each transaction.

### DIFF
--- a/lib/bitcoin/BitcoinProcessor.ts
+++ b/lib/bitcoin/BitcoinProcessor.ts
@@ -708,7 +708,8 @@ export default class BitcoinProcessor {
         transactionTimeHash: transaction.blockHash,
         anchorString: sidetreeData.data,
         transactionFeePaid: transactionFeePaid,
-        normalizedTransactionFee: normalizedFeeModel.normalizedTransactionFee
+        normalizedTransactionFee: normalizedFeeModel.normalizedTransactionFee,
+        writer: sidetreeData.writer
       };
     }
 
--- a/lib/bitcoin/BitcoinProcessor.ts
+++ b/lib/bitcoin/BitcoinProcessor.ts
@@ -1,7 +1,6 @@
 import BitcoinBlockModel from './models/BitcoinBlockModel';
 import BitcoinClient from './BitcoinClient';
 import BitcoinError from './BitcoinError';
-import BitcoinOutputModel from './models/BitcoinOutputModel';
 import BitcoinTransactionModel from './models/BitcoinTransactionModel';
 import ErrorCode from './ErrorCode';
 import LockMonitor from './lock/LockMonitor';
@@ -15,6 +14,7 @@ import ReservoirSampler from './fee/ReservoirSampler';
 import ServiceInfoProvider from '../common/ServiceInfoProvider';
 import ServiceVersionModel from '../common/models/ServiceVersionModel';
 import SharedErrorCode from '../common/SharedErrorCode';
+import SidetreeTransactionData from './SidetreeTransactionData';
 import SlidingWindowQuantileCalculator from './fee/SlidingWindowQuantileCalculator';
 import SpendingMonitor from './SpendingMonitor';
 import TransactionFeeModel from '../common/models/TransactionFeeModel';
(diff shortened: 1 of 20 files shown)


lib/bitcoin/SidetreeTransactionData.ts
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  private static getSidetreeDataFromVOutIfExist (transactionOutput: BitcoinOutputModel, sidetreePrefix: string): string | undefined {
+
+    // check for returned data for sidetree prefix
+    const hexDataMatches = transactionOutput.scriptAsmAsString.match(/\s*OP_RETURN ([0-9a-fA-F]+)$/);
+
+    if (hexDataMatches && hexDataMatches.length !== 0) {
+
+      const data = Buffer.from(hexDataMatches[1], 'hex').toString();
```
thehenrytsai: Sanity: would the following not work?
```suggestion
      const data = hexDataMatches[1];

```

lib/bitcoin/SidetreeTransactionData.ts
```diff
@@ -0,0 +1,88 @@
...
+      return undefined;
+    }
+
+    const sidetreeData = this.getSidetreeDataFromVOutIfExist(bitcoinTransaction.outputs[0], sidetreePrefix);
+
+    if (!sidetreeData) {
+      return undefined;
+    }
+
+    const writer = this.getWriterFromVOutIfExist(bitcoinTransaction.outputs[1]);
```
thehenrytsai: Sanity: doesn't this mean anyone can write large batch by sending a small amount of money to the person who created the value timelock?

mudiali: The person who received the money can write. The sender can't.

mudiali: Updating the algorithm to only look at the inputs to determine the writer. We now check that all the inputs are spending from the same public key (wallet) and then that public key becomes the writer of the transaction. Good catch @thehenrytsai.

lib/bitcoin/SidetreeTransactionData.ts
```diff
@@ -0,0 +1,131 @@
...
+   *
+   * @param bitcoinTransaction The transaction to parse.
+   * @param sidetreePrefix The prefix of the sidetree transactions.
+   *
+   * @returns This object if the transaction is a valid sidetree transaction; undefined otherwise.
+   */
+  public static parse (bitcoinTransaction: BitcoinTransactionModel, sidetreePrefix: string): SidetreeTransactionData | undefined {
+
+    // The sidetree transaction has the following requirements:
+    //  1. There should be only one output with a valid sidetree anchorstring
```
thehenrytsai: Nit: 
```suggestion
    //  1. There must be only one output with a valid sidetree anchorstring

```

lib/bitcoin/SidetreeTransactionData.ts
```diff
@@ -0,0 +1,131 @@
...
+    });
+
+    // Save all the unique public keys.
+    const uniquePublicKeys = new Set<string | undefined>();
+    for (let i = 0; i < transactionInputs.length; i++) {
+      uniquePublicKeys.add(allPublicKeys[i]);
+    }
+
+    // There should be only 1 key in all of the inputs; so if we have more than 1 then that is invalid.
+    if (uniquePublicKeys.size !== 1) {
```
thehenrytsai: Sanity: it seems possible to have `undefined` as the only unique key.

lib/bitcoin/SidetreeTransactionData.ts
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  private static getValidPublicKeyFromInputs (transactionId: string, transactionInputs: BitcoinInputModel[]): string | undefined {
+
+    // First get all the public keys from the inputs
+    const allPublicKeys = transactionInputs.map(input => {
+      const scriptAsmParts = input.scriptAsmAsString.split(' ');
+
+      // If the publickey is not present then just use 'undefined'
+      return scriptAsmParts.length >= 2 ? scriptAsmParts[1] : undefined;
```
thehenrytsai: For assurance, are you able to add a comment explaining why [1] is always the public key when length is great than 1?

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -653,7 +614,8 @@ export default class BitcoinProcessor {
     for (let transactionIndex = 1; transactionIndex < transactions.length; transactionIndex++) {
       const transaction = transactions[transactionIndex];
 
-      const isSidetreeTransaction = this.isSidetreeTransaction(transaction);
+      const sidetreeData = SidetreeTransactionData.parse(transaction, this.sidetreePrefix);
```
thehenrytsai: This call appears to be quite expensive but we call it twice for every transaction. I wonder if it is worth avoiding this. It is fine to file an issue if you think it is worthwhile but not mainnet blocking.

lib/bitcoin/SidetreeTransactionData.ts
mudiali: Yes it is possible. `undefined` is one of the expected outputs and the caller will interpret undefined === 'no public key was found' === not a valid sidetree txn.

lib/bitcoin/SidetreeTransactionData.ts
mudiali: Added a comment to figure out this assumption in the next PR.

lib/bitcoin/BitcoinProcessor.ts
mudiali: Yea that is a bigger refactor issue and I have opened issue #434 to track this.

PR #429, opened 2020-03-06
#271 Add get-value-time-lock support to the IBlockchain

mudiali:


docs/implementation.md
```diff
@@ -480,6 +480,115 @@ HTTP/1.1 500 Internal Server Error
...
+  "code": "value_time_lock_not_found"
+}
+```
+
+### Fetch the writer lock object used for batch writing.
+Fetches the currently active writer lock object written on the blockchain by the Blockchain service. This is used for batch writing.
+
+Returns `HTTP 404 Not Found` with `value_time_lock_not_found` as the `code` parameter value in the JSON body if there is no active lock on the blockchain.
+Returns `HTTP 404 Not Found` with `value_time_lock_in_pending_state` as the `code` parameter value in the JSON body if there is a lock but is not confirmed on the blockchain yet.
+Returns `HTTP 500 Internal Server Error` if there is an error while trying to fetch the lock.
```
thehenrytsai: Sanity: since this is supposed to be unexpected, maybe no need to specify?

mudiali: Yes, this is unexpected. I just checked and we aren't documenting 'internal server' errors. I'll take this one out.

PR #431, opened 2020-03-07
#266 - Updated revoke operation schema

thehenrytsai:
1. Renamed delete operation to revoke operation.
2. Updated revoke operation schema.
3. Minor refactoring + removed obsolete code.


lib/core/Resolver.ts
```diff
@@ -46,6 +46,11 @@ export default class Resolver {
       return undefined;
     }
 
+    // If last operation is a revoke. No need to continue, report DID status as 'revoked'.
+    if (didResolutionModel.metadata!.nextRecoveryOtpHash === undefined) {
```
mudiali: Is this the only way to check the 'revoke' status? I guess we don't have the `lastOperationType` or similar to really know what the last op was do we?

thehenrytsai: Yeah... this is the only way currently. `DidResolutionModel` will be removed entirely in my next PR and be replaced by the new internal document state model, I can add last operation type to it since it will make the code easier to read, my only reservation is that we might not want to add a one-time-use property to a model.

Issue #383, opened 2020-01-28
As a did controller, I should be able to update any property of my did document.


OR13:
As a sidetree method implementer, I should not have to create a new schema for each new property type.

csuwildcat:
Thank you for capturing all the floating bugs that allude to this more general desire. First, I want to ask if you are comfortable with slightly tweaking the title to center on the protocol change itself and the implication. Given just about anything is technically allowed in a DID Document via random inclusion or context extension, I feel this Issue title should read: "Allow arbitrary properties and values to be written in DID operations" - this captures the protocol/outcome reality, vs merely one of the high-level user story outcomes that it implies. 

Here is my proposal for making this possible:

1. We maintain protocol-operative keys with a set of strict protocol handling rules and op action types. This means that we would not hinge our handling of protocol-sensitive keys on whatever flavor of the day a DID Doc assumes, which is a proper separation/abstraction from the presentation format.
2. We add a custom property that allows the implementer to open up their implementation to an arbitrary data value, which they can subsequently parse and handle however they please in the reducer portion of the protocol/code.
3. We create a process for codifying new op action types in the specification, for any new op action types we may want to add to the specification in the future.

This creates a flexible mechanism that allows an implementer to basically make their implementation a free-for-all if they choose, while allowing other implementers a mechanism to make additions in more statically defined ways.

OR13:
I think my title is more descriptive of the actual impact to method implementers, and of the deviance from the did core spec, that sidetree has taken by restricting controller's ability to perform meaningful updates to THEIR did document.

I agree with your proposed solution, let me rephrase it:

1. Sidetree protocol will maintain a list of allowed did operations (just like the did core spec).
2. Specific method implementations can choose to use a method specific operation type to enable users to control all properties of their did document, for methods that want to allow that (this is equivalent to defining what "update" means in different methods).
3. If you want to add new operation types to sidetree you will need to get the sidetree authors to approve your operation type... pretty much exactly they same as adding a property to an `@context` from an interop perspective.

I think this is a tractable solution, I do worry that we are perhapse trying to do the same things as the did core spec, by adding lots of definitions to the sidetree protocol... that won't lead to interoperability amongs anything other than sidetree based did methods.

csuwildcat:
Let me provide some further details to clarify some fundamental assumptions/reasoning:

[...]

PR #432, opened 2020-03-11
#266, #383 - Replaced DidResolutionModel with generic DocumentState

thehenrytsai:
1. Replaced `DidResolutionModel` with `DocumentState` so the state keeping is no longer DID document specific.
1. Removed various classes such as `AnchoredOperation` and `ApplyResult` + various renames for consistency.


lib/core/versions/latest/RequestHandler.ts
```diff
@@ -74,7 +75,17 @@ export default class RequestHandler implements IRequestHandler {
       switch (operation.type) {
         case OperationType.Create:
 
-          const document = await this.applyCreateOperation(operation);
+          const documentState = await this.applyCreateOperation(operation);
```
mudiali: Maybe we can do the transformation in the `handleResolveRequest` function. Seems like that logic is common to both type of resolutions: longForm and did.

lib/core/versions/latest/DocumentComposer.ts
```diff
@@ -166,4 +171,123 @@ export default class DocumentComposer {
...
+      DocumentComposer.applyPatchToDidDocument(didDocument, patch);
+    }
+  }
+
+  /**
+   * Applies the given patch to the given DID Document.
+   */
+  private static applyPatchToDidDocument (didDocument: DidDocumentModel, patch: any) {
+    if (patch.action === 'add-public-keys') {
+      const publicKeySet = new Set(didDocument.publicKey.map(key => key.id));
```
mudiali: Nit: maybe `publicKeyIdSet` a better name.

lib/core/versions/latest/DocumentComposer.ts
```diff
@@ -166,4 +171,123 @@ export default class DocumentComposer {
...
+      const publicKeyMap = new Map(didDocument.publicKey.map(publicKey => [publicKey.id, publicKey]));
+
+      // Loop through all given public key IDs and delete them from the existing public key only if it is not a recovery key.
+      for (let publicKey of patch.publicKeys) {
+        const existingKey = publicKeyMap.get(publicKey);
+
+        // Deleting recovery key is NOT allowed.
+        if (existingKey !== undefined &&
+            existingKey.usage !== KeyUsage.recovery) {
+          publicKeyMap.delete(publicKey);
```
mudiali: Shouldn't this be a throw for the `else` case here? If we do want to throw, maybe the check belongs in the corresponding `validate ...` logic but I understand if we don't want to do the find in the document there.

lib/core/versions/latest/DocumentComposer.ts
```diff
@@ -166,4 +171,123 @@ export default class DocumentComposer {
...
+          }
+        };
+
+        if (didDocument.service === undefined) {
+          didDocument.service = [service];
+        } else {
+          didDocument.service.push(service);
+        }
+      } else {
+        // Else we add to the eixsting service element.
```
mudiali:
```suggestion
        // Else we add to the existing service element.
```

lib/core/versions/latest/DocumentComposer.ts
thehenrytsai: Done!

lib/core/versions/latest/DocumentComposer.ts
thehenrytsai: Done.

lib/core/versions/latest/RequestHandler.ts
thehenrytsai: Great suggestion. Done.

lib/core/versions/latest/DocumentComposer.ts
thehenrytsai: Concept of `usage` is no longer there and will be removed in my next round of refactoring (yes, there is at least one more round of refactoring...), it is no longer possible to remove a recovery key at all because the recovery key is not even in the document anymore. Will add a comment with issue number tracking removal of `usage`.

Issue #417, opened 2020-03-02
combine ipfs, bitcoin, and core error into a common error class


isaacJChen:

OR13:
I assume you mean something like "SidetreeCoreError" with something like https://www.npmjs.com/package/serialize-error

?

sambarnes:
@isaacJChen took a first pass at the issue in the above PR. Just a bunch of renaming and updating import paths. Let me know if you think more functionality needs added (like the `serialize-error` package Orie mentioned?) or if I was off base.

PR #433, opened 2020-03-11
#417 - Combine Error Classes

sambarnes:
In response to #417

Combine `lib/bitcoin/BitcoinError`, `lib/core/SidetreeError`, and `lib/ipfs/IpfsError` into a single `lib/common/SidetreeError`.

Saw two of three packages used `ErrorCode.ts` while the third used `CoreErrorCode.ts`. Renamed it for consistency. Should `lib/common/SharedErrorCode.ts` be renamed as well? I suppose the "shared" aspect is already conveyed by the fact that its located in `common`, but perhaps explicit is better.

Also, I left the `ErrorCode.ts` files in their respective packages with the idea that package specific error codes should remain colocated with the package they're used in, while the shared `SidetreeError` is in the common package.


sambarnes:
@thehenrytsai appreciate the review! should be all resolved now

PR #437, opened 2020-03-11
Removed references to KeyUsage

thehenrytsai:
1. Removed references to KeyUsage.
1. Changed key add patch behavior and changed tests accordingly.


lib/core/versions/latest/OperationProcessor.ts
```diff
@@ -119,8 +119,12 @@ export default class OperationProcessor implements IOperationProcessor {
     let resultingDocument;
     try {
       resultingDocument = await DocumentComposer.applyUpdateOperation(operation, documentState.document);
-    } catch {
-      // Retrun the given document state if error is encountered applying the update.
+    } catch (error) {
+      const didUniqueSuffix = anchoredOperationModel.didUniqueSuffix;
+      const transactionNumber = anchoredOperationModel.transactionNumber;
+      console.debug(`Unable to apply document patch in transaction number ${transactionNumber} for DID ${didUniqueSuffix}: ${error.toString()}.`);
```
mudiali: I suggest that we add more details about the exception: `JSON.stringify(error, Object.getOwnPropertyNames(error)` while logging.

OR13: we should really be using a logger, instead of console, so we can store the logs in various cloud services.

thehenrytsai: There is a separate debate between using a dedicated logger or logger hook that traps logs written as console. We can take this as an agenda item for our weekly meeting. For now I am going to add a method for logging Sidetree errors.

PR #441, opened 2020-03-14
#271 Update the logic to determine the sidetree txn writer.

mudiali:
A valid sidetree transaction writer has the following requirements:
1. There must be at least one input.
1. The first input must be a in the fomat: <signature> <publickey>
1. The output being spent in the first input must be a pay-to-public-key-hash format.

This way we will prove that the writer of the first input (and hence the txn) had access to the <publickey><privatekeypair>. The writer will be the hash-of-<publickey>.


lib/bitcoin/SidetreeTransactionParser.ts
```diff
@@ -0,0 +1,152 @@
...
+      return undefined;
+    }
+
+    return {
+      data: sidetreeData,
+      writer: writer
+    };
+  }
+
+  private getValidSidetreeDataFromOutputs (transactionId: string, transactionOutputs: BitcoinOutputModel[], sidetreePrefix: string): string | undefined {
```
thehenrytsai: It seems odd to pass `transactionId` just for logging purposes. I was thinking, maybe we tweak the requirement to accept the first output valid Sidetree data only, (and perhaps even only look at the first output only), then short circuit the code. This would help simplify the code and also reduce compute.

mudiali: Hmmm .. so the original behavior has been to actively remove any transactions which have multiple anchorstrings. I'm ok with your suggestion but the implementation does mean adding the outputIdx to the BitcoinOutputModel to ensure that we actually get the first one.

lib/bitcoin/SidetreeTransactionParser.ts
```diff
@@ -0,0 +1,152 @@
...
+    const outputBeingSpend = await this.fetchOutput(inputToCheck.previousTransactionId, inputToCheck.outputIndexInPreviousTransaction);
+
+    if (!outputBeingSpend) {
+      return undefined;
+    }
+
+    return this.getPublicKeyHashIfValidScript(outputBeingSpend.scriptAsmAsString);
+  }
+
+  private async fetchOutput (transactionId: string, outputIdxToFetch: number): Promise<BitcoinOutputModel | undefined> {
```
thehenrytsai:
```suggestion
  private async fetchOutput (transactionId: string, outputIndexToFetch: number): Promise<BitcoinOutputModel | undefined> {

```

lib/bitcoin/SidetreeTransactionParser.ts
mudiali: Updating to recognize only the first anchorstring found in the outputs. The underline bitcore-lib returns the outputs in the order that they were written on the blockchain so we are going to take the order in which they were returned.

PR #446, opened 2020-03-18
#266 - Updated map and batch file fetching to be compatible with data pruning

thehenrytsai:
1. Updated map and batch file fetching to be compatible with data pruning.
1. Bring code coverage of `TransactionProcessor` to 100%.


lib/core/versions/latest/TransactionProcessor.ts
```diff
@@ -91,60 +91,112 @@ export default class TransactionProcessor implements ITransactionProcessor {
...
-    const batchFileModel = await BatchFile.parse(fileBuffer);
+    try {
+      const batchFileHash = mapFile.batchFileHash;
+      console.info(`Downloading batch file '${batchFileHash}', max size limit ${ProtocolParameters.maxBatchFileSizeInBytes}...`);
+
+      const fileBuffer = await this.downloadFileFromCas(batchFileHash, ProtocolParameters.maxBatchFileSizeInBytes);
+      const batchFileModel = await BatchFile.parse(fileBuffer);
+
+      return batchFileModel;
+    } catch (error) {
```
mudiali: Nit: 3 is a pattern :) I'm not sure what is a good way to avoid the code duplication (or whether we shoud do it at all). Maybe have a function `isErrorRetryable` is enough? Up to you whether you want to fix or not.

tests/core/TransactionProcessor.spec.ts
```diff
@@ -11,6 +11,9 @@ import MockOperationStore from '../mocks/MockOperationStore';
 import OperationGenerator from '../generators/OperationGenerator';
 import TransactionModel from '../../lib/common/models/TransactionModel';
 import TransactionProcessor from '../../lib/core/versions/latest/TransactionProcessor';
+import MapFile from '../../lib/core/versions/latest/MapFile';
+import SidetreeError from '../../lib/common/SidetreeError';
+import BatchFile from '../../lib/core/versions/latest/BatchFile';
```
isaacJChen: Nit: sort imports

lib/core/versions/latest/TransactionProcessor.ts
thehenrytsai: The two new try-catch blocks differs from the parent one in that it treats unknown errors as not retryable. So I am thinking to not move the two methods into a method.

tests/core/TransactionProcessor.spec.ts
thehenrytsai: Fixed.

lib/core/versions/latest/TransactionProcessor.ts
mudiali: :thumb

PR #447, opened 2020-03-18
#271 Adding value-time-lock support to the batch writer and observer

mudiali:


lib/core/versions/latest/BatchWriter.ts
```diff
@@ -77,4 +83,10 @@ export default class BatchWriter implements IBatchWriter {
     // Remove written operations from queue after batch writing has completed successfully.
     await this.operationQueue.dequeue(queuedOperations.length);
   }
+
+  private getNumberOfOpsToWrite (valueTimeLock: ValueTimeLockModel | undefined, normalizedFee: number): number {
+    const numberOfOpsAllowed = ValueTimeLockVerifier.calculateMaxNumberOfOpsAllowed(valueTimeLock, normalizedFee);
+
+    return Math.min(numberOfOpsAllowed, ProtocolParameters.maxOperationsPerBatch);
```
isaacJChen: Should we log some sort of message here if the maxOperationPerBatch is lower than numberOfOpsAllowed so the admin knows when they are "wasting" money in lock?

lib/core/Blockchain.ts
```diff
@@ -218,11 +218,6 @@ export default class Blockchain implements IBlockchain {
     const responseBody = await JsonAsync.parse(responseBodyString);
 
     if (response.status === HttpStatus.NOT_FOUND) {
-
-      if (responseBody.code === SharedErrorCode.ValueTimeLockInPendingState) {
-        throw new SidetreeError(SharedErrorCode.ValueTimeLockInPendingState);
```
thehenrytsai: Move `ValueTimeLockInPendingState` back to bicoin service's own `ErrorCode`.

lib/core/versions/latest/AnchorFile.ts
```diff
@@ -61,6 +63,13 @@ export default class AnchorFile {
       throw new SidetreeError(ErrorCode.AnchorFileMissingOperationsProperty);
     }
 
+    // This property is optional so if there's an additional propertye besides
+    // the required ones then it must be this optional one.
+    if (anchorFileProperties.length === maxAllowedPropertyCount &&
+       !anchorFileModel.hasOwnProperty('writerLock')) {
+      throw new SidetreeError(ErrorCode.AnchorFileHasUnknownProperty);
```
thehenrytsai: Nit: In principal we try to never reuse an error code, so consider either:

1. Create a new error code. or
1. Attach custom error message, or
1. Follow the pattern found in other validation handling optional parameters, e.g.
```javascript
    const allowedProperties = new Set(['batchFileHash', 'updateOperations']);
    for (let property in mapFile) {
      if (!allowedProperties.has(property)) {
        throw new SidetreeError(ErrorCode.MapFileHasUnknownProperty);
      }
    }
```

lib/core/versions/latest/AnchorFile.ts
```diff
@@ -170,6 +180,7 @@ export default class AnchorFile {
     });
 
     const anchorFileModel = {
+      writerLock: writerLock,
```
thehenrytsai: Nit:
```suggestion
      writerLock,
```

lib/core/versions/latest/ValueTimeLockVerifier.ts
```diff
@@ -0,0 +1,77 @@
...
+   * Calculates the maximum number of operations allowed to be written for the given lock information. If
+   * there is no lock then it returns the number of operations which do not require a lock.
+   *
+   * @param valueTimeLock The lock object if exists
+   * @param normalizedFee The normalized fee for the current block
+   */
+  public static calculateMaxNumberOfOpsAllowed (valueTimeLock: ValueTimeLockModel | undefined, normalizedFee: number) {
+
+    if (valueTimeLock === undefined) {
+      return ProtocolParameters.maxNumberOfOpsForNoValueTimeLock;
```
thehenrytsai: In the spirit of not introducing further parameters unless we must, do we want to derive `maxNumberOfOpsForNoValueTimeLock` from `normalizedFeeToPerOperationFeeMultiplier`?

If we change `normalizedFeeToPerOperationFeeMultiplier` to `normalizedFeeToPerOperationFeeDivisor`, then we can just use the value as is.

lib/core/versions/latest/ValueTimeLockVerifier.ts
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    // Using the following formula:
+    //  requiredLockAmount = normalizedfee * normalizedFeeMultipier * numberOfOps * valueTimeLockMultiplier
+    //
+    // We are going to find the numberOfOps given the requiredLockAmount
+    const feePerOperation = normalizedFee * ProtocolParameters.normalizedFeeToPerOperationFeeMultiplier;
+    const numberOfOpsAllowed = valueTimeLock.amountLocked / (feePerOperation * ProtocolParameters.valueTimeLockAmountMultiplier);
```
thehenrytsai: We should return the max free ops count if the `numberOfOpsAllowed` calculated turns out to be lower.

lib/core/versions/latest/BatchWriter.ts
```diff
@@ -77,4 +83,10 @@ export default class BatchWriter implements IBatchWriter {
     // Remove written operations from queue after batch writing has completed successfully.
     await this.operationQueue.dequeue(queuedOperations.length);
   }
+
+  private getNumberOfOpsToWrite (valueTimeLock: ValueTimeLockModel | undefined, normalizedFee: number): number {
+    const numberOfOpsAllowed = ValueTimeLockVerifier.calculateMaxNumberOfOpsAllowed(valueTimeLock, normalizedFee);
```
thehenrytsai: Consider just let `calculateMaxNumberOfOpsAllowed` perform the `Math.min()` check.

lib/core/versions/latest/TransactionProcessor.ts
```diff
@@ -88,6 +90,18 @@ export default class TransactionProcessor implements ITransactionProcessor {
...
     }
 
+    // Verify required lock if one was needed.
+    const valueTimeLock = anchorFile.model.writerLock
+                          ? await this.blockchain.getValueTimeLock(anchorFile.model.writerLock)
+                          : undefined;
+
+    ValueTimeLockVerifier.verifyLockAmountAndThrowOnError(
+      valueTimeLock,
+      operationCountInAnchorFile,
```
thehenrytsai: Sanity: should this be `paidOperationCount`?

lib/core/versions/latest/ValueTimeLockVerifier.ts
```diff
@@ -0,0 +1,77 @@
...
+   * @param valueTimeLock The value time lock object used for verificiation.
+   * @param numberOfOperations The target number of operations.
+   * @param normalizedFee The normalized fee for the target block.
+   * @param targetTransactionTime The transaction time where the operations were written.
+   */
+  public static verifyLockAmountAndThrowOnError (
+    valueTimeLock: ValueTimeLockModel | undefined,
+    numberOfOperations: number,
+    normalizedFee: number,
+    targetTransactionTime: number,
```
thehenrytsai: Consider rename:
```suggestion
    sidetreeTransactionTime: number,

```

lib/core/versions/latest/ValueTimeLockVerifier.ts
```diff
@@ -0,0 +1,77 @@
...
+    targetTransactionWriter: string): void {
```
thehenrytsai: Consider rename:
```suggestion
    sidetreeTransactionWriter: string): void {

```

lib/core/versions/latest/ValueTimeLockVerifier.ts
```diff
@@ -0,0 +1,77 @@
...
+          targetTransactionTime >= valueTimeLock.unlockTransactionTime) {
+        throw new SidetreeError(
+          ErrorCode.ValueTimeLockTargetTransactionTimeOutsideLockRange,
+          `Target block: ${targetTransactionTime}; lock start time: ${valueTimeLock.lockTransactionTime}; unlock time: ${valueTimeLock.unlockTransactionTime}`);
+      }
+    }
+
+    const maxNumberOfOpsAllowed = this.calculateMaxNumberOfOpsAllowed(valueTimeLock, normalizedFee);
+
+    if (numberOfOperations > maxNumberOfOpsAllowed) {
```
thehenrytsai: Sanity: doesn't look right, 'coz this means a spammer can bypass the protocol maximum if we don't check for it. Come back to the comment earlier about considering having the protocol maximum check within `calculateMaxNumberOfOpsAllowed()`.

lib/core/versions/latest/models/ProtocolParameters.ts
```diff
@@ -14,6 +14,8 @@ export default interface ProtocolParameters {
   maxEncodedHashStringLength: number;
   /** Max number of operations observed per transaction time */
   maxNumberOfOpsPerTransactionTime: number;
+  /** Maximum number of operations allowed with no lock. */
+  maxNumberOfOpsForNoValueTimeLock: number;
```
thehenrytsai: Prefer naming consistency: (not sure how`maxNumberOfOpsPerTransactionTime` fell through review cracks). Open to have the change `operations` to `ops` throughout discussion 'coz I get it gets rather wordy....
```suggestion
  maxNumberOfOperationsForNoValueTimeLock: number;
```

lib/core/versions/latest/ValueTimeLockVerifier.ts
```diff
@@ -0,0 +1,77 @@
...
+export default class ValueTimeLockVerifier {
+
+  /**
+   * Calculates the maximum number of operations allowed to be written for the given lock information. If
+   * there is no lock then it returns the number of operations which do not require a lock.
+   *
+   * @param valueTimeLock The lock object if exists
+   * @param normalizedFee The normalized fee for the current block
+   */
+  public static calculateMaxNumberOfOpsAllowed (valueTimeLock: ValueTimeLockModel | undefined, normalizedFee: number) {
```
thehenrytsai: Prefer consistency with existing code. Open to have the change from `operations` to `ops` throughout discussion 'coz I get it gets rather wordy....
```suggestion
  public static calculateMaxNumberOfOperationsAllowed (valueTimeLock: ValueTimeLockModel | undefined, normalizedFee: number) {

```

lib/core/versions/latest/ValueTimeLockVerifier.ts
```diff
@@ -0,0 +1,77 @@
...
+          ErrorCode.ValueTimeLockTransactionWriterLockOwnerMismatch,
+          `Transaction writer: ${targetTransactionWriter} - Lock owner: ${valueTimeLock.owner}`);
+      }
+
+      // Check the lock duration
+      if (targetTransactionTime < valueTimeLock.lockTransactionTime ||
+          targetTransactionTime >= valueTimeLock.unlockTransactionTime) {
+        throw new SidetreeError(
+          ErrorCode.ValueTimeLockTargetTransactionTimeOutsideLockRange,
+          `Target block: ${targetTransactionTime}; lock start time: ${valueTimeLock.lockTransactionTime}; unlock time: ${valueTimeLock.unlockTransactionTime}`);
```
thehenrytsai:
```suggestion
          `Sidetree transaction time: ${targetTransactionTime}; lock start time: ${valueTimeLock.lockTransactionTime}; unlock time: ${valueTimeLock.unlockTransactionTime}`);

```

lib/core/versions/latest/ValueTimeLockVerifier.ts
```diff
@@ -0,0 +1,77 @@
...
+   * @param targetTransactionTime The transaction time where the operations were written.
+   */
+  public static verifyLockAmountAndThrowOnError (
+    valueTimeLock: ValueTimeLockModel | undefined,
+    numberOfOperations: number,
+    normalizedFee: number,
+    targetTransactionTime: number,
+    targetTransactionWriter: string): void {
+
+    if (valueTimeLock) {
```
thehenrytsai: Discussion: do we want to let the verification succeed if the lock verification fails but the actual operation count is within the "free" limit?

lib/core/versions/latest/BatchWriter.ts
mudiali: Adding a log statement here.

lib/core/versions/latest/AnchorFile.ts
mudiali: Really good point. Will update.

lib/core/versions/latest/AnchorFile.ts
mudiali: Renaming it to `writerLockId` to match the `.getWriterLock(...)` function that is in the `IBlockchain`.

lib/core/versions/latest/BatchWriter.ts
mudiali: I want to keep the ValueTimeLockVerifier to only care about the lock related checks. The `maxOperationsPerBatch` is a protocol check which belongs outside of the verifier IMO. I am refactoring the code to make it more clearer.

lib/core/versions/latest/ValueTimeLockVerifier.ts
mudiali: The `maxOperationPerBatch` check is performed outside in `TransactionProcessor.downloadAndVerifyAnchorFile(...)` function.

lib/core/versions/latest/models/ProtocolParameters.ts
mudiali: Renaming.

lib/core/versions/latest/ValueTimeLockVerifier.ts
mudiali: Changing the name.

lib/core/versions/latest/TransactionProcessor.ts
mudiali: Good catch. Done.

lib/core/versions/latest/ValueTimeLockVerifier.ts
mudiali: Keeping the two values separate to keep the max-free-ops and normalized-fee separate.

lib/core/versions/latest/ValueTimeLockVerifier.ts
mudiali: Good idea. Done.

mudiali: Done.

lib/core/versions/latest/BatchWriter.ts
```diff
@@ -85,8 +85,14 @@ export default class BatchWriter implements IBatchWriter {
   }
 
   private getNumberOfOpsToWrite (valueTimeLock: ValueTimeLockModel | undefined, normalizedFee: number): number {
```
thehenrytsai: Nit:
```suggestion
  private getNumberOfOperationsToWrite (valueTimeLock: ValueTimeLockModel | undefined, normalizedFee: number): number {
```

PR #452, opened 2020-03-20
#271 Adding lock duration check

mudiali:


lib/bitcoin/lock/LockResolver.ts
```diff
@@ -163,4 +170,22 @@ export default class LockResolver {
 
     return blockInfo.height;
   }
+
+  private isLockDurationValid (startBlock: number, unlockBlock: number): boolean {
```
thehenrytsai: To make the protocol rule clear, I think we should not give leeway at verification side (else the protocol would have to say: x number of blocks + 2 is acceptable). It's better to add the leeway on the lock creation side such that they create a lock with duration greater than the protocol minimum.

mudiali: Hmmm ... good point.

mudiali: Extending the lock duration by ~1 day. Giving a long-ish duration as once the lock transaction is written the lock is there for the full duration (1 month currently) so want to give enough time just in case.

Issue #453, opened 2020-03-20
Sidetree has defined Method Versioning Scheme


OR13:
https://github.com/decentralized-identity/ethr-did-resolver/blob/develop/doc/did-method-spec.md

As discussed on the WG call 2020-03-05.

We are proposing to adopt the following method specific versioning scheme prefix:

- did:ion:testnet3:Ei1d…
- did:elem:ropsten:Ei1d…

OR13:
@csuwildcat @peacekeeper any objections to formalizing this.

OR13:
Implemented

PR #465, opened 2020-03-23
Add Method Specific Versioning to Spec

OR13:
Addresses: https://github.com/decentralized-identity/sidetree/issues/453

- Reduce number of directories to `./docs`.
- Add language about did method network suffix use


OR13:
@csuwildcat ^ see how index.html is conflicted! this is why it must be ignored.

PR #468, opened 2020-03-23
spec: generic DID method in text

troyronda:
- replace ION reference with Sidetree.
- make ion and elem examples.

note: the SVG diagram should also be updated with generic labels rather than ION.


troyronda:
Created #490

Issue #449, opened 2020-03-19
Change create & recover operation request to supply the initial document state using patches


thehenrytsai:

OR13:
Create is a patch from `{}` to: did document without id, and using fragids for all `id` properties that are normally required to be URIs...

Recover is a patch from `{ existing did core spec compliant doc1 }` to `{ existing did core spec compliant doc2 }`, where doc1 is completely replaces with doc2.

Both should use a full document update json patch IMO... unless we are confident that we know the JSON Schema for a did document... (we don't) see: https://github.com/w3c/did-core-registries/pull/11

OR13:
We discussed 2 proposals:

1. Keep everything the same.

2. Wrap Create and Recovery operations in patches, so that all operations are patches... so the API is consistent.

We seem to have rough agreement on: 2... understanding that it adds a small amount of bloat, but makes the API (Operations Structures) more consistent.

We agreed that we can use "Patches" without agreeing that they be RFC JSON Patches or JSON Schema custom patches... We agreed that there are 2 locations where we want to address performance:

1. Observation Time
2. Resolution Time

We agreed that DID Methods can make different choices here.

Issue #455, opened 2020-03-21
Remove/deactivate service endpoint support 


thehenrytsai:
The current support for service endpoint is limited. Need to clean it up before mainnet.

csuwildcat:
We may even be able to cut it for this release

csuwildcat:
I don't want to completely remove the concept of this from the code, given there are some rather straightforward service endpoint types I would like to make available as quickly as we can after this mainnet freeze/stabilization phase.

PR #477, opened 2020-03-25
#449 - Changed operation requests to supply the initial document state using patches

thehenrytsai:
1. Changed operation requests to supply the initial document state using patches.
1. Unified the `OperationDataModel` between create, recover, and update operations.


lib/core/versions/latest/DocumentComposer.ts
```diff
@@ -219,100 +221,136 @@ export default class DocumentComposer {
...
-        for (let serviceEndpoint of patch.serviceEndpoints) {
-          if (!serviceEndpointSet.has(serviceEndpoint)) {
-            service.serviceEndpoint.instances.push(serviceEndpoint);
-          }
+    // If service not found, create a new service element and add it to the property.
+    if (service === undefined) {
+      service = {
+        type: patch.serviceType,
+        serviceEndpoint: {
+          '@context': 'schema.identity.foundation/hub',
```
OR13: Not a valid URI.

thehenrytsai: Thanks for calling out, this whole service endpoint will be revamped completely within the week by issue #455.

PR #478, opened 2020-03-25
rename otp to commit reveal

isaacJChen:


docs/protocol.md
```diff
@@ -429,7 +430,7 @@ POST / HTTP/1.1
 {
   "type": "update",
   "didUniqueSuffix": "The unique suffix of the DID to be updated.",
-  "updateOtp": "The one-time password to be used for this update.",
+  "updateRevealValue": "Commitment hash to be used for this update.",
```
isaacJChen: This is a mistake, this should be revealvalue

docs/protocol.md
```diff
@@ -65,7 +65,7 @@ The DID unique suffix is computed deterministically by hashing, then encoding th
 
 1. Hash of the initial document.
 1. Recovery key.
-1. Hash of one-time password for recovery.
+1. Hash of reveal value for recovery, aka commitment hash.
```
isaacJChen: of the commitment value

docs/protocol.md
```diff
@@ -305,15 +306,15 @@ POST / HTTP/1.1
   "recoveryKey": {
     "publicKeyHex": "A SECP256K1 public key expressed in compressed HEX format."
   },
-  "nextRecoveryOtpHash": "Hash of the one-time password to be used for the next recovery."
+  "nextRecoveryCommitmentHash": "Commitment hash to be used for the next recovery."
```
thehenrytsai: Same for all instances.
```suggestion
  "nextRecoveryCommitmentHash": "Commitment hash for the next recovery."
```

lib/core/versions/latest/DocumentComposer.ts
```diff
@@ -239,7 +239,7 @@ export default class DocumentComposer {
       // Loop through all given public keys and add them if they don't exist already.
       for (let publicKey of patch.publicKeys) {
         // NOTE: If a key ID already exists, we will just replace the existing key.
-        // Not throwing error will minimize the need (thus risk) of reusing exposed update OTP.
+        // Not throwing error will minimize the need (thus risk) of reusing exposed update commitment/reveal value.
```
thehenrytsai: Please rename all occurrences of "commitment/reveal value" to just "reveal value".

lib/core/versions/latest/Operation.ts
```diff
@@ -11,8 +11,8 @@ import UpdateOperation from './UpdateOperation';
  * A class that contains Sidetree operation utility methods.
  */
 export default class Operation {
-  /** Maximum allowed encoded OTP string length. */
-  public static readonly maxEncodedOtpLength = 50;
+  /** Maximum allowed encoded commitment/reveal value string length. */
+  public static readonly maxEncodedCommitmentRevealValueLength = 50;
```
thehenrytsai:
```suggestion
  public static readonly maxEncodedRevealValueLength = 50;
```

lib/core/versions/latest/CreateOperation.ts
```diff
@@ -12,11 +12,11 @@ import SidetreeError from '../../../common/SidetreeError';
 interface SuffixDataModel {
   operationDataHash: string;
   recoveryKey: PublicKeyModel;
-  nextRecoveryOtpHash: string;
+  nextRecoveryCommitmentHash: string;
```
thehenrytsai: Discussion: not blocking check in, but I thought I would be fine with "CommitmentHash" but the more I think about it the more I think it should be just "Commitment", one of the reason is that if we ever change the way we compute the commitment, then calling it a hash would not be correct.

docs/protocol.md
thehenrytsai: I prefer the way it is.

tests/generators/OperationGenerator.ts
```diff
@@ -121,7 +121,7 @@ export default class OperationGenerator {
     const hubServiceEndpoint = 'did:sidetree:value0';
     const service = OperationGenerator.createIdentityHubUserServiceEndpoints([hubServiceEndpoint]);
 
-    // Generate the next update and recover operation commitment/reveal pair.
```
thehenrytsai: I thought this was fine, but no need to change back now.

lib/core/versions/latest/OperationProcessor.ts
```diff
@@ -167,7 +167,7 @@ export default class OperationProcessor implements IOperationProcessor {
 
     const operation = await RecoverOperation.parse(anchoredOperationModel.operationBuffer);
 
-    // Verify the commit/reveal value hash.
+    // Verify the reveal value hash.
```
thehenrytsai: Some what odd, but I guess it is technically accurate.

Issue #321, opened 2019-10-29
Replace `revoke` operation as `deactivate` operation instead


thehenrytsai:
1. Replace all referenced to `delete` in code and documentation as `revoke` going forward.
1. Disambiguate the revoked from not-found when resolving.
1. Revocation must be performed by `recover` key.
1. Revocation is the final valid operation.

OR13:
Is this really https://w3c.github.io/did-core/#deactivate ?

I suggest that deactivated DIDs have their total information minimized, forensic analysis will return all the versions before the point of deactivation, but I still think it would be wise to set the DID Document to basically:

```
{
"@context": "...",
"id":  "did:example:123",
"deactivated": true
}
```
Is there a reason why the DID Document should retain anything that would be useful for correlation in a deactivated state? Certainly the user will not be able to do anything about it, after the fact.

peacekeeper:
There's an older issue about this in the DID resolution spec repo.. After a DID has been deactivated, and you resolve it, do you get back a DID document with a special flag (as in your example), or do you get back a null value, or does the DID resolver raise an error?
See here: https://github.com/w3c-ccg/did-resolution/issues/5

peacekeeper:
I put this on the agenda for tomorrow's DID resolution call, in case you have time: https://docs.google.com/document/d/1qYBaXQMUoB86Alquu7WBtWOxsS8SMhp1fioYKEGCabE/

csuwildcat:
@thehenrytsai we should rename all of our Revoke refs and update the docs/spec to reflect Deactivate as the operation type. I can update the spec.

OR13:
implemented

OR13:
@thehenrytsai @csuwildcat I'm removing the spec v1 label, as this is implemented there.

csuwildcat:
I believe this change is now reflected everywhere in the spec as of https://github.com/decentralized-identity/sidetree/pull/523, but eyes-on would help validate.

PR #487, opened 2020-03-25
#486 - Renamed DocumentState to DidState

thehenrytsai:
1. Renamed DocumentState to DidState.
1. Fixed various opportunistic exports to follow convention (default exports).


docs/protocol.md
```diff
@@ -51,10 +51,16 @@ The following lists the parameters used by this version of the Sidetree protocol
...
-A [_DID Document_](https://w3c-ccg.github.io/did-spec/#ex-2-minimal-self-managed-did-document
-) is a document containing information about a DID, such as the public keys of the DID owner and service endpoints used. Sidetree protocol enables the creation of, lookup for, and updates to DID Documents through _Sidetree operations_.
+Sidetree protocol allows the following operations to be performed against a DID:
 
-An update operation to a document contains only the changes from the previous version of the document. Create and recover operations require a full document state of the DID as input.
+1. Create
+1. Resolve
+1. Update
+1. Recover
+1. Revoke
```
mudiali: `Deactivate`?

thehenrytsai: I'll leave it as is and let #321 handle it. I want the doc to match the code, and plus we will need to remove that portion of the doc and refer to the new spec document going forward.

PR #488, opened 2020-03-26
#271 Fixing bugs after doing e2e LockMonitor testing

mudiali:
1. Any errors during LockMonitor.initialize(...) should not be suppressed. It is better to stop the execution at the service startup if there are any issues then.
1. If the satoshi client is unable to calculate the fee, then catch and rethrow those errors. Previously NaN feerate was being returned and the error was being thrown much later hiding the original problem.
1. Using the bitcore-lib utility functions to convert BTC to satoshis. Manual conversion was causing some decimal errors if the BTC amount was too small.
1. Fixing some lint errors in master branch.


lib/bitcoin/BitcoinClient.ts
```diff
@@ -387,9 +398,16 @@ export default class BitcoinClient {
     };
 
     const response = await this.rpcCall(request, true);
+
+    if (!response.feerate ||
+        (response.errors && response.errors.length > 0)) {
+      const error = response.errors ? JSON.stringify(response.errors) : `Feerate is undefined`;
+      throw new Error(`Fee rate could not be estimated. Error: ${error}`);
```
thehenrytsai: Sanity: Should we throw SidetreeError?

mudiali: The BitcoinClient does not throw any SidetreeErrors so I want to change it all together. Let me open an issue for that.

PR #499, opened 2020-03-26
rework service endpoint (issue #455)

isaacJChen:


lib/core/versions/latest/DocumentComposer.ts
```diff
@@ -84,18 +84,7 @@ export default class DocumentComposer {
...
-          throw new SidetreeError(ErrorCode.DocumentComposerServiceTypeNotString);
-        }
-
-        // 'serviceEndpoint' is required.
-        const serviceEntryType = typeof serviceEntry.serviceEndpoint;
-        if (serviceEntryType !== 'string' && serviceEntryType !== 'object') {
-          throw new SidetreeError(ErrorCode.DocumentComposerServiceEndpointMissingOrIncorrectType);
-        }
-      }
+      DocumentComposer.validateServiceEndpoints(document.service);
```
thehenrytsai: Should be changed to `serviceEndpoints` according to the new paradigm (specified in protocol.md), basically we are decoupling our document state from the DID spec. Note that I have already changed `publicKey` to `publicKeys` in my recent PR.

lib/core/versions/latest/DocumentComposer.ts
```diff
@@ -84,18 +84,7 @@ export default class DocumentComposer {
       }
```
thehenrytsai: Move the array check into `validateServiceEndpoints`.

lib/core/versions/latest/DocumentComposer.ts
```diff
@@ -197,25 +186,61 @@ export default class DocumentComposer {
...
+      if (id.length > 20) {
+        throw new SidetreeError(ErrorCode.DocumentComposerPatchServiceEndpointIdsIdTooLong);
+      }
     }
+  }
 
+  /**
+   * Validates update patch for adding service endpoints.
+   */
+  private static validateAddServiceEndpointsPatch (patch: any) {
```
thehenrytsai: Related to the comment above, the array check is common and can move into `validateServiceEndpoints`. Also probably need to add check for patch property length like `validateAddPublicKeysPatch()`, so that patch with unknown properties won't be considered valid.

lib/core/versions/latest/DocumentComposer.ts
```diff
@@ -197,25 +186,61 @@ export default class DocumentComposer {
...
     }
 
-    for (let serviceEndpoint of patch.serviceEndpoints) {
-      if (typeof serviceEndpoint !== 'string') {
-        throw new SidetreeError(ErrorCode.DocumentComposerPatchServiceEndpointNotString);
+    DocumentComposer.validateServiceEndpoints(patch.serviceEndpoints);
+  }
+
+  private static validateServiceEndpoints(serviceEndpoints: any[]) {
+    for (let serviceEndpoint of serviceEndpoints) {
```
thehenrytsai: Need to make sure there are no unknown properties.

lib/core/versions/latest/DocumentComposer.ts
```diff
@@ -293,63 +318,52 @@ export default class DocumentComposer {
...
-    const serviceEndpointSet = new Set(service.serviceEndpoint.instances);
+    const idToRemove = patch.serviceEndpointIds;
 
-    // Loop through all given public key IDs and add them from the existing public key set.
-    for (let serviceEndpoint of patch.serviceEndpoints) {
-      serviceEndpointSet.delete(serviceEndpoint);
+    for (const id of idToRemove) {
+      if (idToIndexMapper.has(id)) {
+        const idx = idToIndexMapper.get(id);
+        document.service.splice(idx, 1);
```
thehenrytsai: Sanity: is this a bug? The fact that splice occurs changes original indices.

lib/core/versions/latest/DocumentComposer.ts
```diff
@@ -197,25 +186,61 @@ export default class DocumentComposer {
...
       throw new SidetreeError(ErrorCode.DocumentComposerPatchServiceEndpointsNotArray);
     }
 
-    for (let serviceEndpoint of patch.serviceEndpoints) {
-      if (typeof serviceEndpoint !== 'string') {
-        throw new SidetreeError(ErrorCode.DocumentComposerPatchServiceEndpointNotString);
+    DocumentComposer.validateServiceEndpoints(patch.serviceEndpoints);
+  }
+
+  private static validateServiceEndpoints(serviceEndpoints: any[]) {
```
thehenrytsai: Sanity: Probably didn't run linter check?
```suggestion
  private static validateServiceEndpoints (serviceEndpoints: any[]) {
```

lib/core/versions/latest/RequestHandler.ts
```diff
@@ -102,6 +102,7 @@ export default class RequestHandler implements IRequestHandler {
             body: document
           };
           break;
+        // these cases do nothing because we cannot return an original document
```
thehenrytsai:
```suggestion
        // these cases do nothing because we do not know the latest document state unless we resolve.
```

tests/common/DidServiceEndpoint.ts
```diff
@@ -7,14 +7,11 @@ export default class DidServiceEndpoint {
   /**
    * Test if a MongoDB service is running at the specified url.
    */
-  public static createHubServiceEndpoint (instances: string[]): DidServiceEndpointModel {
```
thehenrytsai: Can we remove this class altogether or maybe tweak the method to be more general purpose because:
1. "Hub" is no longer a used term.
1. This method has become quite pointless, if it's only not hardcoding ID.
1. The description became incorrect at some point.

tests/generators/OperationGenerator.ts
```diff
@@ -514,27 +514,25 @@ export default class OperationGenerator {
     didUniqueSuffix: string,
     updateRevealValue: string,
     nextUpdateCommitmentHash: string,
-    endpointsToAdd: string[],
-    endpointsToRemove: string[],
+    idToAdd: string | undefined,
+    idsToRemove: string[],
```
thehenrytsai:
```suggestion
    idOfServiceEndpointToAdd: string | undefined,
    idsOfServiceEndpointsToRemove: string[],
```

tests/generators/OperationGenerator.ts
```diff
@@ -578,18 +576,14 @@ export default class OperationGenerator {
 
   /**
    * Generates a single element array with a identity hub service object for DID document
-   * @param instances the instance field in serviceEndpoint. A list of DIDs
+   * @param id the id field in serviceEndpoint.
    */
-  public static createIdentityHubUserServiceEndpoints (instances: string[]): any[] {
+  public static createIdentityHubUserServiceEndpoints (id: string): any[] {
```
thehenrytsai: This method no longer makes sense in the new context... can we change it so it becomes:
```suggestion
  public static generateServiceEndpoints (count: number): DidServiceEndpointModel[] {
```

tests/json/didDocumentTemplate.json
```diff
@@ -22,15 +22,9 @@
   ],
   "service": [
     {
-      "id": "IdentityHub",
+      "id": "did:sidetree:value0",
```
thehenrytsai: Sanity: I am suspecting this file is no longer needed, let's try to remove it.

tests/core/DocumentComposer.spec.ts
```diff
@@ -3,8 +3,122 @@ import DocumentModel from '../../lib/core/versions/latest/models/DocumentModel';
 import ErrorCode from '../../lib/core/versions/latest/ErrorCode';
 import JasmineSidetreeErrorValidator from '../JasmineSidetreeErrorValidator';
 import SidetreeError from '../../lib/common/SidetreeError';
+import OperationGenerator from '../generators/OperationGenerator';
```
thehenrytsai: Nit^9000: sort.

lib/core/versions/latest/DocumentComposer.ts
```diff
@@ -197,25 +186,61 @@ export default class DocumentComposer {
...
 
-    if (patch.serviceType !== 'IdentityHub') {
-      throw new SidetreeError(ErrorCode.DocumentComposerPatchServiceTypeMissingOrUnknown);
+    for (const id of patch.serviceEndpointIds) {
+      if (typeof id !== 'string') {
+        throw new SidetreeError(ErrorCode.DocumentComposerPatchServiceEndpointIdsIdNotString);
+      }
+      if (id.length > 20) {
+        throw new SidetreeError(ErrorCode.DocumentComposerPatchServiceEndpointIdsIdTooLong);
+      }
```
thehenrytsai: Consider taking this out as a common method `validateServiceEndpointId()` to be used by two places, then you can also remove these error codes.

lib/core/versions/latest/DocumentComposer.ts
isaacJChen: Good catch, I always forget...

lib/core/versions/latest/DocumentComposer.ts
isaacJChen: Yes, it is a bug. Good catch.

Issue #278, opened 2019-08-21
Define sidetree core did context


OR13:
As we've seen with https://github.com/w3c-ccg/did-spec/pull/270

changes in the context can cause issues for client implementing JSON-LD (such as the univeral resolver).

We hacked around this issue in element by downgrading the context to the one used by BTCR (which is older, but supports publicKeyHex).

Unfortunately, there is no way to update existing DIDs to make them resolvable with the current protocol.

https://github.com/decentralized-identity/element/pull/68/files#diff-25e28c635b49bf7e5aa63a8a6ac95419R7

We need a way to change the context for a DID Document after it was created. This will also become more important as JSON-LD Support expands, and new key types are supported, such a PGP, or ZKPs.

OR13:
I propose an new update operation type to resolve this issue.

OR13:
https://github.com/w3c/did-core-registries/issues/19

OR13:
This is done.

PR #500, opened 2020-03-27
Define sidetree context

OR13:
Addresses: https://github.com/decentralized-identity/sidetree/issues/278


docs/spec/markdown/context.md
```diff
@@ -0,0 +1,55 @@
...
+}
+```
+
+### usage
+
+Deprecated. DO NOT USE.
+
+### publicKeyJwk
+
+A secp256k1 public key in JWK format. A JSON Web Key (JWK) is a JavaScript Object Notation (JSON) data structure that represents a cryptographic key. Read [RFC7517](https://tools.ietf.org/html/rfc7517).
```
troyronda: @OR13 We will also allow for other curves in our implementation.

Issue #395, opened 2020-02-10
Set per-op fee window large enough to smooth the per-op fee over a span of years


csuwildcat:
The desire is to smooth the base unit value variable the per-op fee is a proxy for over a long duration, such that fee increases/decreases would require steady movement in a trending direction over many months or years of time to realize a significant effect on the per-op fee.

PR #503, opened 2020-03-27
#395 Update the quantile calculator to smooth the fee curve

mudiali:
Added a new factor parameter to make sure that the current quantile does not deviate too much from the previous quantile value.

### DIFF
--- a/lib/bitcoin/BitcoinProcessor.ts
+++ b/lib/bitcoin/BitcoinProcessor.ts
@@ -107,6 +107,7 @@ export default class BitcoinProcessor {
     this.quantileCalculator = new SlidingWindowQuantileCalculator(BitcoinClient.convertBtcToSatoshis(1),
       ProtocolParameters.windowSizeInGroups,
       ProtocolParameters.quantileMeasure,
+      ProtocolParameters.quantileDeviationWindowFactor,
       mongoQuantileStore);
     this.transactionSampler = new ReservoirSampler(ProtocolParameters.sampleSizePerGroup);
 
--- a/lib/bitcoin/BitcoinProcessor.ts
+++ b/lib/bitcoin/BitcoinProcessor.ts
@@ -107,7 +107,7 @@ export default class BitcoinProcessor {
     this.quantileCalculator = new SlidingWindowQuantileCalculator(BitcoinClient.convertBtcToSatoshis(1),
       ProtocolParameters.windowSizeInGroups,
       ProtocolParameters.quantileMeasure,
-      ProtocolParameters.quantileDeviationWindowFactor,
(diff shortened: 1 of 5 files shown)


lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -107,6 +107,7 @@ export default class BitcoinProcessor {
     this.quantileCalculator = new SlidingWindowQuantileCalculator(BitcoinClient.convertBtcToSatoshis(1),
       ProtocolParameters.windowSizeInGroups,
       ProtocolParameters.quantileMeasure,
+      ProtocolParameters.quantileDeviationWindowFactor,
```
thehenrytsai: Would change factor to percentage be easier to comprehend? e.g.
`maxQuantileDeviationPercentage`, or `maxQuantileChangePercentage`, or `maxSubsequentQuantileChangePercentage`?

mudiali: I was thinking about that too but wanted to stick with Arvind's original suggestions (and also not do the math :) ...) but I agree and will update.

lib/bitcoin/fee/SlidingWindowQuantileCalculator.ts
```diff
@@ -30,12 +30,14 @@ export default class SlidingWindowQuantileCalculator {
...
    * The value to use as an input to the ValueApproximator.
    */
-  private readonly feeApproximate: number = 1.1;
+  private readonly feeApproximate: number = 1.414;
   /**
    * Size of a frequency vector; 1 + max normalized value
    */
   private frequencyVectorSize: number;
 
+  private maxQuantileDeviationPercentageValue: number;
```
thehenrytsai: I guess you took it out as a member variable for testing purposes.

lib/bitcoin/models/ProtocolParameters.ts
```diff
@@ -15,8 +15,8 @@ export default interface ProtocolParameters {
   /** Quantile measure we use for proof of fee; e.g., 0.5 would be the median */
   quantileMeasure: number;
 
-  /** Factor used to calculate how much the quantile values allowed to deviate from the previous value */
-  quantileDeviationWindowFactor: number;
+  /** The max amount that a quantile value are allowed to deviate from the previous value */
```
thehenrytsai:
```suggestion
  /** The max amount that a quantile value is allowed to deviate from the previous value */
```

lib/bitcoin/fee/SlidingWindowQuantileCalculator.ts
mudiali: That and I wanted to divide by 100 and save the value as member.

Issue #472, opened 2020-03-24
Remove # from being required in Key ID of public keys given in the operation requests


thehenrytsai:

thehenrytsai:
Fixed.

PR #508, opened 2020-03-28
#472 - Removed # from being required in IDs + restricted IDs to Base64URL char set

thehenrytsai:
1. Removed # from being required/allowed in IDs.
1. Added '#' to IDs during external DID document transform.
1. Added base64url character set check for IDs.


lib/core/versions/latest/DocumentComposer.ts
```diff
@@ -249,15 +248,28 @@ export default class DocumentComposer {
     }
   }
 
-  private static validateServiceEndpointId (serviceEndpointId: any) {
-    if (typeof serviceEndpointId !== 'string') {
-      throw new SidetreeError(ErrorCode.DocumentComposerPatchServiceEndpointIdNotString);
+  private static validateId (id: any) {
+    if (typeof id !== 'string') {
+      throw new SidetreeError(ErrorCode.DocumentComposerIdNotString);
```
mudiali: Nit: include the input in the error message? The tests will have to change to use the JasmieSidetreeErrorValidator class though so up to you if the change is big.

lib/core/versions/latest/DocumentComposer.ts
```diff
@@ -381,11 +389,19 @@ export default class DocumentComposer {
 
     didDocument.id = did;
 
-    // Only update the publickey if the array is present
+    // Only update `publickey` if the array is present
```
OR13: You could include jsonld.js as a dependency, and use `@base` to do this, for all the properties at once... `assertionMethod` (required by the VC Data Model), `authentication` (required by CHAPI), keyAgreementKey, (required by EDVs)...

This solution looks a little brittle to me, but it gets the job done for the current set of features that ION DIDs support.

lib/core/versions/latest/DocumentComposer.ts
thehenrytsai: Updated.

lib/core/versions/latest/DocumentComposer.ts
thehenrytsai: Created issue #512 to rework the transform logic.

Issue #474, opened 2020-03-24
Limit one writer to write only one transactions per block


mudiali:
Requirements:
1. This should occur before any other rate limiting schemes that we apply currently.
1. If a writer has multiple transactions written in a single block then we will only choose the first one.

PR #511, opened 2020-03-30
limit one transaction per writer per block #474

isaacJChen:


docs/protocol.md
```diff
@@ -182,11 +182,15 @@ Sidetree protocol defines the following mechanisms to enable scaling, while prev
 
 #### Rate limiting
    
-   To prevent spam attack causing transaction and operation stores to grow at an unhealthy rate, rate limiting is put in place. 
+   To prevent spam attack causing transaction and operation stores to grow at an unhealthy rate, 2 types of rate limiting is put in place. 
 
-   The current implementation caps the number of operations and transactions allowed to be observed and processed. The selection logic when the caps are exceeded is the following:
+   1. writer rate limiting
+   Each writer is allowed 1 transaction per transaction time. If a writer has more than 1 transaction, the one with the lowest transaction number is chosen to be considered.
```
thehenrytsai: To be clear, it should be "writer that require staking", smaller batch writers are not subject to this.
Also, going forward, when documentation needs to be updated:

1. we should stop updating this document, and file an issue with "documentation" tag and assign it to Daniel or Orie. and
1. Remove the relevant section in this doc.

I will let you decide if you want to it for this issue.

lib/core/versions/latest/TransactionSelector.ts
```diff
@@ -42,11 +42,21 @@ export default class TransactionSelector implements ITransactionSelector {
     const transactionsPriorityQueue = TransactionSelector.getTransactionPriorityQueue();
 
     const currentBlockHeight = transactions[0].transactionTime;
+    const writerToTransactionNumMap = new Map();
```
thehenrytsai: Nit: avoid abbreviation (Num in this method) unless really really bad.

lib/core/versions/latest/TransactionSelector.ts
```diff
@@ -71,8 +58,33 @@ export default class TransactionSelector implements ITransactionSelector {
...
+  private static validateTransactions (transactions: TransactionModel[], currentTransactionTime: number) {
+    for (const transaction of transactions) {
+      // expect all transactions to be in the same transaction time
+      if (transaction.transactionTime !== currentTransactionTime) {
+        throw new SidetreeError(ErrorCode.TransactionsNotInSameBlock, 'transaction must be in the same block to perform rate limiting, investigate and fix');
+      }
+    }
+  }
+
+  // if multiple transactions have the same writer, take the first one in the array and enqueue into trasnactionPriorityQueue
```
thehenrytsai: Nit: commenting style outside of method is with `/**`. Consider move '//' comments within the method.

PR #518, opened 2020-04-01
#517 Add data initialization for the quantile store

mudiali:
1. Based on conversation with @arvinda removing Math.round from the `ValueApproximator` and adding it to the `SlidingWindowQuantileCalculator`. Also, changing it to `Math.ceil` to be more accurate.
1. Added a new class which initializes the quantile data store with static values to ensure that the very first boot time is not long.


lib/bitcoin/fee/SlidingWindowQuantileStoreInitializer.ts
```diff
@@ -0,0 +1,109 @@
...
+        genesisBlockNumber,
+        ProtocolParameters.groupSizeInBlocks,
+        ProtocolParameters.windowSizeInGroups,
+        ProtocolParameters.sampleSizePerGroup,
+        initialQuantileValue,
+        mongoDbStore);
+  }
+
+  private async addDataIfNecessary (): Promise<boolean> {
+    const dbIsEmpty = (await this.mongoDbStore.getFirstGroupId()) === undefined;
```
mudiali, line 61: Note by @arvinda: what if the node crashes in the middle. Let's do more checks to make sure that we are in a complete and not partial state.

lib/bitcoin/fee/SlidingWindowQuantileCalculator.ts
```diff
@@ -177,7 +183,9 @@ export default class SlidingWindowQuantileCalculator {
    * Get the quantile as of a specific groupId.
    */
   public getQuantile (groupId: number): number | undefined {
-    return this.historicalQuantiles.get(groupId);
+    const quantile = this.historicalQuantiles.get(groupId);
+
+    return quantile ? Math.ceil(quantile) : undefined;
```
thehenrytsai, line 188: Sanity: quantile is expressed in satoshis right?

lib/bitcoin/fee/SlidingWindowQuantileStoreInitializer.ts
```diff
@@ -0,0 +1,109 @@
...
+
+    const dataInitializer = SlidingWindowQuantileStoreInitializer.createInstance(
+      genesisBlockNumber,
+      normalizedQuantile,
+      mongoDbStore);
+
+    await dataInitializer.addDataIfNecessary();
+  }
+
+  private static createInstance (
```
thehenrytsai, line 46: Nit: this method seems to be unnecessary since it is private.

lib/bitcoin/fee/SlidingWindowQuantileStoreInitializer.ts
```diff
@@ -0,0 +1,109 @@
...
+    // Figure out how far in the past do we need to go.
+    //  - If the genesis block belongs to group: X then the end-group is `X - 1`. The
+    //    genesis block will be processed normally and its group will be entered normally.
+    const groupOfGenesisBlock = Math.floor(this.genesisBlockNumber / this.groupSizeInBlocks);
+    const endGroupId = groupOfGenesisBlock - 1;
+
+    //  - Start is going back at least full window size (-2 === go a little further back just to be safe)
+    let startGroupId = endGroupId - this.windowSizeInGroup - 2;
+
+    console.info(`Genesis block: ${this.genesisBlockNumber}; Start group: ${startGroupId}; End group; ${endGroupId}`);
```
thehenrytsai, line 79: Nit: consider chronological order.
```suggestion
    console.info(`Genesis block: ${this.genesisBlockNumber}; Initialization start group: ${startGroupId}; Initialization end group: ${endGroupId}`);

```

lib/bitcoin/fee/SlidingWindowQuantileStoreInitializer.ts
```diff
@@ -0,0 +1,109 @@
...
+    const frequencyVector = new Array<number>(this.sampleSizePerGroup);
+    frequencyVector.fill(this.initialQuantileValue);
+
+    // Get the value to be actually saved in the DB.
+    const encodedFrequencyVector = RunLengthTransformer.encode(frequencyVector);
+
+    console.info(`Inserting Quantile: ${this.initialQuantileValue}; Frequency vector: ${encodedFrequencyVector}`);
+
+    // Now insert the values
+    while (startGroupId <= endGroupId) {
```
thehenrytsai, line 97: Style: Prefer not morphing the meaning of a variable. Consider creating a temp variable.

lib/bitcoin/fee/SlidingWindowQuantileCalculator.ts
mudiali: The `.round()` call was removed from the `ValueApproximator` but we want to keep the fraction values as now we have a window (.1%) of deviation allowed. Also the `ValueApproximator` does not look at this as satoshis but just some value. If anything either the `getQuantile` output (like here) or the caller of this function should round the values as the values are expressed as satoshis only at that level.

lib/bitcoin/fee/SlidingWindowQuantileStoreInitializer.ts
mudiali: This is really only to help with testing. The `initializeDatabaseIfEmpty` calls this function and I can mock the output of this function when testing the `initializeDatabaseIfEmpty` function.

lib/bitcoin/fee/SlidingWindowQuantileStoreInitializer.ts
mudiali: Fixed.

lib/bitcoin/fee/SlidingWindowQuantileStoreInitializer.ts
thehenrytsai: I see! Subtle, is it possible to add a comment to say this is for testing only? In case another smart ass like me comes along and thinks he knows better.

lib/bitcoin/fee/SlidingWindowQuantileCalculator.ts
thehenrytsai: Agree with you that should not put it in `ValueApproximator`, but then I don't like that there are two definitions or quantiles. One "fake ones" that `historicalQuantiles` stores, and "real ones" that `getQuantile` returns. Maybe we should make sure they both mean the same thing by computing the ceiling at write time in `calculateCurrentQuantile()`?

lib/bitcoin/fee/SlidingWindowQuantileStoreInitializer.ts
mudiali: Fair. Will add a comment for calrity.

lib/bitcoin/fee/SlidingWindowQuantileCalculator.ts
mudiali: The 'current' quantile is only allowed to deviate a limited from the previous vlaues. We use the `historicalQuantiles` to get the previous value so saving the rounded values in there will affect that comparison. Now off course, we can rearchitect to save the previous value separately in my mind, it is better that internally we use the 'non-rounded' quantiles and only expose the rounded value to the public callers. To me it is more confusing to save the rounded values in `historicalQuantiles` and then non-rounded ones at other places such as `valueApproximator` etc.

Issue #527, opened 2020-04-06
Fixed security vulnerabilities in dependencie


thehenrytsai:
Probably updating dependencies will do the job.

isaacJChen:
Done by the pr above.

PR #529, opened 2020-04-07
upgrade ipfs and spec-up

isaacJChen:
- Tested IPFS, no functional change
- spec-up version upgraded by @csuwildcat


isaacJChen:
#527

Issue #514, opened 2020-03-31
Move from HEX key representation to JWK


csuwildcat:
In order to move to a format natively and universally compliant with all key types allowed in DID Documents, we need to migrate from HEX representations of keys to JWK.

thehenrytsai:
Implemented.

PR #530, opened 2020-04-07
#514 - Moved from HEX key representation to JWK

thehenrytsai:


docs/protocol.md
```diff
@@ -612,7 +565,12 @@ HTTP/1.1 200 OK
     {
       "id": "key1",
       "type": "Secp256k1VerificationKey2018",
```
OR13, line 567: You probably don't want to use this... you probably want to use: https://github.com/w3c-ccg/lds-jws2020/blob/master/docs/example/didDoc.json#L18

because it covers all key types.

lib/core/versions/latest/util/Jwk.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    if (publicKey.crv !== 'secp256k1') {
+      throw new SidetreeError(ErrorCode.JwkEs256kMissingOrInvalidCrv);
+    }
+
+    if (typeof publicKey.x !== 'string') {
+      throw new SidetreeError(ErrorCode.JwkEs256kMissingOrInvalidTypeX);
+    }
+
+    if (typeof publicKey.x !== 'string') {
```
isaacJChen, line 59: this should be publickey.y, also a test should have caught this :p

docs/protocol.md
```diff
@@ -290,9 +290,7 @@ POST / HTTP/1.1
 ```json
 {
   "patchDataHash": "Hash of the patch data.",
-  "recoveryKey": {
-    "publicKeyHex": "A SECP256K1 public key expressed in compressed HEX format."
-  },
+  "recoveryKey": "A SECP256K1 public key expressed in compressed JWK format.",
```
isaacJChen, line 293: I thought we are abandoning `protocol.md`?

lib/core/versions/latest/util/Jwk.ts
OR13: agreed... I'd prefer to not see this kind of manual processing, and instead for us to rely on an audited popular library... such as: https://www.npmjs.com/package/jose

docs/protocol.md
OR13: yes, I also feel like we should not be updating that, and instead we should be updating the swagger yaml or the core spec...

lib/core/versions/latest/util/Jwk.ts
isaacJChen: I think this is fine because we want specific values for certain fields. @thehenrytsai is using the `jose` library but we still need to validate the keys we get from a document is accepted by the network (aka the curve it uses and the key type)

docs/protocol.md
thehenrytsai: We talked about having a separate API spec, as opposed to having this definition in the main core spec. The swagger yaml does not contain the Core service API yet, I filed #537 for moving definitions to the yaml and would like to address it separately, meanwhile updated this doc so that the information is not out of date.

lib/core/versions/latest/util/Jwk.ts
thehenrytsai: Unit test to ensure 100% code coverage will follow, wanted to unblock multiple JWK related issues to be worked on concurrently, so getting this PR out ASAP.

docs/protocol.md
thehenrytsai: Thanks Orie for bring this to our attention, I have pinged Daniel to further clarify and specify requirements on this in the Core spec, will implement the changes if any in a separate PR.

PR #543, opened 2020-04-08
Abstract the bitcoin wallet to support remote wallets

mudiali:


lib/bitcoin/BitcoinClient.ts
```diff
@@ -5,6 +5,8 @@ import BitcoinInputModel from './models/BitcoinInputModel';
 import BitcoinLockTransactionModel from './models/BitcoinLockTransactionModel';
 import BitcoinOutputModel from './models/BitcoinOutputModel';
 import BitcoinTransactionModel from './models/BitcoinTransactionModel';
+import BitcoinWallet from './BitcoinWallet';
```
OR13, line 8: someday we won't be bundling bitcoin dependencies like this... sigh...

lib/bitcoin/interfaces/IBitcoinWallet.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  /**
+   * Gets the public key associated with this wallet.
+   *
+   * @returns The public key associated with this wallet as Buffer.
+   */
+  getPublicKey (): Buffer;
+
+  /**
+   * Gets the address objecr associated with this wallet.
```
thehenrytsai, line 16:
```suggestion
   * Gets the address object associated with this wallet.

```

lib/bitcoin/interfaces/IBitcoinWallet.ts
```diff
@@ -0,0 +1,36 @@
...
+   */
+  getAddress (): Address;
+
+  /**
+   * Signs the specified transaction using the keys associated with this wallet.
+   *
+   * @param sidetreeTransaction The sidetree transaction to sign.
+   * @returns The signed transaction.
+   */
+  signTransaction (sidetreeTransaction: Transaction): Promise<Transaction>;
```
thehenrytsai, line 26: Nit: Maybe `signSidetreeTransaction` to disambiguate from bitcoin transactions.

lib/bitcoin/interfaces/IBitcoinWallet.ts
```diff
@@ -0,0 +1,36 @@
...
+ * Represents a bitcoin wallet.
+ */
+export default interface IBitcoinWallet {
+
+  /**
+   * Gets the public key associated with this wallet.
+   *
+   * @returns The public key associated with this wallet as Buffer.
+   */
+  getPublicKey (): Buffer;
```
thehenrytsai, line 13: Consider 










c



```suggestion
  getPublicKeyHex (): string;

```
Buffer is too generic without further specification on what the serialization format is, consider just use hex directly since it looks like that's what is used later on.

lib/bitcoin/interfaces/IBitcoinWallet.ts
mudiali: Yes it should be transaction. This is a leftover for when I had a different function name.

lib/bitcoin/interfaces/IBitcoinWallet.ts
mudiali: Good idea.

lib/bitcoin/BitcoinClient.ts
mudiali: :) I think we are ready for the mono repo change @OR13

OR13: don't joke about such things :) ... I am waiting for @thehenrytsai and @csuwildcat to tell me we are good to move to that structure...

lib/bitcoin/interfaces/IBitcoinWallet.ts
```diff
@@ -8,12 +8,12 @@ export default interface IBitcoinWallet {
   /**
    * Gets the public key associated with this wallet.
    *
-   * @returns The public key associated with this wallet as Buffer.
+   * @returns The public key associated with this wallet as a he string.
```
thehenrytsai, line 11:
```suggestion
   * @returns The public key associated with this wallet as a hex string.

```

lib/bitcoin/interfaces/IBitcoinWallet.ts
```diff
@@ -23,7 +23,7 @@ export default interface IBitcoinWallet {
    * @param sidetreeTransaction The sidetree transaction to sign.
```
thehenrytsai, line 23:
```suggestion
   * @param transaction The sidetree transaction to sign.

```

tests/core/DocumentComposer.spec.ts
```diff
@@ -113,7 +113,7 @@ describe('DocumentComposer', async () => {
         serviceEndpoints: [{
           id: 'someId',
           type: 'someType',
-          serviceEndpoint: 'someEndpoint',}]
+          serviceEndpoint: 'someEndpoint'}]
```
thehenrytsai, line 116: Thank you!

Issue #479, opened 2020-03-25
Add key usage to `DocumentModel` so we can denote keys that can perform updates


thehenrytsai:

csuwildcat:
The related spec info for this should now be available here: https://identity.foundation/sidetree/docs/spec/#add-public-keys

PR #544, opened 2020-04-08
add key usage #479

isaacJChen:
#479


lib/core/enums/PublicKeyUsage.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Sidetree public key usage.
+ */
+enum PublicKeyUsage {
```
OR13, line 4: You might consider aligning these with https://w3c-ccg.github.io/ld-proofs/#proof-purpose

or https://www.w3.org/TR/WebCryptoAPI/#cryptokey-interface-types

or maybe not using the term general term "usage".... for example, these are really sidetreeProtocolKeyUsage... these keys might have uses according to existing standards linked above, beyond just what sidetree is doing.

lib/core/versions/latest/DocumentComposer.ts
```diff
@@ -23,19 +24,35 @@ export default class DocumentComposer {
 
     const document = didState.document as DocumentModel;
 
-    // Only transform `publicKeys` to  `publickey` if the array is present.
-    let publicKeys;
+    // Only populate `publicKey` if general usage exists.
+    // Only populate `authentication` if auth usage exists.
```
OR13, line 28: This implies we are sorta supporting https://w3c-ccg.github.io/ld-proofs/#proof-purpose, but not being explicit about it... probably better to be explicit.

tests/generators/OperationGenerator.ts
```diff
@@ -84,12 +84,13 @@ export default class OperationGenerator {
...
    */
-  public static async generateKeyPair (id: string): Promise<[PublicKeyModel, JwkEs256k]> {
+  public static async generateKeyPair (id: string, usage?: string[]): Promise<[PublicKeyModel, JwkEs256k]> {
     const [publicKey, privateKey] = await Jwk.generateEs256kKeyPair();
     const publicKeyModel = {
       id,
       type: 'Secp256k1VerificationKey2018',
-      jwk: publicKey
+      jwk: publicKey,
+      usage: usage || ['ops', 'general', 'auth']
```
OR13, line 93: Likely that implementations that follow did core, and json-ld data model will not model things this way, but this is just one implementation.

OR13:
I have reservations around what we are doing to the did core data model, by skinning it like this...  as long as no normative language around this "usage" property is added to the spec, i'm fine with these things, but I will object to adding normative language that attempts to overload terminology already defined / in use by existing crypto standards...

tests/generators/OperationGenerator.ts
```diff
@@ -84,12 +84,13 @@ export default class OperationGenerator {
    * Mainly used for testing.
    * @returns [publicKey, privateKey]
    */
-  public static async generateKeyPair (id: string): Promise<[PublicKeyModel, JwkEs256k]> {
+  public static async generateKeyPair (id: string, usage?: string[]): Promise<[PublicKeyModel, JwkEs256k]> {
```
thehenrytsai, line 87: Should call out in comment above what happens if usage is not supplied.

thehenrytsai:
I think the `DocumentComposer` is missing a very important check in `validateOperationKey()` in that usage must be checked, else it defeats the purpose of `ops` usage type.

lib/core/enums/PublicKeyUsage.ts
thehenrytsai, line 4: Nit: can we rename to KeyUsage? Because it is not so much that it is usage of the "public key" itself. But not too insistent on it.

lib/core/versions/latest/DocumentComposer.ts
isaacJChen: I am following the spec Daniel created, don't know if that is sufficient.

PR #546, opened 2020-04-09
Move spec to a top-level folder

csuwildcat:
Spec should be in its own path, moving accordingly.

This will break existing links:

https://identity.foundation/sidetree/docs/spec/ -> https://identity.foundation/sidetree/spec


csuwildcat:
Instead of identity.foundation/sidetree/docs/spec, the new URL for the spec would be identity.foundation/sidetree/spec, which a couple folks voiced an opinion in favor of (and I like as well, because it's more terse).

Issue #515, opened 2020-03-31
Switch to JWS Compact format for signed data in operations


csuwildcat:
To better align with the more commonly used compact format (which almost all JWT implementations use), change from the flattened JSON format to the compact format for JWS representations in the operations sections of our CAS files.

OR13:
We need examples for all DID Operations, in compact JWS form.

This is a missing link between the HTTP API Spec and the Core Spec.

OR13:
We need to update the core spec and swagger... https://identity.foundation/sidetree/docs/swagger/

with the exact compact JWS that is submitted.

PR #553, opened 2020-04-11
#515 - Switched from flattened JWS to compact JWS for operation requests

thehenrytsai:
1. Switched from flattened JWS to compact JWS for operation requests
1. Disallowed `kid` in the JWS for recover and deactivate operations.
1. Increased code coverage.


lib/core/versions/latest/Encoder.ts
```diff
@@ -45,13 +53,22 @@ export default class Encoder {
...
+    const isBase64UrlString = Encoder.isBase64UrlString(input);
+    if (!isBase64UrlString) {
+      throw new SidetreeError(ErrorCode.EncoderValidateBase64UrlStringInputNotBase64UrlString, `Input '${input}' not a Base64URL string.`);
+    }
+  }
+
+  /**
+   * Tests if the given string is a Base64URL string.
+   */
+  public static isBase64UrlString (input: string): boolean {
```
isaacJChen, line 65: Why not just have a validation function that throws? I feel like that is how we've been handling these things in other classes.

lib/core/versions/latest/util/Jws.ts
```diff
@@ -128,15 +147,23 @@ export default class Jws {
...
+  public static parseCompactJws (compactJws: any, expectKidInHeader: boolean): Jws {
+    return new Jws(compactJws, expectKidInHeader);
+  }
+
+  /**
+   * Creates a compact JWS string using the given input. No string validation is performed.
    */
-  public static parse (input: any): Jws {
-    return new Jws(input);
+  public static createCompactJws (protectedHeader: string, payload: string, signature: string): string {
```
mudiali, line 166: Nit: looks like this function could be private? Understand if you want to keep it public to be symmetrical with `parseCompactJws`.

lib/core/versions/latest/Encoder.ts
```diff
@@ -25,12 +26,19 @@ export default class Encoder {
...
    */
   public static decodeAsString (encodedContent: string): string {
-    Encoder.validateBase64UrlString(encodedContent);
+    return Encoder.decodeBase64UrlAsString(encodedContent);
+  }
+
+  /**
+   * Decodes the given Base64URL string into the original string.
+   */
+  public static decodeBase64UrlAsString (input: string): string {
```
mudiali, line 38: Do we need this new function? The `decodeAsString` can be refactored to this as well right?

lib/core/versions/latest/Encoder.ts
thehenrytsai: You are right there is already `validateBase64UrlString()` right above it which throws a generic error. I wanted in this case to throw a more specific error in `DocumenComposer` hence adding a generic method for this. I can remove it you feel like it is not adding much value.

lib/core/versions/latest/util/Jws.ts
thehenrytsai: It's public so tests can use it (to create invalid data mostly).

lib/core/versions/latest/Encoder.ts
thehenrytsai: The intention here is that this function will ALWAYS decode using Base64URL even if the default encoding scheme for the Encoder changes in future versions, to avoid bug creeping into decoding of things that's not tied directly to Sidetree spec, in this case, we are decoding JWS header, which is ALWAYS going to be base64URL.

lib/core/versions/latest/Encoder.ts
isaacJChen: I'm indifferent, just want to make sure we follow the same pattern as much as possible throughout the codebase, no need to change if there's a reason for it.

PR #555, opened 2020-04-13
fix ipfs

isaacJChen:


lib/ipfs/RequestHandler.ts
```diff
@@ -56,7 +56,7 @@ export default class RequestHandler {
    * Handles read request
    * @param base64urlEncodedMultihash Content Identifier Hash.
    */
-  public async handleFetchRequest (base64urlEncodedMultihash: string, maxSizeInBytes?: number): Promise<ResponseModel> {
+  public async handleFetchRequest (base64urlEncodedMultihash: string, maxSizeInBytes?: number, retryCount?: number): Promise<ResponseModel> {
```
thehenrytsai, line 59: Design: Not a fan of the recursive usage here. Retry is an internal implementation detail, can we hide it internally, like using a loop or something?

lib/ipfs/IpfsStorage.ts
```diff
@@ -44,10 +40,29 @@ export default class IpfsStorage {
...
+      repo: repo !== undefined ? repo : localRepoName
+    };
+    const node = await IPFS.create(options);
+    return node;
+  }
+
+  /**
+   * restarts the IPFS node
+   */
+  public async restart () {
```
thehenrytsai, line 60: I am a bit uneasy that restart gets called by the handlers. Say 50 downloads/calls happens (which can happen, since these calls tend to come in batches within a block of transaction), then it seems `restart()` thus `node.stop()` can happen 50 times. An alternative design could be that we use a "monitor" that checks IPFS node state every 1 minute and only it restarts the node.

isaacJChen: Okay, I like this idea.

lib/ipfs/IpfsStorage.ts
```diff
@@ -44,8 +42,48 @@ export default class IpfsStorage {
...
+    const localRepoName = 'sidetree-ipfs';
+    const options = {
+      repo: repo !== undefined ? repo : localRepoName
+    };
+    const node = await IPFS.create(options);
+    return node;
+  }
+
+  /**
+   * performs health check once every minute
```
thehenrytsai, line 62: Nit: Get rid off reference to "every minute":
```suggestion
   * Start periodic health check.
```

lib/ipfs/IpfsStorage.ts
```diff
@@ -44,8 +42,48 @@ export default class IpfsStorage {
...
+  private async healthCheck () {
+    console.log('performing health check');
+    if (!this.healthy) {
+      console.log('Unhealthy, restarting IPFS node...');
+      await this.restart();
+      this.healthy = true;
+    } else {
+      console.log('healthy');
+    }
+    setTimeout(async () => this.healthCheck(), this.healthCheckInternalInSeconds * 1000);
```
thehenrytsai, line 77: Pattern: consider putting this whole block in try catch-catch, and call `setTimeout()` in `finally` so the monitor never dies.

lib/ipfs/IpfsStorage.ts
```diff
@@ -23,13 +26,8 @@ export default class IpfsStorage {
...
     }
-
-    const localRepoName = 'sidetree-ipfs';
-    const options = {
-      repo: repo !== undefined ? repo : localRepoName
-    };
-    const node = await IPFS.create(options);
-    IpfsStorage.ipfsStorageSingleton = new IpfsStorage(node);
+    const node = await IpfsStorage.getNode(repo);
+    IpfsStorage.ipfsStorageSingleton = new IpfsStorage(node, repo);
```
thehenrytsai, line 30: Discussion: I find this initialization a bit confusing (the relationship between IpfsStorage, node, and repo) every time I come back to it after a month or so.

Do you think it would be cleaner if:
1. We only call `getNode()` internally (probably by monitor). and
1. Pass only repo into constructor.
1. Set `healthy` to false in  constructor. (and optionally rename `healthy` to perhaps `reinitializeRequired`)

lib/ipfs/IpfsStorage.ts
```diff
@@ -58,14 +96,20 @@ export default class IpfsStorage {
    *          The result `code` is set to `FetchResultCode.NotAFile` if the content being downloaded is not a file (e.g. a directory).
    */
   public async read (hash: string, maxSizeInBytes: number): Promise<FetchResult> {
-    const fetchResult = await this.fetchContent(hash, maxSizeInBytes);
+    try {
```
thehenrytsai, line 99: We probably should also add the same try-catch pattern for `write()`.

PR #556, opened 2020-04-13
#534 Allow bitcoinwallet to be injected via constructor

mudiali:
This will allow the node admins to pass in their own implementation of a bitcoin wallet if they want.


lib/bitcoin/BitcoinClient.ts
```diff
@@ -42,12 +41,21 @@ export default class BitcoinClient {
...
-    bitcoinWalletImportString: string,
+    bitcoinWalletImportString: string | undefined,
+    bitcoinWallet: IBitcoinWallet | undefined,
     private requestTimeout: number,
     private requestMaxRetries: number,
     private sidetreeTransactionFeeMarkupPercentage: number) {
 
-    this.bitcoinWallet = new BitcoinWallet(bitcoinWalletImportString);
+    if (bitcoinWalletImportString && bitcoinWallet) {
+      throw new SidetreeError(ErrorCode.BitcoinProcessorWalletAndImportStringSpecified, 'Both bitcoin wallet and import string are not allowed.');
```
isaacJChen, line 51: Nit: This error message is confusing to read for me. I think it will be clearer if we say something like `can only have either bitcoin wallet or import string configured, having both at the in the config is not allowed.` The message you used sounded like both are not allowed like "Both Ali and Isaac are not allowed to be here".

mudiali: Good point. Will update.

lib/bitcoin/BitcoinClient.ts
```diff
@@ -42,12 +41,21 @@ export default class BitcoinClient {
     private bitcoinPeerUri: string,
     bitcoinRpcUsername: string | undefined,
     bitcoinRpcPassword: string | undefined,
-    bitcoinWalletImportString: string,
+    bitcoinWalletImportString: string | undefined,
+    bitcoinWallet: IBitcoinWallet | undefined,
```
thehenrytsai, line 45: What about:
```suggestion
    bitcoinWalletOrImportString: IBitcoinWallet | string,
```

mudiali: Oh my!! I love javascript :) Let me try and will update.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -3,6 +3,7 @@ import BitcoinClient from './BitcoinClient';
 import BitcoinTransactionModel from './models/BitcoinTransactionModel';
 import ErrorCode from './ErrorCode';
 import IBitcoinConfig from './IBitcoinConfig';
+import IBitcoinWallet from './interfaces/IBitcoinWallet';
```
thehenrytsai, line 6: Would makes sense to export this interface explicitly in the exports in `index.ts, it's fine if you want to do it later.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -94,7 +95,7 @@ export default class BitcoinProcessor {
   /** at least 10 blocks per page unless reaching the last block */
   private static readonly pageSizeInBlocks = 10;
 
-  public constructor (config: IBitcoinConfig) {
+  public constructor (config: IBitcoinConfig, bitcoinWallet?: IBitcoinWallet) {
```
thehenrytsai, line 98: Optional: if you took the earlier suggestion, you might want to make the signature the same here, and remove it from IBitcoinConfig.
```suggestion
  public constructor (config: IBitcoinConfig, bitcoinWalletOrImportString: IBitcoinWallet | string) {
```

lib/bitcoin/BitcoinClient.ts
mudiali: Error no longer needed after refactor.

lib/bitcoin/BitcoinProcessor.ts
mudiali: Updated to pass in the bitcoinwallet as part of the config.

Issue #571, opened 2020-04-16
Update REST API spec with DID Resolution HTTPS binding


troyronda:
For DID result, I assume the following is an example Sidetree [DID Resolution Result](https://w3c-ccg.github.io/did-resolution/#did-resolution-result) structure:

```json
{
    "@context": "https://www.w3.org/ns/did-resolution/v1",
    "didDocument": DID_DOCUMENT_OBJECT,
    "methodMetadata": {
        "operationPublicKeys": [OPERATION_PUBLIC_KEY_OBJECT, ...],
        "recoveryKey": RECOVERY_PUBLIC_KEY_OBJECT,
        "published": boolean
    }
}
```

Notes:
- The DID Resolution spec uses a response `Content-Type` value set to `application/did+ld+json`.
- Are we specifying that `operationPublicKeys`, `recoveryKey` MUST always be returned?
  - If not, when should they be returned?
- Are we specifying that `published` MUST always be returned?
  - If not, what is the assumed default?
- Are we supporting `If the input DID exists and the result is a service endpoint URL` resulting in `HTTP response status code MUST be 303`?
  - See step 5 of [HTTPS Binding](https://w3c-ccg.github.io/did-resolution/#bindings-https)
  - Would we do so for unpublished DIDs?
- Also see current [reference implementation code](https://github.com/decentralized-identity/sidetree/blob/018bc7cba3698e977fdd657b161d1eef4fba6e32/lib/core/versions/latest/DocumentComposer.ts#L89-L99)

troyronda:
Fixed REST API description to return DID Resolution Result in #573. There are potential followups listed in the issue description above.

csuwildcat:
Has this progressed since this last comment? What's outstanding?

PR #573, opened 2020-04-17
Reference DID Resolution spec in REST API

troyronda:
- references the [DID Resolution HTTP(S) binding](https://w3c-ccg.github.io/did-resolution/#bindings-https).
- notes that the create operation also returns the DID resolution result.
- notes that method metadata also includes `operationPublicKeys` and `recoveryKey`.

Followups:
- Should we document additional details from DID Resolution binding. Examples:
  -  The server MUST include the `Content-Type` header, with the value set to `application/did+ld+json`.
- Questions posed in https://github.com/decentralized-identity/sidetree/issues/571
- Should `published` (eventually) be moved to resolverMetadata?

### DIFF
--- a/spec/markdown/rest-api.md
+++ b/spec/markdown/rest-api.md
@@ -4,16 +4,24 @@ The following sections define the Sidetree resolution and operations endpoints. 
 
 ### Sidetree Resolution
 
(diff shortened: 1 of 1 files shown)


OR13:
killing it @troyronda

Issue #464, opened 2020-03-23
Migrate CI to Github Actions


OR13:
They are much better, and we already use them for spec deployment.

tplooker:
Done, closing

PR #585, opened 2020-04-21
#464 - Migrate CI to Github Actions

kg0r0:
This is a pull request for this issue https://github.com/decentralized-identity/sidetree/issues/464

I added CI configuration for GitHub Actions.
Please close this pull request if you do not want it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [10.15.3]
+        node-version: [10.15.3, 12]
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,18 @@
+name: CI
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [10.15.3]
```
thehenrytsai, line 10: Can we raise the node version to 12?

kg0r0: I added the setting of version 12, so it runs on both versions.

If you only need version 12, can you let me know?

Issue #579, opened 2020-04-19
Swagger API MUST use snake case.


troyronda:
For consistency with the Sidetree protocol spec, it would make sense to use underscores rather than snake-case.

OR13:
use underscore everywhere, all properties must use "_" not "-".

move swagger next to api document.

PR #596, opened 2020-04-22
Update swagger API to use underscores

troyronda:
Updated JSON properties to use underscores.

Closes: #579

### DIFF
--- a/docs/swagger/api.yml
+++ b/docs/swagger/api.yml
@@ -19,7 +19,7 @@ paths:
     post:
       tags:
         - Sidetree
-      summary: Writes an Opertaion to Sidetree.
+      summary: Writes an Operation to Sidetree.
       operationId: writeSidetreeOperation
       parameters:
         - in: body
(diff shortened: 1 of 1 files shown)


troyronda:
Possible followup: what is the convention to use for query parameters?
Examples:
- `max-size`
- `transaction-time-hash`

OR13:
@thehenrytsai @troyronda @kdenhartog I've always been confused why we use kebab-case for query params... I prefer understore (snake_case) for consistency.

OR13:
@troyronda  I recommend you switch the params to snake_case, and wait for someone to object.

troyronda:
@OR13 done

kdenhartog:
@OR13 number 4 in consistency is key section here gives a reason why. https://restfulapi.net/resource-naming/

With that said, we typically use kebab-case but I'm not going to object if api endpoints use snake case.

OR13:
@kdenhartog great reference... I was assuming that this only applied to query strings, but seems like you are thinking that the path segments should also be considered... 

Can you open an issue and cite that resource, I feel like we should base our API on a styleguide... and I would be fine noting that paths should be kebab case, query strings and json objects should be snake-case.

kdenhartog:
> snake-case.

snake_case. FTFY 😎 

Also, will do.

csuwildcat:
To clarify this:

1. Underscores should be used in JSON payload keys
2. Dashes should be used for path segments and URL Parameters
3. If we ever specify internal APIs in IDL, functions should be camel cased, and variables likely underscored (less rigidity on this last item)

OR13:
^ move discussion to github issue. don't discuss things on a merged PR.

PR #600, opened 2020-04-23
Clarify reference to DID Suffix of Create operations

EzequielPostan:
We were referring to the DID Suffix of operations of any type. However,
Create operations do not have a DID Suffix (while Update, Deactivate and
Recover operations do). We defined the DID Suffix of Create operations
as the suffix that their suffix_data field produce


troyronda:
note: we now have two spec-related directories: `spec/` and` api/`

Issue #603, opened 2020-04-23
Method name should not require a network ID in config


thehenrytsai:
`sidetree:test` works, but `sidetree` does not.

thehenrytsai:
Fixed.

PR #604, opened 2020-04-23
#603 - Fixed method name should not require a network ID in config

thehenrytsai:


tests/core/Did.spec.ts
```diff
@@ -5,11 +5,21 @@ import OperationGenerator from '../generators/OperationGenerator';
...
+        () => (Did as any).getEncodedCreateRequestFromDidString('badURL', ['anyMethodName']),
+        ErrorCode.DidInvalidDidString
+      );
+    });
+  });
+
   describe('create()', async () => {
     it('should create a short-form DID succssefully.', async () => {
-      const expectedDidMethodName = 'did:sidetree:';
+      const expectedDidMethodName = 'sidetree';
```
OR13, line 20: do you have a test for `sidetree:testnet` ?

lib/core/versions/latest/Did.ts
```diff
@@ -63,7 +64,7 @@ export default class Did {
    */
   public static async create (didString: string, didMethodName: string): Promise<Did> {
     const methodNameParts = didMethodName.split(':');
-    if (methodNameParts.length < 2) {
+    if (methodNameParts.length > 2) {
```
isaacJChen, line 67: What if method name is `sidetree:some:thing`? then we don't accept that?

thehenrytsai: I'd say we don't, unless we find a need for it, for now.

OR13: Thats not valid, please review did core: https://w3c.github.io/did-core/#did-syntax

OR13: Apparently, this case is allowed... https://github.com/transmute-industries/vc.transmute.world/blob/master/src/services/unlockedDIDs/did:v1:test:nym:z6MkhdmzFu659ZJ4XKj31vtEDmjvsi5yDZG5L7Caz63oP39k.json

tests/core/Did.spec.ts
thehenrytsai: Added a testnet DID test.

lib/core/versions/latest/Did.ts
```diff
@@ -29,12 +30,13 @@ export default class Did {
    * @param didMethodName The expected DID method given in the DID string. The method throws SidetreeError if mismatch.
    */
   private constructor (did: string, didMethodName: string) {
-    if (!did.startsWith(didMethodName)) {
+    this.didMethodName = didMethodName;
```
logangirvin, line 33:
```typescript
(did: string, public didMethodName: string) {
```

lib/core/versions/latest/Did.ts
```diff
@@ -44,17 +46,17 @@ export default class Did {
     }
 
     if (this.isShortForm) {
-      this.uniqueSuffix = did.substring(didMethodName.length);
+      this.uniqueSuffix = did.substring(didPrefix.length);
     } else {
       // This is long-form.
-      this.uniqueSuffix = did.substring(didMethodName.length, indexOfQuestionMarkChar);
+      this.uniqueSuffix = did.substring(didPrefix.length, indexOfQuestionMarkChar);
```
logangirvin, line 52: Couldn't these be combined with 43/45? It makes substring'ing on indexOfQuestionMarkChar >=0 at compile time.

lib/core/versions/latest/Did.ts
```diff
@@ -94,35 +95,72 @@ export default class Did {
...
+    if (firstIndexOfDot === -1) {
+      throw new SidetreeError(ErrorCode.DidInitialStateValueContainsNoDot);
+    }
+
+    const lastIndexOfDot = initialState.lastIndexOf('.');
+    if (lastIndexOfDot !== firstIndexOfDot) {
+      throw new SidetreeError(ErrorCode.DidInitialStateValueContainsMoreThanOneDot);
+    }
+
+    if (firstIndexOfDot === (initialState.length - 1)) {
```
logangirvin, line 149: No check for `firstIndexOfDot === 0` ?

tests/core/Did.spec.ts
```diff
@@ -105,4 +118,49 @@ describe('DID', async () => {
...
+      const initialState = 'abc.123.'; // Intentionally having more than 1 '.'
+
+      await JasmineSidetreeErrorValidator.expectSidetreeErrorToBeThrownAsync(
+        async () => (Did as any).constructCreateOperationFromInitialState(initialState), ErrorCode.DidInitialStateValueContainsMoreThanOneDot
+      );
+      done();
+    });
+
+    it('should throw if the given initial state string has more than one dot.', async (done) => {
+      const initialState = 'abc.'; // Intentionally not having two parts after splitting by '.'
```
logangirvin, line 142: consider also testing `.abc`

lib/core/versions/latest/Did.ts
```diff
@@ -63,24 +65,23 @@ export default class Did {
    */
   public static async create (didString: string, didMethodName: string): Promise<Did> {
     const methodNameParts = didMethodName.split(':');
-    if (methodNameParts.length < 2) {
+    if (methodNameParts.length > 2) {
```
isaacJChen, line 68: Thought we want to remove this check?

lib/core/versions/latest/Did.ts
thehenrytsai: Done.

tests/core/Did.spec.ts
thehenrytsai: Done.

lib/core/versions/latest/Did.ts
thehenrytsai: Alright, I'm getting rid off it... I got rid of it anyways in the original change.

lib/core/versions/latest/Did.ts
thehenrytsai: Not sure if the ask is to combine the two if statements. If so, I'd prefer to keep them separate for readability for dummies.

lib/core/versions/latest/Did.ts
thehenrytsai: Would do it if there were no other props or other props are declared and instantiated the same way. In this case prefer to keep the declaration of the `didMethodName` property explicit together with  other props for consistency.

Issue #583, opened 2020-04-20
Use relative path for `authentication` entries when referencing keys in `publicKey`


thehenrytsai:
Use relative path for `authentication` entries when referencing keys in `publicKey`, or else the reference value length will be large than replicating the key itself:

```javascript
"authentication": [
[...]

PR #611, opened 2020-04-24
Revert the #583 revert

thehenrytsai:


lib/core/versions/latest/Did.ts
```diff
@@ -107,7 +108,7 @@ export default class Did {
     let initialStateValue;
 
     // Verify that `-<method-name>-initial-state` is the one and only parameter.
-    didStringUrl.searchParams.forEach((value, key) => {
+    for (const [key, value] of didStringUrl.searchParams) {
```
isaacJChen, line 111: Cool didn't know you can do this!

PR #617, opened 2020-04-27
make ipfs creation sync

isaacJChen:


lib/ipfs/RequestHandler.ts
```diff
@@ -16,28 +16,16 @@ export default class RequestHandler {
...
   public ipfsStorage: IpfsStorage;
-
   private serviceInfo: ServiceInfo;
 
-  /**
-   * create an instance of request handler
-   * @param fetchTimeoutInSeconds Timeout for fetch request. Fetch request will return `not-found` when timed-out.
-   * @param repo Optional IPFS datastore implementation.
-   */
-  public static async create (fetchTimeoutInSeconds: number, repo?: any) {
```
thehenrytsai, line 27: Optional: Generally I avoid invoking constructor directly and prefer factory method if possible, so you might want to keep the `create` method. The reasoning can be found on the internet, but in summary it's to anticipate changes. e.g.
If for whatever reason, the initialization needs to become async again. Or if we want control how many request handlers that gets created etc.

Issue #615, opened 2020-04-26
publicKeyJwk in overlapping contexts


troyronda:
I assume should either be made the same `sec:publicKeyJwk` as in `https://www.w3.org/ns/did/v1`  or removed from `context-v1.jsonld`.

In https://www.w3.org/ns/did/v1:

```
    "publicKeyJwk": {
      "@id": "sec:publicKeyJwk",
      "@type": "@json"
    },
```

In https://identity.foundation/sidetree/context-v1.jsonld:
```
    "publicKeyJwk": {
      "@id": "sidetree:publicKeyJwk",
      "@type": "@json"
    }
```

troyronda:
@OR13 what do you think we should do with this one?

OR13:
we should use did core... we are not allowed to overwrite did core actually.

The spec can define the term again, but it should point to the normative defintion at did core.

csuwildcat:
No idea what this is about, but can we close it?

OR13:
@csuwildcat when a term in a did document is not defined in a context, it creates an invalid did document... so sidetree did documents that contained "usage" were not valid, and would explode when used with JSON-LD, until we defined "usage" in a custom context.

Sidetree supports this custom context today: https://github.com/decentralized-identity/sidetree/blob/master/contexts/sidetree-v0.0.0.jsonld#L7

You can see that we are defining publicKeyHex, because it's not defined in did core... if it were, we would remove it from our context, as we have done for publicKeyJwk.


I am closing this issue now, because it is resolved.

PR #646, opened 2020-05-06
update context document to remove publicKeyJwk and add JsonWebKey2020

### LINKED ISSUE #612: JwsVerificationKey2020 not in the contexts
The contexts listed in https://identity.foundation/sidetree/spec/#context do not contain `JwsVerificationKey2020`

kdenhartog:
Signed-off-by: Kyle Den Hartog <kyle.denhartog@mattr.global>

This should close #612 and #615

@OR13 can you double check to make sure I did this right?

### DIFF
--- a/contexts/sidetree-v0.0.0.jsonld
+++ b/contexts/sidetree-v0.0.0.jsonld
@@ -16,6 +16,9 @@
       "@container": "@set"
     },
     "usage": "sidetree:usage",
-    "JsonWebKey2020": "ldssk:JsonWebKey2020"
+    "JsonWebKey2020": {
+      "@id":"ldssk:JsonWebKey2020",
+      "@type": "@json"
+    }
   }
 }
--- a/contexts/sidetree-v0.0.0.jsonld
+++ b/contexts/sidetree-v0.0.0.jsonld
@@ -3,11 +3,8 @@
     "@version": 1.1,
     "did-core-registries": "https://w3c.github.io/did-core-registries/#",
     "sidetree": "https://identity.foundation/sidetree/spec/#",
+    "ldssk": "https://w3c-ccg.github.io/lds-jws2020/contexts/#",
     "publicKeyHex": "did-core-registries:publicKeyHex",
-    "publicKeyJwk": {
-      "@id": "sidetree:publicKeyJwk",
-      "@type": "@json"
-    },
     "recovery": {
       "@id": "sidetree:recovery",
       "@type": "@id",
@@ -18,6 +15,7 @@
       "@type": "@id",
       "@container": "@set"
     },
-    "usage": "sidetree:usage"
+    "usage": "sidetree:usage",
+    "JsonWebKey2020": "ldssk:JsonWebKey2020"
   }
 }


context-v1.jsonld
```diff
@@ -16,6 +12,11 @@
       "@id": "sidetree:operation",
       "@type": "@id",
       "@container": "@set"
+    },
+    "JsonWebKey2020": {
```
OR13, line 16: it should look like this: https://github.com/w3c/did-core-registries/blob/master/contexts/did-v0.11.jsonld#L14

kdenhartog:
This isn't updated anywhere else that I can reference it though. Are you thinking I use something like this?
```
{
 "jwk":"https://w3c-ccg.github.io/lds-jws2020/contexts/lds-jws2020-v0.0.jsonld",
 "JsonWebKey2020":"jwk:JsonWebKey2020",
 ...
}
```

OR13:
@kdenhartog thats not how JSON-LD works...

A term is defined in a context:

https://github.com/w3c-ccg/security-vocab/blob/master/contexts/security-v2.jsonld#L13

A Vocabulary for the term definition is provided:

https://github.com/w3c-ccg/security-vocab/blob/master/contexts/security-v1.jsonld#L7

The term is defined at the URL:

https://w3c-ccg.github.io/security-vocab/#Ed25519VerificationKey2018

When you can't use `sec`, you must define your own vocabulary, as you did for BBS+:

https://github.com/mattrglobal/jsonld-signatures-bbs-spec/blob/master/contexts/v0.1.jsonld#L6
https://github.com/mattrglobal/jsonld-signatures-bbs-spec/blob/master/contexts/v0.1.jsonld#L10

Here is the current term definition, which must be preserved with a deprecation warning:

https://w3c-ccg.github.io/lds-jws2020/contexts/#JwsVerificationKey2020

The new one would be: https://w3c-ccg.github.io/lds-jws2020/contexts/#JsonWebKey2020

Its definition in here: https://w3c-ccg.github.io/lds-jws2020/contexts/lds-jws2020-v0.0.jsonld

would look like this:

```
"JsonWebKey2020": "ldssk:JsonWebKey2020",
```

OR13:
ahh, I see you are also trying to implement the nesting feature here... perhapse it would be better to do that here first: https://github.com/w3c-ccg/lds-jws2020

show that tests still pass, and then change the structure of the context definitions.

kdenhartog:
I haven't gotten back to cleaning up this spec yet. Will get around to it within the week (aiming for before our meeting this week).

kdenhartog:
@OR13 I updated this finally, but was a bit confused by one of your comments. This may still be unaddressed.

> "the current term definition, which must be preserved with a deprecation warning" 

To me this implies I should be adding `JsonVerificationKey2020`. I thought we were redefining it so that we didn't have deprecate at this layer and could instead only include `JsonWebKey2020`. Did I misunderstand that?

contexts/sidetree-v0.0.0.jsonld
```diff
@@ -18,6 +15,7 @@
       "@type": "@id",
       "@container": "@set"
     },
-    "usage": "sidetree:usage"
+    "usage": "sidetree:usage",
+    "JsonWebKey2020": "ldssk:JsonWebKey2020"
```
OR13, line 19: Can you add `"@type": "@json"z`

OR13: if we don't do this, you will errors about `kty` and `kid` not being defined

kdenhartog:
@OR13 fixed

Issue #590, opened 2020-04-21
Namespace authority


kdenhartog:
naming of the uri’s (e.g. ion vs elem vs etc). We lack any sort of description around how methods should be named. This will be important as we have competing interests such as multiple sidetree methods running on ethereum. Not encouraging this, but understanding this is a necessary countermeasure.

How does this assist with the forking problem?

csuwildcat:
Can you clarify a bit about what you are thinking, and how you would approach either language or some mechanism to do this?

kdenhartog:
Right now, we implicitly build the naming of the method using `did:methodName:network:....` If we want to continue with this path we should make it explicit. Maybe, we want to go the route of `did:sidetree:ion:test`. In any case I think we should be making recommendations around how the front part of the did gets decided and may want to make a note about how it's used to handle forks. For example, maybe the way we indicate a fork is instead of using "main" we use last 5 characters of the first block of a fork prepended to the block number. Handling forking I think can be noted and not defined in specv1, but we should at the least note it.

This isn't anything that's going to be difficult to define, but we should put a bit of thought in it ahead of time I would think.

OR13:
@kdenhartog can you add a "note" about forking... ETH / ETC...

kdenhartog:
Pending close once PR is merged.

OR13:
This was merged, closing.

PR #647, opened 2020-05-06
add warning about forking of decentralized ledger system

kdenhartog:
Signed-off-by: Kyle Den Hartog <kyle.denhartog@mattr.global>

Addresses issue #590

### DIFF
--- a/docs/spec/method-versioning.md
+++ b/docs/spec/method-versioning.md
@@ -21,3 +21,6 @@ An ELEM DID on the Ethereum Ropsten testnet is defined as follows:
 
 `did:elem:ropsten:EiD0x0JeWXQbVIpBpyeyF5FDdZN1U7enAfHnd13Qk_CYpQ`
 
+:::warning Ledger Forking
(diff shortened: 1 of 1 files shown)


OR13:
@kdenhartog make sure to never commit the html, only the markdown files...

csuwildcat:
Kyle, if you just put this in the same place in the MD file, it will render the warning:

```markdown
:::warning Ledger Forking
Implementers should be aware that if the underlying decentralized ledger system were to fork, the identifiers will also be forked. In this case, the a new identifier must be created either through an indication at the network layer or with a completely new method name to identify the decentralized identifiers of the forked network.
:::
```

kdenhartog:
fixed, thanks for the help.

OR13:
@kdenhartog we changed the structure on you, sorry :)

kdenhartog:
> 
> 
> @kdenhartog we changed the structure on you, sorry :)

All good, should I do anything else other than resolve the merge conflict? Now that I know spec-up is about editing markdown files I'll stick with that. I personally enjoy writing in markdown over the other formats most specs require. IETF with XML was probably the least fun, but I think I just picked the wrong tool in that case cause I've seen markdown tools for producing IETF drafts as well.

Issue #659, opened 2020-05-12
JSON Web Signature Deactivate


troyronda:
In https://identity.foundation/sidetree/spec/#verifying

> A Recover Operation is only valid when signed by a key listed in recovery. Other operations are only valid when signed with keys listed in operations.

Should also mention Deactivate Operation using a key listed in recovery.

PR #665, opened 2020-05-14
spec: mention deactivate operation in verifying JWS

troyronda:
Closes: #659

### DIFF
--- a/docs/spec/json-web-signatures.md
+++ b/docs/spec/json-web-signatures.md
@@ -25,7 +25,7 @@ If `kid` is a DID URL, the verification key MAY be resolved as follows:
 
 1. Resolve the `kid` value using a resolver.
 2. iterate the `operations` and `recovery` verificationMethods, until a verificationMethod with `id` equal to `kid` is found.
-3. A [Recover Operation](https://identity.foundation/sidetree/spec/#recover) is only valid when signed by a key listed in `recovery`. Other operations are only valid when signed with keys listed in `operations`. If the key is not present in the expected collection, verification fails, and the operation is considered invalid. If the key is present in the expected collection proceed.
+3. A [Recover Operation](https://identity.foundation/sidetree/spec/#recover) or a [Deactivate Operation](https://identity.foundation/sidetree/spec/#deactivate) are only valid when signed by a key listed in `recovery`. Other operations are only valid when signed with keys listed in `operations`. If the key is not present in the expected collection, verification fails, and the operation is considered invalid. If the key is present in the expected collection proceed.
 3. Convert the discovered verificationMethod to JWK if necessary.
 4. Perform [JWS Verification](https://tools.ietf.org/html/rfc7515#section-5.2) using the JWK.
 5. The operation is considered valid if key material was present in the correct collection and the signature is valid.


docs/spec/json-web-signatures.md
```diff
@@ -25,7 +25,7 @@ If `kid` is a DID URL, the verification key MAY be resolved as follows:
 
 1. Resolve the `kid` value using a resolver.
 2. iterate the `operations` and `recovery` verificationMethods, until a verificationMethod with `id` equal to `kid` is found.
-3. A [Recover Operation](https://identity.foundation/sidetree/spec/#recover) is only valid when signed by a key listed in `recovery`. Other operations are only valid when signed with keys listed in `operations`. If the key is not present in the expected collection, verification fails, and the operation is considered invalid. If the key is present in the expected collection proceed.
+3. A [Recover Operation](https://identity.foundation/sidetree/spec/#recover) or a [Deactivate Operation](https://identity.foundation/sidetree/spec/#deactivate) are only valid when signed by a key listed in `recovery`. Other operations are only valid when signed with keys listed in `operations`. If the key is not present in the expected collection, verification fails, and the operation is considered invalid. If the key is present in the expected collection proceed.
```
csuwildcat, line 28: I was about to merge, but I am wondering what the implications of this are - is this saying that it expects a compliant implementation to have added two new Verification Methods, `operations` and `recovery`, to DID Documents it outputs?

OR13: Hmm, I feel like this should not be repeated here... and instead, we should refer to https://identity.foundation/sidetree/spec/#verifying

which describes how to verify a JWS under both the "internal data model" and "did core / vc-jwt" paradigm.

OR13: @csuwildcat we had a call about this exact issue a while ago, i'm happy to sync on it again... essentially we need to define "how do we verify JWS that are linked to sidetree DIDs"... Lack of clarity around how JOSE works with DIDs makes this very difficult... the VC Data Model does not define how key resolution works... The section I linked attempts to fix this issue.

csuwildcat: Yeah, let's talk. The spec is inaccurate right now - namely: there is no such `recovery` op usage designation for recovery keys across implementation choices. I was really confused by the lack of clarity around where (if the implementer chooses the IETF Patch Action) in the DID Document the implementer is required to leak the Sidetree-internal references/logic for which keys are recovery/operation-capable. The root of the issue seems to be that the IETF Patch Action does not (refuses to?) clearly specify these rules within that Patch Action. Leaving that in the Context section for the reader to somehow divine is odd and will likely lead to implementation difficulty.

OR13: This PR is about how to know if an operation is valid... an operation is valid if the JWS can be verified... which should be specified once, clearly in a section called 'Operation Verification' ... currently the best we have is the section on how to verify a JWS.

A sidetree method that chooses not to expose recovery and operation keys is just relying on security through obscurity, and loosing the ability to reuse those keys for things like credential issuance or capability invocation.... I will object to making that a recommended default, but I won't object to allowing it.

csuwildcat: There must be some misunderstanding, because the current spec is explicit that there is exactly one recovery key, and that the recovery key is, and can only be, the currently active recovery key. Further, the operation keys are ID'd, and the lookup happens against those IDs. There is no ambiguity in the spec on this point, at least in relation to the patch route that removes the ambiguity. This section appears to introduce a requirement that you output keys in special places in the DID Doc so the implementation can then identify to itself which keys can be used for what. This is strange, because without any mention of that in the section where patches are actually described, this is very obtuse for the reader. You've essentially hidden a requirement about how keys need to be represented in your patches inside of a downstream verification flow. This should be corrected by enforcing in the IETF JSON Patch variant that there is indeed a mandate that, if you choose that avenue, you MUST output your op/recovery keys to those areas of a DID Doc.

OR13:
I had a call with @csuwildcat about this, summarizing here:

1. We agreed to externalize `operations` and `recovery` consistently.
2. We agreed to define JWS verification using the existing DID Core terminology ( verificationMethod, proofPurpose)
3. We will try and revise the JWS verification section so that we don't need to define verification of the "internal data model"... but also, so we don't need to explicitly require the implementer externalize the data model before performing verification.... in english, this, means we will define that methods are responsible for converting `kid` to public key bytes, and checking if the verificationMethod referenced is authorized for proofPurpose `operations` or `recovery`.

I will try and take a stab at a PR that implements these change, if it successful it will cover the changes noted in this PR.

PR #666, opened 2020-05-14
Update resolution for order agnostic op processing

csuwildcat:


OR13:
proof that daniel works for satan... #666

PR #669, opened 2020-05-16
Move reveal values into signed payloads

csuwildcat:


docs/spec/common-functions.md
```diff
@@ -14,4 +14,10 @@ Pseudo-code example using current protocol defaults:
...
 let HashingOutput = Base64URL( Multihash(DATA, 'sha2-256') );
-```
\ No newline at end of file
+```
+
+### Commitment Value Generation
+
+All commitment values created for the commit/reveal function of Sidetree operations ****MUST**** be sufficiently random values (e.g. a 32 byte cryptographically secure string) and ****MUST**** be unique across operations, without reuse in the operational lineage of a DID.
+
+Commitment values ****SHOULD**** be deterministically regenerable using the keys bound to a given DID. This eliminates the addition of another sensitive value User Agent applications would need to track, secure, and store, which subsequently reduces the possibility of loss.
```
kdenhartog, line 23: I think I'm following what you're after here. Are you thinking we'll be generating these commitment values by doing something like Hash(`privateKey` || `dataCommited`) ?

csuwildcat:
The hash can't be of a known value, like the data you're committing, because then others would know the reveal value that protects you from spam. It would be more like wallets using a DID's seed-regenerable private key to sign a deterministic incrementing value that changes for each op, and using it to hash the leading 32 bytes of that signature as the commitment. This way the entire thing is regenerable, if you assume a deterministic K value in secp256k1 (or are using some other key type with deterministic signing assurance)

PR #670, opened 2020-05-18
Define Operation Verification

OR13:
Addresses: https://github.com/decentralized-identity/sidetree/issues/659

### DIFF
--- a/docs/spec/json-web-signatures.md
+++ b/docs/spec/json-web-signatures.md
@@ -32,14 +32,12 @@ The JWS header is parsed and a `kid` is extracted.
 1. Iterate the verificationMethods, until a verificationMethod with `id` equal to `kid` is found.
 2. Convert the discovered verificationMethod to JWK if necessary.
 3. Perform [JWS Verification](https://tools.ietf.org/html/rfc7515#section-5.2) using the JWK.
-4. The operation is considered valid if key material was present in the correct collection and the signature is valid.
-
 
 #### Operation Verification
 
 Sidetree operations are considerd valid when the JWS can be verified, and where the key used is associated with the correct proof purpose.
 
-`operations`, `recovery` are proof purposes for verifying sidetree operations which sidetree DID Methods MUST support. A sidetree operation MUST be signed by a key associated with exactly one of these proof purposes.
+`operations`, `recovery` are proof purposes for verifying sidetree operations which sidetree DID Methods MUST support, but which MAY NOT be exposed externally via the DID Document or Resolver Method Meta Data.
 
 An [Update Operation](https://identity.foundation/sidetree/spec/#update) MUST be signed by a key associated with the `operations` proof purpose. 
 
--- a/docs/spec/json-web-signatures.md
+++ b/docs/spec/json-web-signatures.md
@@ -35,11 +35,11 @@ The JWS header is parsed and a `kid` is extracted.
 
 #### Operation Verification
 
-Sidetree operations are considerd valid when the JWS can be verified, and where the key used is associated with the correct proof purpose.
+Sidetree operations are considered valid when the JWS can be verified, and where the verificationMethod used is associated with the correct proof purpose.
 
-`operations`, `recovery` are proof purposes for verifying sidetree operations which sidetree DID Methods MUST support, but which MAY NOT be exposed externally via the DID Document or Resolver Method Meta Data.
+`operation`, `recovery` are proof purposes for verifying sidetree operations which sidetree DID Methods MUST support, and which MAY be exposed externally via the DID Document or Resolver Method Meta Data.
 
-An [Update Operation](https://identity.foundation/sidetree/spec/#update) MUST be signed by a key associated with the `operations` proof purpose. 
+An [Update Operation](https://identity.foundation/sidetree/spec/#update) MUST be signed by a key associated with the `operation` proof purpose. 
 
 A [Recover Operation](https://identity.foundation/sidetree/spec/#recover) MUST by signed by a key associated with the `recovery` proof purpose. 
 
@@ -62,5 +62,5 @@ DID Core also defines proof purposes which sidetree DID Methods MAY support.
 :::
 
 ::: warning
-  It is not recommended to reuse verificationMethod's for multiple proof purposes.
+  It is not recommended to reuse verificationMethods for multiple proof purposes.
 :::
--- a/docs/spec/json-web-signatures.md
+++ b/docs/spec/json-web-signatures.md
@@ -19,25 +19,50 @@ Here is an example of a decoded JWS header:
 }
 ```
 
+::: warning
+  It is recommended that `kid` be a DID URL. If it is not, method implementers might need to rely on additional context to uniquely identify the associated verificationMethod. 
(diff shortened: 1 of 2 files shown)


docs/spec/json-web-signatures.md
```diff
@@ -19,25 +19,50 @@ Here is an example of a decoded JWS header:
...
-3. A [Recover Operation](https://identity.foundation/sidetree/spec/#recover) is only valid when signed by a key listed in `recovery`. Other operations are only valid when signed with keys listed in `operations`. If the key is not present in the expected collection, verification fails, and the operation is considered invalid. If the key is present in the expected collection proceed.
-3. Convert the discovered verificationMethod to JWK if necessary.
-4. Perform [JWS Verification](https://tools.ietf.org/html/rfc7515#section-5.2) using the JWK.
-5. The operation is considered valid if key material was present in the correct collection and the signature is valid.
+The JWS header is parsed and a `kid` is extracted.
 
+1. Iterate the verificationMethods, until a verificationMethod with `id` equal to `kid` is found.
+2. Convert the discovered verificationMethod to JWK if necessary.
+3. Perform [JWS Verification](https://tools.ietf.org/html/rfc7515#section-5.2) using the JWK.
+4. The operation is considered valid if key material was present in the correct collection and the signature is valid.
```
OR13, line 35: This last line seems to belong under operation verification.

docs/spec/json-web-signatures.md
```diff
@@ -19,25 +19,48 @@ Here is an example of a decoded JWS header:
...
-If `kid` is not a DID URL, the verification key MAY be resolved as follows:
+#### Operation Verification
 
-1. Iterate an internal representation of keys registered for use with the DID until a jwk with `kid` equal to `kid` is found.
-2. Ensure that the `usage` property of the internal key represenation is `recovery` if the operation is a recovery operation or `ops` if the operation is not Recovery or Update.
-3. Perform [JWS Verification](https://tools.ietf.org/html/rfc7515#section-5.2) using the JWK.
-4. The operation is considered valid if key material was present with the correct usage and the signature is valid.
+Sidetree operations are considerd valid when the JWS can be verified, and where the key used is associated with the correct proof purpose.
+
+`operations`, `recovery` are proof purposes for verifying sidetree operations which sidetree DID Methods MUST support, but which MAY NOT be exposed externally via the DID Document or Resolver Method Meta Data.
```
OR13, line 40: @csuwildcat I think its still true that all Sidetree DID Methods MUST support operations and recovery... I think what you are saying is that these proof purposes MAY NOT be exposed via either the did document or resolver meta data.

docs/spec/patches.md
```diff
@@ -147,6 +147,10 @@ The `ietf-json-patch` Patch Action describes a mechanism for modifying a DID's s
 1. The object ****MUST**** include an `action` property, and its value ****MUST**** be `ietf-json-patch`.
 2. The object ****MUST**** include a `patches` property, and its value ****MUST**** be an array of [IETF JSON Patch](https://tools.ietf.org/html/rfc6902) operation objects.
 
+If `ietf-json-patch` is used to add or remove from a proof purpose collection, such as `operations`, `recovery` or `assertionMethod`, per the DID Core spec, a unique `id` property MUST be present.
```
troyronda, line 150: I assume unique for that collection.

docs/spec/json-web-signatures.md
```diff
@@ -19,25 +19,48 @@ Here is an example of a decoded JWS header:
...
+`authentication` for use with Verifiable Presentations, and general authentication flows.
+`capabilityInvocation` and `capabilityDelegation` for use with Object Capabilities used by Secure Data Stores / Encrypted Data vaults. 
+
+::: warning
+  verificationMethod objects can be embedded, or referenced by `id`.
+:::
 
 ::: warning
-  Operations may be valid, and yet still rejected in the resolution process.
+  It is not recommended to reuse verificationMethod's for multiple proof purposes.
```
troyronda, line 65: I assume "verificationMethods"

docs/spec/json-web-signatures.md
troyronda, line 40: I assume `operation` would be singular to match with other verification methods / proof purposes?

csuwildcat: To clarify: I am definitely saying all Sidetree methods need to somehow track keys in relation to those two purposes, because they need to know what keys do what things, but simply asking that we either don't forcibly project out op/recovery-only keys into the doc, or at least make it optional and caution that if you do project these keys, a wallet implementer must be hyper vigilant in the DID Doc-referenced invocations they execute in their business logic, because leaving a 'hanging chad' could mean the wallet inadvertently interacts with a key and an external party that was not intended.

If this text means: "You need to have a _concept_ of these purposes, such that an op can be processed against the right key, but DO NOT project the keys into the doc unless they are explicitly marked as dual-use keys the user wants to put in, say, `authentication`", then yes, I am totally fine with that. I was not arguing that it had to be a MAY NOT on the latter part of that language, because I wanted to be sensitive to others who view the attack surface/vectors differently than I do. Still fine just leaving that a MAY, if it makes you more comfortable for Element.

docs/spec/json-web-signatures.md
```diff
@@ -19,25 +19,48 @@ Here is an example of a decoded JWS header:
...
+3. Perform [JWS Verification](https://tools.ietf.org/html/rfc7515#section-5.2) using the JWK.
 
-If `kid` is not a DID URL, the verification key MAY be resolved as follows:
+#### Operation Verification
 
-1. Iterate an internal representation of keys registered for use with the DID until a jwk with `kid` equal to `kid` is found.
-2. Ensure that the `usage` property of the internal key represenation is `recovery` if the operation is a recovery operation or `ops` if the operation is not Recovery or Update.
-3. Perform [JWS Verification](https://tools.ietf.org/html/rfc7515#section-5.2) using the JWK.
-4. The operation is considered valid if key material was present with the correct usage and the signature is valid.
+Sidetree operations are considerd valid when the JWS can be verified, and where the key used is associated with the correct proof purpose.
```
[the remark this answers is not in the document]
OR13: thanks, it should say verifcationMethod instead of key... we could theoretically use `ethereumAddress` or `bitcoinAddress` and EC Recover to verify sidetree operations.

docs/spec/json-web-signatures.md
OR13: good catch.

docs/spec/json-web-signatures.md
OR13: yes.

docs/spec/patches.md
OR13: yes, I attempted to clarify this further.

OR13:
https://github.com/decentralized-identity/sidetree/pull/670#discussion_r427652908

I agree MAY NOT is confusing, I switched it to MAY.

docs/spec/json-web-signatures.md
OR13: @csuwildcat can you review again, and either approve or request change.

Issue #654, opened 2020-05-07
Ensure resolution processing accounts for true operation order regardless of ledger time order


csuwildcat:
Operations should be processed based on their deterministic, owner-intended order from one to another, regardless of the exact order of their anchoring transactions. Operation ordering should transcend ledger time ordering, with ledger time ordering being used as a 'tie breaker' if there are two valid ops that deterministic ancestors of a previous operation. In a case where two operations share the same ancestor, the earliest in ledger time order shall be deemed valid. There are two options for v1 of the spec:

1. Use what is available within operations today to accomplish this by iterating all operations and processing them in the linked order their commit/reveal values naturally form, using earliest ledger time to resolve any conflicts (as noted above).

2. Add a version number to operations that makes the order explicit and enables more performant processing. In a version number scheme, you would want two numbers that separate Recovery/Checkpoint ops from Update ops. This would be a new value in the Anchor/Map File entries of each operation - for example: `{ op_number: 1.2, ... }`, wherein the `1` means that it is the first recovery in recovery ordering, and the `2` means the operation is the second update after the first recovery. During resolution you would sort by this number, then sort by ledger time order, resulting in the earliest numbers first, then ordered by earliest appearance in the ledger. This would provide an efficient basis for iterating operations during resolution.

My preference is Option 2, but we could do Option 1 if we didn't want to add an explicit mechanism and specification text that directs its use.

OR13:
+1 to option 1, thanks for the clear description of the problem.

csuwildcat:
@OR13 I screwed up and transposed my preference, so if you think Option 1, maybe we should chat?

csuwildcat:
Pros of Option 1: it requires very little spec change, and technically we can make sure to do this today (so, for example, ops aren't needlessly dropped if transactions land out of the expected order) using the values in the system we already have. Cons: it doesn't allow for straightforward sorting and order processing, which will be far less performant.

Pros of Option 2: it would allow fast sorting of ops and you could process them quickly in one pass, without retracing a bunch. It also could unlock the ability to migrate ledgers more easily (if that was ever necessary)

OR13:
If you don't need it for security reasons, it shouldn't be added.

Can you walk me through how I would abuse this as an attacker who had discovered a commit / reveal value?... pretty sure I can force a bunch of garbage operations to be stored...

I think there are a bunch not fully documented security considerations with option 2.

csuwildcat:
[...]

PR #674, opened 2020-05-19
#654 - Use commit-reveal chain as operation order instead of ledger time

thehenrytsai:


lib/core/versions/0.7.0/OperationProcessor.ts
```diff
@@ -62,6 +64,27 @@ export default class OperationProcessor implements IOperationProcessor {
...
 
+  public async getRevealValue (anchoredOperationModel: AnchoredOperationModel): Promise<Buffer> {
+    if (anchoredOperationModel.type === OperationType.Create) {
+      throw new SidetreeError(ErrorCode.OperationProcessorCreateOperationDoesNotHaveRevealValue);
+    }
+
+    const operation = await Operation.parse(anchoredOperationModel.operationBuffer);
+
+    let encodedRevealValue;
+    if (operation.type === OperationType.Recover) {
```
OR13, line 75: nit: prefer to see switch statements.

lib/core/Resolver.ts
```diff
@@ -2,79 +2,231 @@ import AnchoredOperationModel from './models/AnchoredOperationModel';
...
+  private async applyUpdateOperations (startingDidState: DidState, commitValueToOperationMap: Map<string, AnchoredOperationModel[]>)
+    : Promise<DidState> {
+    let didState = startingDidState;
 
-    // Apply "update/delta" operations.
-    didState = await this.applyOperations(updateOperationsToBeApplied, didState);
+    while (commitValueToOperationMap.has(didState.nextUpdateCommitmentHash!)) {
+      let operationsWithCorrectRevealValue: AnchoredOperationModel[] = commitValueToOperationMap.get(didState.nextUpdateCommitmentHash!)!;
+
+      // Sort using blockchain time.
```
OR13, line 144: is this list size controllable by an attacker? In other words, once a reveal value is published, can an attacker replay it to cause this list to be arbitrarily large?

lib/core/Resolver.ts
```diff
@@ -2,79 +2,231 @@ import AnchoredOperationModel from './models/AnchoredOperationModel';
...
+    let didState = await this.applyCreateOperation(operationsByType.createOperations);
 
-    // If no valid full operation is found at all, the DID is not anchored.
+    // If can't construct an initial DID state.
     if (didState === undefined) {
       return undefined;
     }
 
-    // If last operation is a deactivate. No need to continue further.
+    // Apply recovery/deactivate operations until an opertaion matching the next recovery commitment cannot be found.
```
OR13, line 36: "opertaion" => "operation"

tests/core/Resolver.spec.ts
```diff
@@ -187,5 +189,107 @@ describe('Resolver', () => {
...
+
+      // Intentionally insert earliest valid recover operation in between the other two operations to test sorting.
+      // Intentionally using the resolver's map construction method to test operations with the same reveal value are placed in the same array.
+      const updateCommitValueToOperationMap: Map<string, AnchoredOperationModel[]>
+        = await (resolver as any).constructCommitValueToOperationLookupMap([updateOperation3, updateOperation1, updateOperation2]);
+      const nextUpdateCommitment = createOperationData.createOperation.delta!.updateCommitment;
+      const updatesWithSameReveal = updateCommitValueToOperationMap.get(nextUpdateCommitment);
+      expect(updatesWithSameReveal).toBeDefined();
+      expect(updatesWithSameReveal!.length).toEqual(3);
+      // const updateCommitValueToOperationMap = new Map<string, AnchoredOperationModel[]>();
```
OR13, line 270: nit: remove commented code

lib/core/Resolver.ts
csuwildcat: An attacker can always _attempt_ to replay any commit values, even invalid ones. However, the entire point of the commit reveal scheme is that the hash check to determine validity is much faster than a sig check, and because we're ledger time ordering for each claimed commitment, replayed commits will usually never be evaluated, because the user's correct commit will be encountered first and used. Because the user's commitments generally will appear first, their successful eval mean you can simply stop iterating that commit's bucket of possible matching ops.

tests/core/Resolver.spec.ts
thehenrytsai: Oops... will definitely remove!

lib/core/Resolver.ts
thehenrytsai: Thanks!! Fixed all occurrence.

lib/core/versions/0.7.0/OperationProcessor.ts
thehenrytsai: Personally I don't refer it, but since you asked, I will do it just for you @OR13!

tests/core/Resolver.spec.ts
```diff
@@ -187,5 +189,107 @@ describe('Resolver', () => {
...
+      spyOn(operationProcessor, 'apply').and.throwError('any error');
+
+      const createOperationData = await OperationGenerator.generateAnchoredCreateOperation({ transactionTime: 1, transactionNumber: 1, operationIndex: 1 });
+      const initialDidState = await (resolver as any).applyOperation(createOperationData.anchoredOperationModel, undefined);
+
+      // Expecting undefined to be returned instead of error being thrown.
+      expect(initialDidState).toBeUndefined();
+      done();
+    });
   });
```
mudiali, line 292: Similar comment as above. I think we are missing tests for the `resolve` where if applying create operation didn't return anything then we return undefined. Similar for recovery/update application etc. Again I would mock the individual `apply...` function instead of constructing complex inputs.

lib/core/Resolver.ts
mudiali: Orie has a good point about the list being too large though but I'm not sure if we have a good way to block it.

lib/core/Resolver.ts
```diff
@@ -2,79 +2,231 @@ import AnchoredOperationModel from './models/AnchoredOperationModel';
...
     didState: DidState | undefined
     ): Promise<DidState | undefined> {
     let appliedDidState = didState;
+
+    // NOTE: MUST NOT throw error, else a bad operation can be used to denial resolution for a DID.
+    try {
+      const operationProcessor = this.versionManager.getOperationProcessor(operation.transactionTime);
+      appliedDidState = await operationProcessor.apply(operation, appliedDidState);
+    } catch (error) {
+      console.log(`Skipped bad operation for DID ${operation.didUniqueSuffix} at time ${operation.transactionTime}. Error: ${error}`);
```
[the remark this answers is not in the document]
thehenrytsai: Thanks @mudiali for reminding!

tests/core/OperationProcessor.spec.ts
```diff
@@ -627,4 +627,13 @@ describe('OperationProcessor', async () => {
       });
     });
   });
+
+  describe('getRevealValue()', () => {
+    it('should throw if a create operation is given.', async () => {
```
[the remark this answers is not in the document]
thehenrytsai: We have tests that tests these happy paths already, but true not directly against this method. I am seeing recurring theme about how and what to test, me probably need a meeting.

tests/core/Resolver.spec.ts
```diff
@@ -187,5 +189,107 @@ describe('Resolver', () => {
...
+
+      // Expecting the new state to contain info of the first recovery operation.
+      expect(newDidState.lastOperationTransactionNumber).toEqual(2);
+      expect(newDidState.nextRecoveryCommitmentHash).toEqual(recoveryOperation1Data.recoverOperation.signedData.recovery_commitment);
+
+      done();
+    });
+  });
+
+  describe('applyUpdateOperations()', () => {
```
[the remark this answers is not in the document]
thehenrytsai: Agree to no to one-BIG-test-to-rule-them-all, and am guilty with a few of those tests!

But I am a believer in simple-scenario-test-that-hits-them-all if you will. My principal for writing unit tests is actually to hit as much real code as possible and only mock in the face of: complexity, reliability, or latency. In this case I don't think any of the 3 exceptions apply. 

Here I need to make sure sort works before passing to `applyFirstValidOperation()`, if I mock `applyFirstValidOperation`'s return value, then I would not know if I have sorted the ops correctly (e.g. reverse order). 

`didState` cannot be `undefined` as per method signature.

tests/core/Resolver.spec.ts
thehenrytsai: Multiple scenario hits the `applyCreateOperation()` returning `undefined`:

* should ignore a deactivate operation of a non-existent did
* should ignore updates to DID that is not created

For this test I did mock the `apply()` method. I do hear the competing principal for writing these tests. So let's chat.

lib/core/Resolver.ts
thehenrytsai: One cannot prevent an attacker playing an op using the same reveal value, the simplest way to do so is just keep replaying the same valid op again and again. But as @csuwildcat pointed out, we will process the the first valid one in blockchain time and discard the rest.

One thing worth noting is that in operation store we already index by blockchain time and we get by sorting in this order. So this sort here is potentially "not needed", and likely an order N operation in practice.

PR #679, opened 2020-05-20
move reveal value into signed data

isaacJChen:


tests/core/RecoverOperation.spec.ts
```diff
@@ -74,7 +74,15 @@ describe('RecoverOperation', async () => {
...
-      (recoverOperationRequest.recovery_reveal_value as any) = 123; // Intentionally incorrect type.
+      const signedDataPayloadObject = {
+        delta_hash: 'deltaHash',
+        recovery_reveal_value: 123,
+        recovery_key: newRecoveryPublicKey,
+        recovery_commitment: unusedNextRecoveryCommitmentHash
+      };
+      const signedData = await OperationGenerator.signUsingEs256k(signedDataPayloadObject, recoveryPrivateKey);
+
+      recoverOperationRequest.signed_data = signedData; // Intentionally incorrect type for reveal value.
```
thehenrytsai, line 85: Since we are just testing for the error thrown which got moved into , consider just do what "should throw if signedData contains an additional unknown property." does? ie. Move this test under `parseSignedDataPayload()` and invoke `parseSignedDataPayload()` directly and bypass/remove the additional setup code earlier. Same goes to other occurences.

lib/core/versions/latest/OperationProcessor.ts
```diff
@@ -71,13 +71,13 @@ export default class OperationProcessor implements IOperationProcessor {
     let encodedRevealValue;
     switch (operation.type) {
       case OperationType.Recover:
-        encodedRevealValue = (operation as RecoverOperation).recoveryRevealValue;
+        encodedRevealValue = (operation as RecoverOperation).signedData.recovery_reveal_value;
```
thehenrytsai, line 74: My fault: do you think we should define the internal `signedDataModel` with camelCasing (only parser handles snake_casing validation and conversion in very isolated places) so we can get rid of this ugly snake_casing code creeping into places now? `CreateOperation` already does this for `SuffixDataModel`... again this was me letting this slipped through the cracks...

Issue #329, opened 2019-11-21
Revisit the BitcoinProcessor::revertDatabase(...) function refactor


mudiali:
The revert logic in the bitcoin-processor seems different from the similar logic in Observer.ts::revertInvalidTransactions(...). Also, there could be a bug in the bitcoinprocessor.ts logic where there might be a loop if the firstTransaction in the DB is invalid. See this comment: https://github.com/decentralized-identity/sidetree/pull/315#discussion_r348766258

Another potential improvement with this bug could be how we keep the 'lastProcessedTransaction' variable updated. If we refactor the revertInvalidTransaction function as following two functions:
1. findFirstValidTransaction
1. revertTransactionsGreaterThan(block: number)

Then the current usage of the revertInvalidTransaction within processTransactions(...) function will change to the following. 
````
private async processTransactions (startBlock?: IBlockInfo, endBlockHeight?: number): Promise<IBlockInfo> {
  ...
  if (!startValid) {
    this.lastProcessedBlock = await this.findFirstValidTransaction();
    revertTransactionGreatherThan(this.lastProcessedBlock!height);

    startBlockHeight = this.lastProcessedBlock.height;
   }
  ...
}
````
This way we make sure that the lastProcessedBlock is updated and the worst behavior for the client is that they will get stale data (see lastProcessedBlock usage in ````getNormalizedFee```` function):

thehenrytsai:
Implemented.

PR #684, opened 2020-05-22
#329 - Simplified bitcoin processor revert logic

thehenrytsai:


lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -466,28 +466,29 @@ export default class BitcoinProcessor {
...
+    // If there is a valid processed block, we will start processing the block following it, else start processing from the genesis block.
+    const startingBlockHeight = lastValidBlock ? lastValidBlock.height + 1 : this.genesisBlockNumber;
+
+    return this.bitcoinClient.getBlockInfoFromHeight(startingBlockHeight);
   }
 
   private async getStartingBlockForPeriodicPoll (): Promise<IBlockInfo | undefined> {
 
-    const lastProcessedBlockVerified = await this.verifyBlock(this.lastProcessedBlock!.height, this.lastProcessedBlock!.hash);
+    const lastProcessedBlockValid = await this.verifyBlock(this.lastProcessedBlock!.height, this.lastProcessedBlock!.hash);
```
mudiali, line 477: Generally I'm a bit wary of the revertDatabases() returning undefined. So on line 483, the `lastProcessedBlock` could be undefined but on line 477 here the `lastProcessBlock!` is asserted. I think it will be better if each routine just returns the 'correct' last valid block.

thehenrytsai: One of the main point for this PR is to call out that there is no such "last valid block" if entire DB is reverted, in which case `undefined` to me is the right value to be returned by `revertDatabases()`.

The ! asserts here is not related to the PR change and I tried not to touch that part of the code, this was just a variable rename.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -497,51 +498,42 @@ export default class BitcoinProcessor {
   }
 
   /**
-   * Begins to revert databases until consistent with blockchain, returns last good height
-   * @returns last valid block height before the fork
+   * Begins to revert databases until consistent with blockchain.
+   * @returns A known valid block before the fork. `undefined` if no known valid block can be found.
    */
-  private async revertDatabases (): Promise<IBlockInfo> {
+  private async revertDatabases (): Promise<IBlockInfo | undefined> {
```
mudiali, line 504: I think if we change the contract to say that the function returns then we don't have to return undefined.

> A known valid block before the fork. genesisBlock if no known valid block can be found.

thehenrytsai: Synced up offline and took a valid suggestion of not assigning `this.lastProcessedBlock` to undefined.

lib/bitcoin/BitcoinProcessor.ts
thehenrytsai: Synced up offline and took a valid suggestion of not assigning `this.lastProcessedBlock` to undefined.

PR #687, opened 2020-05-24
Further generalized two getStartingBlock...() methods

thehenrytsai:


lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -159,11 +159,12 @@ export default class BitcoinProcessor {
...
-    const startingBlock = await this.getStartingBlockForInitialization();
-
-    console.info(`Starting block: ${startingBlock.height} (${startingBlock.hash})`);
-    await this.processTransactions(startingBlock);
+    // NOTE:
+    // We always need to start the processing from the first block of a fee sampling group
+    // so that in-memory state for fee sampling will be repoppulated yielding correct fee calculation,
+    // so we trim the databases to make sure this condition is met.
+    // We also initialize the `lastProcessedBlock` so that processing loop can begin.
+    this.lastProcessedBlock = await this.trimDatabasesToLastFeeSamplingGroupBoundary();
```
thehenrytsai, line 167: @mudiali, this is what was logically different from between the two `getStartingBlock...()` methods. This allows as to get rid of one of the two methods entirely.

A question for you: what makes this.lockMonitor.initialize(); break if not all blocks are processed? I ask because if there is no way around it, we can definitely invoke `getStartingBlockForPeriodicPoll()` and `processingTransaction()` like the original code, I took it away because `periodPoll()` would have done the same thing already, of course this requires `this.lockMonitor.initialize()` to not break.

OR13:
CI is failing, we should generally avoid reviewing this in this state.

lib/bitcoin/BitcoinProcessor.ts
thehenrytsai: Synced up with @mudiali and will keep the existing requirement for processing all transactions in `initialize()` and differ improvements to issue #692.

thehenrytsai:
@OR13 Agree in general with your statement. As a side note, we should takeout travis CI as it tends to be less stable and we have no need of it now.

Issue #689, opened 2020-05-26
Abstract Feedback


OR13:
Handle concept of aging... remove "emerging".

https://w3c-ccg.github.io/did-method-registry/ -> did spec registries?

tplooker:
"user owned" or "user controlled"?

OR13:
@tplooker we should avoid the word "owner", did core already does this... please open a PR :)

PR #693, opened 2020-05-27
feat: rewording of abstract

tplooker:
I had a bit of an attempt at re-wording the abstract along the lines of what we discussed on the call. Open to thoughts and suggestions @csuwildcat @OR13


tplooker:
Tagging #689 for relevance

Issue #697, opened 2020-05-28
Remove recovery reveal values


thehenrytsai:
Remove recovery reveal values, instead use the hash of the recovery public key as the recovery commitment, and only expose the recovery public key upon recover and deactivate operations.

thehenrytsai:
Implemented.

PR #699, opened 2020-05-28
#697 - Removed recovery reveal values

thehenrytsai:


package.json
```diff
@@ -9,6 +9,7 @@
   "dependencies": {
     "base64url": "3.0.1",
     "bitcore-lib": "8.7.1",
+    "canonicalize": "1.0.1",
```
mudiali, line 12: Sanity: Just a concern that if there are changes to the canonicalize scheme b/w versions then we could have a problem. We can decide never to change it and that's ok unless there's a vulnerability and we are forced to upgrade.

csuwildcat: This is true, but this is also an implementation of an IETF draft that is about to move to the standards track (https://tools.ietf.org/html/draft-rundgren-json-canonicalization-scheme-17), so I don't think they're going to change much of it in any significant ways. One reason I doubt we would be affected is that most of the more complex parts of JCS come into play with more complex payloads of exotic values, unicode points, etc. and given we're dealing with a roughly flat set of objects with normal values, I doubt they'll change the most primitive ordering logic for props, for example.

thehenrytsai: @csuwildcat can add input here. The assumption here is that the canonization implements JCS correctly, same goes to how we assume the JWK library we depend on implements JWK spec correctly. You are right we would have a problem if they end up being spec non-compliant, I think we would want to talk about which of the two obvious choices we do in that case.

PR #703, opened 2020-05-29
Update implementation.md

julioCabDu:
Update image links and make transactionTimeHash cohesive with transactionTime in the examples.


julioCabDu:
Also, what about naming blockTime and blockTimeHash instead of transactionIdem? Since transactions are timestamped in blocks.

thehenrytsai:
Thanks @julioCabDu for the PR, approved. Please file a separate issue for the `ledgerTime` rename suggestion so we can track it!

Issue #688, opened 2020-05-26
Change anchored operation count to plain number instead of encoded value


csuwildcat:
Currently looks like this: `ECcAAA.QmWd5PH6vyRH5kMdzZRPBnf952dbR4av3Bd7B2wBqMaAcf`

Will look like this: `10000.QmWd5PH6vyRH5kMdzZRPBnf952dbR4av3Bd7B2wBqMaAcf`

thehenrytsai:
Assigning to Daniel to update spec and close.

csuwildcat:
Added in latest PR

PR #705, opened 2020-05-29
#688 - Changed anchored operation count format

thehenrytsai:


lib/core/versions/latest/AnchoredDataSerializer.ts
```diff
@@ -40,47 +35,34 @@ export default class AnchoredDataSerializer {
...
-      throw new SidetreeError(ErrorCode.AnchoredDataNumberOfOperationsNotFourBytes,
-                              `Input has ${bytesBuffer.length} bytes.`);
+  private static parsePositiveInteger (input: string): number {
+    // NOTE:
+    // /<expression>/ denotes regex.
+    // ^ denotes beginning of string.
+    // $ denotes end of string.
+    // [1-9] denotes leading '0' not allowed.
+    // \d* denotes followed by 0 or more decimal digits.
+    const isPositiveInteger = /^[1-9]\d*$/.test(input);
```
mudiali, line 60: Consider: may be possible to avoid regex and make it simpler?

````javascript
const number = Number(input);

if (Number.isNaN(number) || number < 0) throw;
````

thehenrytsai: Will keep regex, `Number("0011")` is parsed as 11, which is not what we would want to allow.

PR #709, opened 2020-05-30
Clarify serialization of JWS

jmandel:
I'm not sure if the comment should be trusted or if the example data should be trusted here -- but

*  example data looks like a JSON serialization format, i.e. per https://tools.ietf.org/html/rfc7515#section-3.2

* Comment suggests compact serialization format, i.e. per https://tools.ietf.org/html/rfc7515#section-3.1

I've written this pull request assuming that the example is correct on the comment is wrong, but if I have it backwards then the example should be updated.

### DIFF
--- a/docs/spec/file-structures.md
+++ b/docs/spec/file-structures.md
@@ -26,7 +26,7 @@ Anchor Files contain [Create](#create), [Recover](#recover), and [Deactivate](#d
     "recover": [
       {
         "did_suffix": SUFFIX_STRING,
-        "signed_data": { // Base64URL encoded, compact JWS
+        "signed_data": { // JSON serialized JWS per rfc7515#section-3.2
           "protected": {...},
           "payload": {
             "recovery_commitment": COMMITMENT_HASH,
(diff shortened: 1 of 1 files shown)


jmandel:
Looking through the rest of the specification, I think maybe I do have a backwards here. If that's the case: the ad hoc syntax for representing a compact JWS is super misleading. It would be better to omit the detailed sub-properties from the example JSON and provide them below rather than making it look like inline JSON.

OR13:
@csuwildcat feel free to close this if you take it over.

OR13:
@csuwildcat @jmandel any updates on this?

jmandel:
I continue to find the ad-hoc syntax very confusing. If I'm the only one, we can close this :-)

csuwildcat:
I'll try to get to it today.

OR13:
@jmandel I also find it confusing. My preference is for us to use JSON, with real world examples, and comments to add context. I think we may be able to get there now that ION is committed to less breaking changes.

OR13:
@csuwildcat @jmandel 25 day old PR on a critical security related topic. I will close this in 24 hours, if no action is taken :)

OR13:
Closed due to inactivity.

PR #710, opened 2020-05-30
Only limit key types for `ops` keys

jmandel:
There are a couple of practical issues with limiting the key types in this way:

* The registry of cryptographic suites includes only signature algorithms and no encryption algorithms. So if you are registering a key for purposes other than authentication, there will not be any suitable cryptographic suite value in the current registry.

* over time the list of registered values will change. As new values appear on the registry, would previously invalid update operations become retroactively valid? This would change the history of the identifier document, which seems obviously wrong. But there is no temporal link to say "as of when", so I'm not sure how this is meant to work.

It seems easier just to eliminate this restriction on key types, or perhaps apply the restriction  only when the key is being registered as a sidetree ops key.


OR13:
I agree, i think this is an ION thing that is overly constraining the sidetree spec.

csuwildcat:
This was more related to our collective decision to stick to JWK-based representation for sidetree operation keys. Are you all saying we should not check the key types for valid JWK types, even though we all agreed that we would use JWKs? How is this ION-specific?

csuwildcat:
Please do not merge this at this point, because it may represent an implementation breaking change. We must discuss this on the next Sidetree call.

jmandel:
To be clear, there's no problem with requiring jwk based keys -- the problem (for ION and sidetree in general) is avoiding constraints on `type` to be sure we don't prevent encryption keys and other key types from being added to a Did document.

(In prototypes I've had to *lie* about the key type in order to get my keys added, given the current restrictions.)

csuwildcat:
Proposed language modification for 3.2 of the section in question:

> The object MUST include a type property, and its value SHOULD be the identifier type string of a registered Cryptographic Suite that supports the publicKeyJwk representation - for example:
> - EcdsaSecp256k1VerificationKey2019
> - JwsVerificationKey2020

jmandel:
A `SHOULD` here would solve my problem for sure -- but it's still kind of frustrating because it seems to imply that developers "should" not be using encryption key-types, because you need to violate the "should" to use one.

I'd prefer to say `SHOULD` for ops keys only, and in `ion` perhaps raise it to a `SHALL`. But if the constraint applies to non-ops keys, it just doesn't make sense.

csuwildcat:
The SHOULD directive here is about preferring registered cryptographic suites over custom ones, that's all. I think folks should definitely do so when an existing, registered, standard suite exists - don't you agree?

jmandel:
Sure, if suites exist that actually meet the need. Today I don't think they do (or I'm having trouble finding them).

In any case, as long as the text is relaxed to SHOULD and the reference implementation follows suit, I'm covered.

docs/spec/patches.md
```diff
@@ -52,7 +52,7 @@ The `add-public-keys` _Patch Action_ describes the addition of cryptographic key
...
-    2. The object ****MUST**** include a `type` property, and its value ****MUST**** be the identifier type string of a registered [Cryptographic Suite](https://w3c-ccg.github.io/ld-cryptosuite-registry/) that supports the `publicKeyJwk` representation (examples below). If the value is not a registered type, or of a type that does not support the `publicKeyJwk` representation, the entire _Patch Action_ ****MUST**** be discarded, without any of it being used to modify the DID's state.
+    2. The object ****MUST**** include a `type` property, and its value ****SHOULD**** be the identifier type string of a registered [Cryptographic Suite](https://w3c-ccg.github.io/ld-cryptosuite-registry/) that supports the `publicKeyJwk` representation (examples below). If the key usages include `ops` and the `type` value is not a registered type, indicates a type that does not support the `publicKeyJwk` representation, the entire _Patch Action_ ****MUST**** be discarded, without any of it being used to modify the DID's state.
```
csuwildcat, line 55:
```suggestion
    2. The object ****MUST**** include a type property, and its value ****SHOULD**** be the identifier type string of a registered [Cryptographic Suite](https://w3c-ccg.github.io/ld-cryptosuite-registry/) that supports JWK representations - for example:
```

PR #728, opened 2020-06-05
make update use update key as commit reveal

isaacJChen:
- update now uses the same commit reveal scheme as recover
- ops keys removed


lib/core/versions/latest/ErrorCode.ts
```diff
@@ -51,7 +51,6 @@ export default {
   DocumentComposerIdTooLong: 'document_composer_id_too_long',
   DocumentComposerInvalidSignature: 'document_composer_invalid_signature',
   DocumentComposerKeyNotFound: 'document_composer_key_not_found',
```
thehenrytsai, line 53: Sanity: should be removed?

tests/core/RequestHandler.spec.ts
```diff
@@ -234,8 +230,9 @@ describe('RequestHandler', () => {
     const [, anySigningPrivateKey] = await Jwk.generateEs256kKeyPair();
     const [, anyNextUpdateCommitmentHash] = OperationGenerator.generateCommitRevealPair();
```
thehenrytsai, line 231: Sanity: I think `OperationGenerator.generateCommitRevealPair()` should be removed completely with this change?

tests/core/Resolver.spec.ts
```diff
@@ -110,13 +108,12 @@ describe('Resolver', () => {
...
 
       // Create the recover operation and insert it to the operation store.
-      const [update1RevealValueAfterRecovery, update1CommitmentHashAfterRecovery] = OperationGenerator.generateCommitRevealPair();
       const recoverOperationJson = await OperationGenerator.generateRecoverOperationRequest(
         didUniqueSuffix,
         recoveryPrivateKey,
         newRecoveryPublicKey,
         newSigningPublicKey,
-        update1CommitmentHashAfterRecovery,
+        Multihash.canonicalizeThenHashThenEncode(newSigningPublicKey.jwk),
```
thehenrytsai, line 116: Sanity: this parameter should be removed since `newSigningPublicKey` is passed in, this should be computed internally.

tests/core/MongoDbOperationStore.spec.ts
```diff
@@ -1,10 +1,12 @@
 import AnchoredOperationModel from '../../lib/core/models/AnchoredOperationModel';
 import IOperationStore from '../../lib/core/interfaces/IOperationStore';
-import JwkEs256k from '../../lib/core/models/JwkEs256k';
 import MongoDb from '../common/MongoDb';
 import MongoDbOperationStore from '../../lib/core/MongoDbOperationStore';
 import OperationGenerator from '../generators/OperationGenerator';
 import UpdateOperation from '../../lib/core/versions/latest/UpdateOperation';
+import PublicKeyModel from '../../lib/core/versions/latest/models/PublicKeyModel';
+import Multihash from '../../lib/core/versions/latest/Multihash';
+import JwkEs256k from '../../lib/core/models/JwkEs256k';
```
thehenrytsai, line 9: Nit: sort.

tests/core/MongoDbOperationStore.spec.ts
```diff
@@ -21,36 +23,36 @@ async function createOperationStore (mongoDbConnectionString: string): Promise<I
...
+    const [newPublicKey, newPrivateKey] = await OperationGenerator.generateKeyPair(`key${i}`);
     const operationRequest = await OperationGenerator.createUpdateOperationRequestForAddingAKey(
       didUniqueSuffix,
-      updateRevealValueEncodedString,
+      currentPublicKey.jwk,
       newPublicKey,
-      nextUpdateCommitmentHash,
-      signingKeyId,
-      signingPrivateKey
+      Multihash.canonicalizeThenHashThenEncode(newPublicKey.jwk),
```
thehenrytsai, line 45: Sanity: no need to pass this parameter as new public key is passed in.

tests/generators/OperationGenerator.ts
```diff
@@ -323,7 +317,7 @@ export default class OperationGenerator {
    */
   public static async createUpdateOperationRequest (
     didUniqueSuffix: string,
-    updateRevealValue: string,
+    updateKey: JwkEs256k,
     nextUpdateCommitmentHash: string,
     patches: any,
     signingKeyId: string,
```
thehenrytsai, line 323: Sanity: I thought you mentioned you were going to remove the key ID? Which I think is the right thing.

tests/generators/OperationGenerator.ts
```diff
@@ -338,7 +332,7 @@ export default class OperationGenerator {
     const encodedDeltaString = Encoder.encode(deltaJsonString);
 
     const signedDataPayloadObject = {
-      update_reveal_value: updateRevealValue,
+      update_key: updateKey,
       delta_hash: deltaHash
     };
     const signedData = await OperationGenerator.signUsingEs256k(signedDataPayloadObject, signingPrivateKey, signingKeyId);
```
thehenrytsai, line 338: Consider just changing method signature to just not allow key ID for `OperationGenerator.signUsingEs256k()`

lib/core/versions/latest/UpdateOperation.ts
```diff
@@ -2,6 +2,8 @@ import DeltaModel from './models/DeltaModel';
 import Encoder from './Encoder';
 import ErrorCode from './ErrorCode';
 import JsonAsync from './util/JsonAsync';
+import Jwk from './util/Jwk';
+import JwkEs256k from '../../models/JwkEs256k';
 import Jws from './util/Jws';
 import Multihash from './Multihash';
 import Operation from './Operation';
```
thehenrytsai, line 9: Search for `const expectKidInHeader = true;` in this file, it needs to be `false` now to be consistent with other operations. Once this is done, you'll probably need to update `Jws` class and related method to ensure it's code coverage remains at 100%.

tests/generators/OperationGenerator.ts
```diff
@@ -462,7 +454,7 @@ export default class OperationGenerator {
    */
   public static async createUpdateOperationRequestForAddingAKey (
     didUniqueSuffix: string,
-    updateRevealValue: string,
+    updateKey: JwkEs256k,
```
thehenrytsai, line 457: Consider not requiring to pass the public key, since private key is passed already you can use `getEs256kPublicKey()` internally.. The goal being trying to make the method as easy to use as possible.

tests/core/OperationProcessor.spec.ts
```diff
@@ -21,20 +21,24 @@ import RecoverOperation from '../../lib/core/versions/latest/RecoverOperation';
...
 import UpdateOperation from '../../lib/core/versions/latest/UpdateOperation';
+// import JsonCanonicalizer from '../../lib/core/versions/latest/util/JsonCanonicalizer';
 
 async function createUpdateSequence (
   didUniqueSuffix: string,
   createOp: AnchoredOperationModel,
-  firstUpdateRevealValue: string,
+  firstUpdateKey: JwkEs256k,
   numberOfUpdates: number,
   publicKeyId: string,
```
thehenrytsai, line 31: Sanity: remove.

tests/core/MongoDbOperationStore.spec.ts
isaacJChen: We need this because the added key could be another key it's just for testing we pass in the same thing for convenience. (IRL this should never happen because then you are exposing your public key)

tests/generators/OperationGenerator.ts
isaacJChen: We need it for some tests that test when the key is invalid

tests/core/MongoDbOperationStore.spec.ts
thehenrytsai: You are right that the key to add should be separated from the newUpdateKey, I was not thinking clearly. But in which case, i think we should pass in a JWK to this generator method.

Optional: I personally would even prefer to generate and pass an entirely different key, but this would be optional as I understand that adds more burden to the PR.

tests/generators/OperationGenerator.ts
thehenrytsai: Agree based on the principal that it is fine for test libraries to be more flexible and less usable.

tests/core/MongoDbOperationStore.spec.ts
thehenrytsai: Agree based on the principal that it is fine for test libraries to be more flexible and less usable.

tests/generators/OperationGenerator.ts
```diff
@@ -241,7 +222,6 @@ export default class OperationGenerator {
   public static async generateCreateOperationRequest (
     recoveryPublicKey: JwkEs256k,
     signingPublicKey: PublicKeyModel,
```
thehenrytsai, line 224: To your point about non-operation keys should be separate from operation keys, I think we should add `updatePublicKey` between line 223 and 224 to disambiguate.

isaacJChen: Good catch.

tests/core/MongoDbOperationStore.spec.ts
```diff
@@ -140,17 +139,15 @@ describe('MongoDbOperationStore', async () => {
...
-    const [, anyUnusedCommitmentHash] = OperationGenerator.generateCommitRevealPair();
 
     // Generate an update operation.
     const operationRequest = await OperationGenerator.createUpdateOperationRequestForHubEndpoints(
       didUniqueSuffix,
-      'anyUnusedUpdateRevealValue',
-      anyUnusedCommitmentHash,
+      createOperationData.signingPublicKey.jwk,
+      createOperationData.signingPrivateKey,
+      Multihash.canonicalizeThenHashThenEncode({}),
```
thehenrytsai, line 148: Consider using `generateRandomHash()`.

tests/core/OperationProcessor.spec.ts
```diff
@@ -191,11 +186,10 @@ describe('OperationProcessor', async () => {
     const nextUpdateCommitmentHash = 'EiD_UnusedNextUpdateCommitmentHash_AAAAAAAAAAA';
     const updateOperationRequest = await OperationGenerator.createUpdateOperationRequest(
       didUniqueSuffix,
-      firstUpdateRevealValue,
+      signingPublicKey.jwk,
+      signingPrivateKey,
```
thehenrytsai, line 190: I think these should be named updatePublic/PrivateKey now. But if it's not a simple rename, then forget it!

isaacJChen: https://github.com/decentralized-identity/sidetree/issues/730

Issue #732, opened 2020-06-08
Rename `usage` to `purpose` for public keys


thehenrytsai:
It is recommended to us by downstream user that `purpose` is a better property name, and it also aligns with DID spec better.

thehenrytsai:
Implemented.

PR #733, opened 2020-06-08
#732 - Renamed `usage` to `purpose` for public keys

thehenrytsai:


lib/core/versions/latest/models/PublicKeyModel.ts
```diff
@@ -5,5 +5,5 @@ export default interface PublicKeyModel {
   id: string;
   type: string;
   jwk: any;
-  usage: string[];
+  purpose: string[];
```
mudiali, line 8: Should the type be `PublicKeyPurpose`?

thehenrytsai: No, this is external facing, thus needs to be identical to what the spec says. I will update line 2 to describe it better!

thehenrytsai: Done.

PR #736, opened 2020-06-09
Handle the initial case where no keys are present in a Document

jmandel:
I've been running into an error with a `create` operation that invokes `add-public-keys` patches: the DocumentComposer fails because it tries to treat a missing `public_keys` property as an array.

This adds an explicit interface-level indication that `public_keys` may be absent, and adds a test to ensure that `add-public-keys` works against an empty doc.

### DIFF
--- a/lib/core/versions/0.9.0/models/PublicKeyModel.ts
+++ b/lib/core/versions/0.9.0/models/PublicKeyModel.ts
@@ -1,4 +1,4 @@
-import PublicKeyPurpose from "../PublicKeyPurpose";
+import PublicKeyPurpose from '../PublicKeyPurpose';
 
 /**
  * Data model representing a public key in the 'publicKey' array in patches.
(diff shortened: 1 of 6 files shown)


lib/core/versions/0.9.0/models/PublicKeyModel.ts
```diff
@@ -1,4 +1,4 @@
-import PublicKeyPurpose from "../PublicKeyPurpose";
+import PublicKeyPurpose from '../PublicKeyPurpose';
```
jmandel, line 1: This was a pre-existing lint error fixed by `npm run lint`

lib/core/versions/latest/models/PublicKeyModel.ts
```diff
@@ -1,4 +1,4 @@
-import PublicKeyPurpose from "../PublicKeyPurpose";
+import PublicKeyPurpose from '../PublicKeyPurpose';
```
jmandel, line 1: This was a pre-existing lint error fixed by npm run lint.

lib/core/versions/0.9.0/models/PublicKeyModel.ts
thehenrytsai: Thank you!

lib/core/versions/latest/models/PublicKeyModel.ts
thehenrytsai: Thank you!

PR #761, opened 2020-06-23
raw data parser working and tested

isaacJChen:


lib/bitcoin/BitcoinRawDataParser.ts
```diff
@@ -0,0 +1,85 @@
...
+  };
+  private static sizeBytesLength = 4;
+
+  /**
+   * Parse the given raw block data file. It can throw error if block data is invalid when validating magic bytes,
+   * creating new Block, or validating size
+   * @param rawBlockDataFileBuffer The file, in buffer form, to be parsed as blocks
+   */
+  public static parseRawDataFile (rawBlockDataFileBuffer: Buffer): any {
+    let rawBlockDataFileString = rawBlockDataFileBuffer.toString('hex');
```
thehenrytsai, line 24:
```suggestion
    const rawBlockDataFileString = rawBlockDataFileBuffer.toString('hex');
```

lib/bitcoin/BitcoinRawDataParser.ts
```diff
@@ -0,0 +1,85 @@
...
+
+    // split the hex by magic bytes
+    if (rawBlockDataFileString.startsWith(BitcoinRawDataParser.magicBytes.testnet)) {
+      magicBytes = BitcoinRawDataParser.magicBytes.testnet;
+      hexStrings = rawBlockDataFileString.split(magicBytes);
+    } else if (rawBlockDataFileString.startsWith(BitcoinRawDataParser.magicBytes.mainnet)) {
+      magicBytes = BitcoinRawDataParser.magicBytes.mainnet;
+      hexStrings = rawBlockDataFileString.split(magicBytes);
+    } else {
+      throw new Error('Invalid block data');
```
thehenrytsai, line 37: Use SidetreeError.

lib/bitcoin/BitcoinRawDataParser.ts
```diff
@@ -0,0 +1,85 @@
...
+      if (currentHex === '') {
+        currentHex = hexString;
+      } else {
+        // take care of when non magic bytes get split. Add back the bytes and concat
+        currentHex = `${currentHex}${magicBytes}${hexString}`;
+      }
+      if (BitcoinRawDataParser.verifySize(currentHex)) {
+        // A hex string can be treated as raw block data once size is verified and size bytes removed
+        const rawBlockDataAsString = currentHex.slice(BitcoinRawDataParser.sizeBytesLength * 2);
+        const rawBlockDataAsBuffer = Buffer.from(rawBlockDataAsString, 'hex');
```
thehenrytsai, line 55: Design consideration: is it possible work on Buffer as opposed to HEX string? That would avoid code that converts to HEX, then back to buffer again, the * 2 etc.

lib/bitcoin/BitcoinRawDataParser.ts
```diff
@@ -0,0 +1,85 @@
...
+   */
+  private static magicBytes = {
+    testnet: '0b110907',
+    mainnet: 'f9beb4d9'
+  };
+  private static sizeBytesLength = 4;
+
+  /**
+   * Parse the given raw block data file. It can throw error if block data is invalid when validating magic bytes,
+   * creating new Block, or validating size
```
thehenrytsai, line 20: Can you add here a quick digest of example what the block data look like? I am not sure I understand the format looking at the loop...

e.g. <magic bytes><block1 size><block1 data><block2 size><block2 data>...

lib/bitcoin/BitcoinRawDataParser.ts
isaacJChen: Good point. Will switch to operating with buffers. I used string thinking would be easier to read.

lib/bitcoin/BitcoinRawDataParser.ts
isaacJChen: Ya I will add documentation. For your refernce now, this is the block data 
https://user-images.githubusercontent.com/26752423/76470914-8c729400-63ae-11ea-882f-c5919ee78305.png

lib/bitcoin/BitcoinRawDataParser.ts
```diff
@@ -0,0 +1,85 @@
...
+    } else {
+      throw new Error('Invalid block data');
+    }
+
+    // remove the first empty string after split
+    hexStrings.shift();
+
+    let currentHex = '';
+    let count = 0;
+    for (const hexString of hexStrings) {
```
thehenrytsai, line 45: My head hurts thinking initially `hexString` can be magic bytes, size, and block data, but realizing it can't be magic bytes.

Is it possible to sorta have each increment of the loop to operates at the [currentHex, magic bytes, block] triple increment? e.g.

foreach (block starting point) {
   magic bytes = ...
   size = ...
   block = ...
   verifySize()...
   new Block()?
}

PR #762, opened 2020-06-23
Added code to abort decompression once a limit is reached

thehenrytsai:


lib/core/versions/latest/util/Compressor.ts
```diff
@@ -1,13 +1,17 @@
...
+import SidetreeError from '../../../../common/SidetreeError';
+import ErrorCode from '../ErrorCode';
 
 /**
  * Encapsulates functionality to compress/decompress data.
  */
 export default class Compressor {
 
+  /** The estimated ratio/multiplier of decompressed Sidetree CAS file size compared against the compressed file size. */
+  public static readonly estimatedDecomporessionMultiplier = 3;
```
thehenrytsai, line 12: Open to suggestion on:
1. Alternative location where this constant is stored.
1. What the value should be. I chose 3 because I validated that the compression ratio is very close to 2:1 so setting to 3 means such limit would never be hit, thus resulting: a) this being more of an implementation detail rather than spec requirement, and b) equivalent check on batch writer optional/low-pri.

tests/core/util/Compressor.spec.ts
```diff
@@ -8,8 +10,50 @@ describe('Compressor', async () => {
...
+        error.code === ErrorCode.CompressorMaxAllowedDecompressedDataSizeExceeded) {
+
+        // Further check the error message to ensure that the decompressed bytes are less than the fully decompressed data size,
+        // to ensure that "chunking" is taking place.
+
+        // Parsing out the bytes decompressed from error message.
+        // NOTE: Error message looks like: 'Max data size allowed: 100000 bytes, aborted decompression at 114688 bytes.`
+        let message = error.message;
+        message = message.substring(0, message.length - 6); // Removing the trailing 'bytes.' string.
+        const bytesDecompressed = Number.parseInt(message.substring(message.indexOf('at ') + 3), 10); // Base 10.
```
mudiali, line 43: Consider: Maybe it is better to have a new Error class e.g. `SidetreeCompressorError extends SidetreeError` which takes in a `code` and `decompressedBytes` values so that we don't have to depend upon the error string in the correct format. Or the new class could be `MaxAllowedDecompressSizeReached extends SidetreeError` which calls base w/ the new error code and again exposes a `decompressedBytes' property.

lib/core/versions/latest/util/Compressor.ts
```diff
@@ -1,13 +1,17 @@
 import util = require('util');
 import zlib = require('zlib');
+import SidetreeError from '../../../../common/SidetreeError';
```
mudiali, line 3: Sort plz.

lib/core/versions/latest/AnchorFile.ts
```diff
@@ -36,7 +36,8 @@ export default class AnchorFile {
 
     let anchorFileDecompressedBuffer;
     try {
-      anchorFileDecompressedBuffer = await Compressor.decompress(anchorFileBuffer);
+      const estimatedDecompressedFileSize = ProtocolParameters.maxAnchorFileSizeInBytes * Compressor.estimatedDecomporessionMultiplier;
```
mudiali, line 39: Nit: maybe update the variable name to be `estimatedDecompressedFileSizeInBytes`? Same for the other uses.

lib/core/versions/latest/util/Compressor.ts
mudiali: 1. OK with the location.
2. Is there a scenario where someone can force a node to save the max-sized decompressed file? I don't think that is the case as we validate the file contents before saving it. Calling this out as in that case we are giving a spammer more room.

thehenrytsai: Regarding 2: You are right, we don't save decompressed file itself. This does give spammers more temporary memory to decompress into, e.g. 20MB vs 30MB, but I think it's okay because memory is released after decompression is done.

lib/core/versions/latest/AnchorFile.ts
thehenrytsai: Renamed to `maxAllowedDecompressedSizeInBytes`.

lib/core/versions/0.9.0/util/Compressor.ts
```diff
@@ -23,13 +27,44 @@ export default class Compressor {
...
+    // Create a gunzip transform object.
+    const gunzip = zlib.createGunzip();
 
-    const result = await Compressor.gUnzipAsync(inputAsBuffer);
+    let content = Buffer.alloc(0);
 
-    // Casting result to Buffer as that's what is returned by gzip
-    return result as Buffer;
+    // Handle the data chunks that are decompressed as they come in.
+    gunzip.on('data', (chunk: Buffer) => {
```
isaacJChen, line 38: Do we have to worry about chunk.length being huge? Or is there a default size for chunk that gunzip controls

thehenrytsai: There is a default chunk size 16KB, and I added a test for this exact concern.

tests/core/util/Compressor.spec.ts
thehenrytsai: Synced offline. Decided to not adding a new Error class.

PR #765, opened 2020-06-23
Ische/block file reader

isaacJChen:


lib/bitcoin/BitcoinFileReader.ts
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * concrete implementation of BitcoinFileReader
+ */
+export default class BitcoinFileReader implements IBitcoinFileReader {
+  /**
+   * Constructor
+   * @param bitcoinDataDirectory The same directory used by bitcoin core
+   */
+  public constructor (public bitcoinDataDirectory: string) {}
```
thehenrytsai, line 12: nit:
```suggestion
  public constructor (private bitcoinDataDirectory: string) {}
```

lib/bitcoin/BitcoinFileReader.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  public listBlockFiles (): string[] {
+    const blocksDataDirectoryPath = `${this.bitcoinDataDirectory}/blocks`;
+    let blockDataDir;
+    try {
+      blockDataDir = fs.readdirSync(blocksDataDirectoryPath);
+    } catch (e) {
+      console.error(`Error thrown while reading file system: ${e}`);
+      return [];
```
thehenrytsai, line 21: Sanity: wouldn't return [] potentially allow initialization code silently thinking all files are processed? Of course unless it checks for [] explicitly and fail, but then we can do that here?

lib/bitcoin/interfaces/IBitcoinFileReader.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * Reader for bitcoin files from bitcoin data directory
+ */
+export default interface IBitcoinFileReader {
```
thehenrytsai, line 4: For my understanding, what is the motivation for creating an interface here? For testing? Or are you anticipating a different implementation? I was under the impression that the same implementation would still be used even if we use Azure Files?

lib/bitcoin/BitcoinFileReader.ts
```diff
@@ -0,0 +1,35 @@
...
+    const blockFileList = blockDataDir.filter((fileName) => { return fileName.startsWith('blk'); });
+    return blockFileList;
+  }
+
+  public readBlockFile (fileName: string): Buffer | undefined {
+    try {
+      return fs.readFileSync(`${this.bitcoinDataDirectory}/blocks/${fileName}`);
+    } catch (e) {
+      console.error(`Error thrown while reading file system: ${e}`);
+      return undefined;
```
thehenrytsai, line 32: Similar to above, it seems to me that throwing an Error when failing to perform the method (read block file) is the right thing to do. Returning `undefined` is probably okay too, but at minimum documentation needs to say so in the interface.

isaacJChen: ok, will throw instead of returning default.

lib/bitcoin/interfaces/IBitcoinFileReader.ts
isaacJChen: Using an interface is optional, mainly created it for future proofing. If someone wants to use a reader that doesn't read from local file system, they should be able to implement this interface and have it working.

lib/bitcoin/BitcoinFileReader.ts
isaacJChen: Will throw instead of return default.

lib/bitcoin/BitcoinFileReader.ts
```diff
@@ -9,27 +11,29 @@ export default class BitcoinFileReader implements IBitcoinFileReader {
...
     const blockFileList = blockDataDir.filter((fileName) => { return fileName.startsWith('blk'); });
     return blockFileList;
   }
 
-  public readBlockFile (fileName: string): Buffer | undefined {
+  public readBlockFile (fileName: string): Buffer {
     try {
       return fs.readFileSync(`${this.bitcoinDataDirectory}/blocks/${fileName}`);
     } catch (e) {
+      // this means file does not exist
```
thehenrytsai, line 34: Sanity: Wouldn't it possible for other error conditions?

Issue #769, opened 2020-06-23
Align Sidetree DID Parameters with DID Core Spec


OR13:
https://github.com/w3c/did-core/issues/36

When this gets closed, we should address it.

csuwildcat:
Seems like we would just register `initial-state` as a generic official parameter, given what I read there - right?

OR13:
correct.... sometimes it takes months to end up right back where you started :)

csuwildcat:
![4945415c-e75f-4724-ae39-c5c4813dbedc](https://user-images.githubusercontent.com/131786/85597011-16153d00-b5ff-11ea-85cf-f44ac018607f.jpg)

csuwildcat:
With the move to a fully qualified DID URI for long-form, this is no longer applicable.

PR #770, opened 2020-06-25
Remove method prefix from initial state

OR13:
addresses https://github.com/decentralized-identity/sidetree/issues/769

Removes the method prefix from `initial-state` did parameter.


jmandel:
I missed a window to comment since it looks this was created and merged within an hour, but this is a breaking change to long-form DIDs. I thought our intention was to avoid making breaking changes to long-form DIDs at this stage. The PR here and issue at https://github.com/decentralized-identity/sidetree/issues/769 don't explain the rationale.

If I'm missing something, please let me know. (I'd prefer to see no change, or a spec that ensures `-ion-initial-state` remains acceptable even if `initial-state` is preferred.)

csuwildcat:
@jmandel we can make this a backwards compatible change in the ION implementation

jmandel:
Thanks! Is there a spot where we document these kinds of additional rules for `ion`?

csuwildcat:
@jmandel I will get together an ION spec that is mostly refs to Sidetree, but with the bits that are specific to it. That's one of my deliverables for July

PR #774, opened 2020-06-25
Proposed release process for sidetree reference implementation

tplooker:
The following PR is a proposed CI assisted release process for the sidetree reference implementation.

**Note** - This needs to be tested and the repository needs a `NPM_TOKEN` configured into the secrets of the repository so that github has the ability to publish a package to npm.

Refer to ./docs/release-process.md for the proposed outlined process.

### DIFF
--- a/.github/workflows/any-pr.yml
+++ b/.github/workflows/any-pr.yml
@@ -1,9 +1,10 @@
-name: CI
+name: any-pr
 
-on: [push, pull_request]
+on: [pull_request]
 
 jobs:
-  build:
+  any-pr:
+    name: Build and test
     runs-on: ubuntu-latest
     strategy:
       matrix:
--- a/.github/workflows/push-master.yml
+++ b/.github/workflows/push-master.yml
@@ -1,0 +1,31 @@
+name: any-pr
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  push-master:
+    # Don't run on any commit that features [skip ci] in the commit message
+    if: "! contains(github.event.head_commit.message, '[skip ci]')"
+    name: Build, test and publish unstable release
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [10.15.3]
+    steps:
+    - uses: actions/checkout@v2
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+    - run: npm i
+    - run: npm run test
+    - run: echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" > .npmrc 2> /dev/null
+        env:
+          NPM_TOKEN: ${{ secrets.NPMJS_TOKEN }}
+    - run: git config user.name "DIF CI"
+    - run: git config user.email "ci@identity.foundation"
+    - run: npm whoami
+    - run: npm run publish:unstable
--- a/.github/workflows/push-release.yml
+++ b/.github/workflows/push-release.yml
@@ -1,0 +1,31 @@
+name: any-pr
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  push-master:
+    # Only run on any commit that features the release commit message syntax in the commit message
+    if: "contains(github.event.head_commit.message, 'release commit')"
+    name: Build, test and publish stable release
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [10.15.3]
+    steps:
+    - uses: actions/checkout@v2
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+    - run: npm i
+    - run: npm run test
+    - run: echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" > .npmrc 2> /dev/null
+        env:
+          NPM_TOKEN: ${{ secrets.NPMJS_TOKEN }}
+    - run: git config user.name "DIF CI"
+    - run: git config user.email "ci@identity.foundation"
+    - run: npm whoami
+    - run: npm run publish:release
--- a/docs/release-process.md
+++ b/docs/release-process.md
@@ -19,7 +19,7 @@ To create a stable release follow the following steps
 3. Install the dependencies `npm install`
 5. Build the package `npm build`
 6. Test the package `npm test`
-7. Run `npm run version:release`, note by default this will do a minor package release as we are pre the `1.0.0` release
+7. Run `npm run version:release`, with an appropriate option such as [ `major` | `minor` | `patch` ].
 8. Observe the correctly incremented change to the `package.json`
 9. Push the release branch including the newly created tags `git push origin release --tags`
 10. Open a pull request for the release, once approvals have been sought, merge the pull request using **rebase**,
--- a/docs/release-process.md
+++ b/docs/release-process.md
@@ -24,7 +24,7 @@ To create a stable release follow the following steps
 9. Push the release branch including the newly created tags `git push origin release --tags`
 10. Open a pull request for the release, once approvals have been sought, merge the pull request using **rebase**,
     preserving the commit message as `release commit [skip ci]`
-11. Observe the triggering of the `/.github/workflows/push-release.yaml` github workflow
+11. Observe the triggering of the `/.github/workflows/push-release.yml` github workflow
 
 **Note** It is important that **rebase** is used as the strategy for merging a release pull request as this preserves the created release tag.
 
--- a/docs/release-process.md
+++ b/docs/release-process.md
@@ -1,0 +1,46 @@
+# Release Process
+
+The following document covers the release process for the artifacts in this repository.
+
+## Reference Implementation
+
+The following details the release process of the Sidetree reference implementation `@decentralized-identity/sidetree`, which is a typescript based [npm](https://www.npmjs.com/) package.
+
+In general the release process can be summarized by the following flow
+- Each merged pull request into master results in the automated release of a new unstable package on npm, unless the `[skip ci]` tag is included in the commit message.
+- A release commit merged into master triggers a stable release of the reference implementation.
+
+### Stable Releases
+
+To create a stable release follow the following steps
+
+1. Checkout the head of master `git checkout master && git pull`
+2. Create a new release branch from master called `release`
+3. Install the dependencies `npm install`
+5. Build the package `npm build`
+6. Test the package `npm test`
+7. Run `npm run version:release`, note by default this will do a minor package release as we are pre the `1.0.0` release
+8. Observe the correctly incremented change to the `package.json`
+9. Push the release branch including the newly created tags `git push origin release --tags`
+10. Open a pull request for the release, once approvals have been sought, merge the pull request using **rebase**,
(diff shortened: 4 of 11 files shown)


.github/workflows/push-release.yml
```diff
@@ -0,0 +1,31 @@
...
+    - uses: actions/checkout@v2
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+    - run: npm i
+    - run: npm run test
+    - run: echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" > .npmrc 2> /dev/null
+        env:
+          NPM_TOKEN: ${{ secrets.NPMJS_TOKEN }}
```
OR13, line 27: @thehenrytsai @csuwildcat can one of you add this to the repo secrets?

OR13:
@tplooker this looks good, when do you think it should be converted from draft?

tplooker:
> @tplooker this looks good, when do you think it should be converted from draft? Are there additional changes you want to make?

Happy to mark it out of draft if its generally supported as a go forward approach, I would like to test it prior to merging though

.github/workflows/any-pr.yml
```diff
@@ -1,9 +1,10 @@
-name: CI
+name: any-pr
```
gjgd, line 1: The name should describe what the action does, not when it is triggered IMO.
For example: "Continous Integration" or "Run tests"

.github/workflows/push-master.yml
```diff
@@ -0,0 +1,31 @@
+name: any-pr
```
gjgd, line 1: Same comment -> "Publish unstable release to NPM"

.github/workflows/push-master.yml
```diff
@@ -0,0 +1,31 @@
...
+      with:
+        node-version: ${{ matrix.node-version }}
+    - run: npm i
+    - run: npm run test
+    - run: echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" > .npmrc 2> /dev/null
+        env:
+          NPM_TOKEN: ${{ secrets.NPMJS_TOKEN }}
+    - run: git config user.name "DIF CI"
+    - run: git config user.email "ci@identity.foundation"
+    - run: npm whoami
```
gjgd, line 30: Not needed

.github/workflows/push-release.yml
```diff
@@ -0,0 +1,31 @@
...
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  push-master:
+    # Only run on any commit that features the release commit message syntax in the commit message
+    if: "contains(github.event.head_commit.message, 'release commit')"
```
gjgd, line 11: Note that when a "release commit" is pushed to master, both the unstable and the stable version will be published (unless "skip ci" is also present) Is that the desired behavior?

.github/workflows/push-release.yml
```diff
@@ -0,0 +1,31 @@
+name: any-pr
```
gjgd, line 1: Same comment -> "Publish stable release to NPM"

.github/workflows/push-release.yml
```diff
@@ -0,0 +1,31 @@
...
+      with:
+        node-version: ${{ matrix.node-version }}
+    - run: npm i
+    - run: npm run test
+    - run: echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" > .npmrc 2> /dev/null
+        env:
+          NPM_TOKEN: ${{ secrets.NPMJS_TOKEN }}
+    - run: git config user.name "DIF CI"
+    - run: git config user.email "ci@identity.foundation"
+    - run: npm whoami
```
gjgd, line 30: Not needed

docs/release-process.md
```diff
@@ -0,0 +1,46 @@
...
+2. Create a new release branch from master called `release`
+3. Install the dependencies `npm install`
+5. Build the package `npm build`
+6. Test the package `npm test`
+7. Run `npm run version:release`, note by default this will do a minor package release as we are pre the `1.0.0` release
+8. Observe the correctly incremented change to the `package.json`
+9. Push the release branch including the newly created tags `git push origin release --tags`
+10. Open a pull request for the release, once approvals have been sought, merge the pull request using **rebase**,
+    preserving the commit message as `release commit [skip ci]`
+11. Observe the triggering of the `/.github/workflows/push-release.yaml` github workflow
```
gjgd, line 27:
```suggestion
11. Observe the triggering of the `/.github/workflows/push-release.yml` github workflow
```

package.json
```diff
@@ -46,6 +46,9 @@
   "scripts": {
     "build": "tsc && copyfiles \"lib/**/*.json\" dist && copyfiles \"tests/**/*.js*\" dist && copyfiles \"package.json\" dist",
     "test": "npm run build && nyc jasmine --config=./tests/jasmine.json",
+    "publish:unstable": "./scripts/publish-unstable.sh",
+    "publish:release": "./scripts/publish.sh",
+    "version:release": "yarn version --minor --message \"release commit [skip ci]\"",
```
gjgd, line 51: Not a fan of the use of both yarn and npm in the same project.

gjgd:
@tplooker  "skip ci" and "release commit" feel like tags rather than commit messages. Have you considered using git tags instead ? 

For example:
- An untagged commit to master does not trigger any push to npm registry
- A commit tagged "v1.2.3-UNSTABLE" triggers the unstable workflow
- A commit tagged "v1.2.3-STABLE" triggers the stable workflow

.github/workflows/any-pr.yml
OR13: I noticed this as well... I can see some advantage to a a single workflow per "when it is triggered", but seems like https://help.github.com/en/actions/reference/events-that-trigger-workflows

its probably a best practice to have the workflow functionality named, and doing one thing, and then have multiple workflows run `on: [push, pull_request]`, etc...

tplooker: Happy to rename however I sort of disagree, because in this workflow setup where there are not multiple actions per GH workflow it tends to provide the right level of detail to the developer in the name, rather than a general term like CI that most developers have a different definition of.

How about this as a proposal

any-pr => ci
push-master => cd - importantly the d here means delivery not deployment
push-release => release

.github/workflows/push-master.yml
tplooker: With this proposal would you have the file name? publish-unstable-release-to-npm?

.github/workflows/push-release.yml
tplooker: Correct and see the build script for the commit message that is created when a release commit is made `[skip ci]` is included

.github/workflows/push-release.yml
tplooker: Commented above

package.json
tplooker: Ah nice catch sorry carried over from another project where we use yarn, def should not have both being used here. TBH I would rather see this project converted to using yarn

tplooker:
@gjgd the reason I like having something in the commit message for skipping CI is if a PR is made that say just amends the docs and I don't want another unstable release triggered on merge into master, amending the commit message when merging the PR is easy whereas using a tag would be more of a pain?

w.r.t release commit we can shift to using a release tag, however there still has to be a release commit that bumps the version, changelog and hence should have a consistent name so that someone reviewing the history of master can quickly see the release commits, so why not use this as the convention for triggering the release?

In your workflow, would the PR submitter not have to push a tag to the repo every time we want a new unstable release (i.e alongside their PR)? This just feels like an extra step that many developers would be liable to forget?

.github/workflows/push-master.yml
tplooker: Agree it is not strictly needed, however it gives a clean failure if login to the NPM registry fails, rather than kicking this to failing during publishing the package to npm, happy to remove though

.github/workflows/push-release.yml
tplooker: To be clear I would like to see us move to more standard commit message formats anyway e.g see #775 which would shift this convention to be more like `chore(release): publish [skip ci]`

.github/workflows/push-release.yml
tplooker: Commented above

.github/workflows/any-pr.yml
OR13: these seem more in line with what I would expect, but as you note... if there are only ever going to be these workflows the file name is not that important... what is important is what shows up on the PR Preview, when things get more complicated.... for example:

Screen Shot 2020-06-27 at 2 52 22 PM

... I hope most developers think that CI = Lint + Tests + Build, but I guess that assumption might not be safe.

OR13: Its not a hill I would die on ever, just a comment :)

.github/workflows/push-master.yml
OR13: perhaps a comment to that effect would be wise to prevent continued questioning :)

OR13:
I can see the value in using the commit structure to drive the behavior of release, particularly in a world where you _never_ release from a local setup.

OR13:
@tplooker @csuwildcat @thehenrytsai @codeglobally who has the NPM access token necessary to complete this work, and what stops us from merging these workflow changes as is and letting the workflows fail until the appropriate party adds them to github secrets :)

.github/workflows/push-master.yml
OR13: its thats what the action is going, thats what I would name it.

I tend to prefer actions do 1 thing, and make them very generic so they can be copied easily without modification.

.github/workflows/any-pr.yml
tplooker: Have updated now, let me know what you guys think @gjgd @OR13

package.json
```diff
@@ -46,6 +46,9 @@
   "scripts": {
     "build": "tsc && copyfiles \"lib/**/*.json\" dist && copyfiles \"tests/**/*.js*\" dist && copyfiles \"package.json\" dist",
     "test": "npm run build && nyc jasmine --config=./tests/jasmine.json",
+    "publish:unstable": "./scripts/publish-unstable.sh",
+    "publish:release": "./scripts/publish.sh",
+    "version:release": "npm version --minor --message \"release commit [skip ci]\"",
```
tplooker, line 51: TODO need to drop this `--minor flag`

.github/workflows/push-release.yml
tplooker: Blocked from merging this PR until this secret is added

docs/release-process.md
```diff
@@ -0,0 +1,46 @@
...
+### Stable Releases
+
+To create a stable release follow the following steps
+
+1. Checkout the head of master `git checkout master && git pull`
+2. Create a new release branch from master called `release`
+3. Install the dependencies `npm install`
+5. Build the package `npm build`
+6. Test the package `npm test`
+7. Run `npm run version:release`, with an appropriate option such as [ `major` | `minor` | `patch` ].
```
OR13, line 22: we need a comment here about duplicating a folder / registering the associated typescript modules.

OR13:
@tplooker  pending tweaks to docs, this will be merged.

docs/release-process.md
tplooker: @thehenrytsai if you get a chance would be great to contribute more to the vague explanation I have given

PR #785, opened 2020-07-02
Add suffix data type

isaacJChen:


lib/core/versions/latest/CreateOperation.ts
```diff
@@ -150,6 +152,27 @@ export default class CreateOperation implements OperationModel {
...
+    if (properties.length === 3) {
+      if (typeof suffixData.type !== 'string') {
+        throw new SidetreeError(ErrorCode.CreateOperationSuffixDataTypeIsNotString);
+      }
+
+      if (suffixData.type.length > 4) {
+        throw new SidetreeError(ErrorCode.CreateOperationSuffixDataTypeLengtGreaterThanFour);
+      }
+
+      if (!/^[A-Za-z0-9_-]+$/.test(suffixData.type)) {
```
thehenrytsai, line 165: Use `Encoder.isBase64UrlString()`

lib/core/versions/latest/CreateOperation.ts
```diff
@@ -150,6 +152,27 @@ export default class CreateOperation implements OperationModel {
...
+        throw new SidetreeError(ErrorCode.CreateOperationSuffixDataTypeInvalidCharacter);
+      }
+
+      return {
+        deltaHash: suffixData.delta_hash,
+        recoveryCommitment: suffixData.recovery_commitment,
+        type: suffixData.type
+      };
```
thehenrytsai, line 173: Nit: I believe technically you don't need to dup the return statement here, adding `type: suffixData.type` in the existing return should be sufficient.

PR #787, opened 2020-07-05
Update swagger spec to valid OAS3 format

tmarkovski:
Signed-off-by: Tomislav Markovski <tmarkovski@gmail.com>

Updated swagger to be valid OAS3 spec. Consolidated error response in a single schema. Moved individual error codes to description.
This allows generators to correctly generate code. In addition, strongly typed languages generally prefer defined models in components, instead inline schema.

I tested this with autorest and NSwag, generated C# and typescript code, it looked fine.


OR13:
@tmarkovski thanks for these awesome changes!

PR #791, opened 2020-07-07
Ische/fast init

isaacJChen:


lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -176,8 +188,130 @@ export default class BitcoinProcessor {
...
+
+    console.log(`Begin fast processing block ${startingBlock.height} to ${bestHeight}`);
+    // loop through files and process them until we process all blocks needed
+    while (numOfBlocksToProcess !== 0 && fileIndex >= 0) {
+      const blockFileName = blockFiles[fileIndex];
+      fileIndex--;
+      const blockFile = fileReader.readBlockFile(blockFileName);
+      const blockData = BitcoinRawDataParser.parseRawDataFile(blockFile);
+      // for each block within the current file, add fee info and write to db
+      for (let key in blockData) {
```
mudiali, line 216: You know what I'm going to say :) Can we break this down into smaller functions? I think it would more mantainable and testable if the while loop(s) can be 'read in english' (as much as possible off course). Kinda like:

````
...
while (numOfBlocksToProcess ...) {
  ...
  // my function names are only examples and are not accurate.
  const feeInfoFromFile = this.saveSidetreeTransactionsAndGetFeeInfo(fileName);
  this.removeInvalidFeeAndTransactionDbData(...); 
  ...
}
````

File: lib/bitcoin/BitcoinProcessor.ts [hunk 2]
```diff
@@ -176,8 +188,130 @@ export default class BitcoinProcessor {
...
+      height: bestHeight,
+      hash: bestHash,
+      previousHash: previousHash
+    };
+    console.log('finished fast processing');
+    // TODO: Issue #783
+    // Need to use the fee infos and set fee after here.
+  }
+
+  private addBlockToFeeInfos (feeInfos: any[], indexInFeeInfo: number, block: BitcoinBlockModel) {
```
mudiali, line 258: Generally speaking, I'm uneasy about `any` types. We are using typescript so let's use the type-safety as much as we can IMO. Can we have an actual object type here (even an internal model to keep it within this class?)

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -176,8 +188,130 @@ export default class BitcoinProcessor {
...
+    if (feeInfos[indexInFeeInfo] === undefined) {
+      feeInfos[indexInFeeInfo] = {};
+    }
+    const blockReward = this.getBitcoinBlockReward(block.height);
```
mudiali, line 262: I love that this is in a separate function :) It is so much readable and testable IMO.

lib/common/MongoDbTransactionStore.ts
```diff
@@ -151,6 +151,14 @@ export default class MongoDbTransactionStore implements ITransactionStore {
     await this.transactionCollection!.deleteMany({ transactionNumber: { $gt: Long.fromNumber(transactionNumber) } });
   }
 
+  /**
+   * Remove transactions by transaction time hash
+   * @param transactionTimeHash the transaction time hash which the transactions should be removed for
+   */
+  public async removeTransactionByTransactionTimeHash (transactionTimeHash: string) {
+    await this.transactionCollection!.deleteMany({ transactionTimeHash: { $eq:  transactionTimeHash } });
```
thehenrytsai, line 159: I wonder if it is worthwhile/overkill to index by `transactionTimeHash` now? It is obviously not a problem now, but I would imagine multiple passes against millions of transactions is going to take a while.

lib/bitcoin/BitcoinRawDataParser.ts
```diff
@@ -1,6 +1,8 @@
+import BitcoinClient from './BitcoinClient';
 import ErrorCode from './ErrorCode';
 import SidetreeError from '../common/SidetreeError';
 import { Block } from 'bitcore-lib';
+import BitcoinBlockModel from './models/BitcoinBlockModel';
```
thehenrytsai, line 5: Nit: sort.

lib/bitcoin/BitcoinRawDataParser.ts
```diff
@@ -23,17 +26,20 @@ export default class BitcoinRawDataParser {
    * creating new Block, or validating size
    * @param rawBlockDataFileBuffer The file, in buffer form, to be parsed as blocks
    */
-  public static parseRawDataFile (rawBlockDataFileBuffer: Buffer): any {
+  public static parseRawDataFile (rawBlockDataFileBuffer: Buffer): {[name: string]: BitcoinBlockModel} {
```
thehenrytsai, line 29:
```suggestion
  public static parseRawDataFile (rawBlockDataFileBuffer: Buffer): {[blockHash: string]: BitcoinBlockModel} {
```

lib/bitcoin/BitcoinRawDataParser.ts
```diff
@@ -54,7 +60,30 @@ export default class BitcoinRawDataParser {
         throw new SidetreeError(ErrorCode.BitcoinRawDataParserInvalidBlockData);
       }
 
-      blockMapper[block.hash] = block;
+      // the first transaction, the coinbase, contains the block height in its input
+      const coinbaseInputScript = (block.transactions[0].inputs[0] as any)._scriptBuffer;
```
thehenrytsai, line 64:
```suggestion
      const coinbaseInputScript = (block.transactions[0].inputs[0] as any)._scriptBuffer as Buffer;
```

lib/bitcoin/BitcoinRawDataParser.ts
```diff
@@ -54,7 +60,30 @@ export default class BitcoinRawDataParser {
...
+      const transactionModels = block.transactions.map((transaction: any) => {
+        const bitcoreTransaction = {
+          id: transaction.id,
+          blockHash: block.hash,
+          confirmations: 1, // set to 1 because it has to be confirmed to be on the current longest chain
+          inputs: transaction.inputs,
+          outputs: transaction.outputs
+        };
+        return BitcoinClient.createBitcoinTransactionModel(bitcoreTransaction);
+      });
```
thehenrytsai, line 79: Consider pulling this into a reusable method that can go from bitcore transaction to our transaction. Perhaps: BitcoinClient.convertToBitcoinTransactionModel(bitcoreTransaction).

lib/bitcoin/BitcoinRawDataParser.ts
```diff
@@ -54,7 +60,30 @@ export default class BitcoinRawDataParser {
...
+      // this denotes how many bytes following represent the block height
+      const heightBytes = coinbaseInputScript.readUInt8();
+      // the next n bytes are the block height in little endian
+      const blockHeight = coinbaseInputScript.readUIntLE(1, heightBytes);
+
+      const transactionModels = block.transactions.map((transaction: any) => {
+        const bitcoreTransaction = {
+          id: transaction.id,
+          blockHash: block.hash,
+          confirmations: 1, // set to 1 because it has to be confirmed to be on the current longest chain
```
thehenrytsai, line 74: Looks like you are just assigning a value here because you have to, if so consider explicitly say so in the comments.
```suggestion
          confirmations: 1, // Unused, but need to set to a value to be able reuse `BitcoinTransactionModel`.
```
Alternatively, you could have `BitcoinTransactionModel` and `ConfirmedBitcoinTransactionModel`, but I would agree that's probably for annoying than helpful.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -166,7 +172,13 @@ export default class BitcoinProcessor {
     }
 
     console.info(`Starting block: ${startingBlock.height} (${startingBlock.hash})`);
-    await this.processTransactions(startingBlock);
+    // this reads into the raw block files and parse to speed to the initial startup
+    // will use this instead of the rpc processing when ready
```
thehenrytsai, line 176: Nit: move into between line 177 & 178.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -176,8 +188,130 @@ export default class BitcoinProcessor {
...
+        for (let hash of hashes) {
+          if (hash !== currentHash) {
+            delete currentFeeData[hash];
+            await this.transactionStore.removeTransactionByTransactionTimeHash(hash);
+          }
+        }
+        const confirmedFeeData = currentFeeData[currentHash];
+        if (currentHash === bestHash) {
+          previousHash = confirmedFeeData.prevHash;
+        }
```
thehenrytsai, line 241: I think the `previousHash` here is very misleading, I thought it is related to `currentHash`, but looks like it really is just related to `bestHash`. Suggest to decouple (remove) this block and get this data through other means for line 251. e.g. There is the `feeInfos` for instance.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -166,7 +172,13 @@ export default class BitcoinProcessor {
     }
 
     console.info(`Starting block: ${startingBlock.height} (${startingBlock.hash})`);
-    await this.processTransactions(startingBlock);
+    // this reads into the raw block files and parse to speed to the initial startup
+    // will use this instead of the rpc processing when ready
+    if (this.bitcoinDataDirectory) {
+      await this.fastProcessTransactions(startingBlock);
+    } else {
+      await this.processTransactions(startingBlock);
```
thehenrytsai, line 180: Is it possible to add the equivalent code for fetching the coinbase transaction data in `processTransaction` also?

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -176,8 +188,130 @@ export default class BitcoinProcessor {
...
+      let currentFeeData = feeInfos[numOfBlocksToProcess - 1];
+      while (currentFeeData !== undefined && currentFeeData[currentHash] !== undefined) {
+        // delete all unneeded key value pairs and db records
+        const hashes = Object.keys(currentFeeData);
+        for (let hash of hashes) {
+          if (hash !== currentHash) {
+            delete currentFeeData[hash];
+            await this.transactionStore.removeTransactionByTransactionTimeHash(hash);
+          }
+        }
```
thehenrytsai, line 237: Consider moving this into a method, perhaps `removeTransactionsWithBlockchainTimeHashes(hashesToRemove)`. (there is a question about what to do with `delete currentFeeData[hash];`

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -176,8 +188,130 @@ export default class BitcoinProcessor {
...
+      const blockFileName = blockFiles[fileIndex];
+      fileIndex--;
+      const blockFile = fileReader.readBlockFile(blockFileName);
+      const blockData = BitcoinRawDataParser.parseRawDataFile(blockFile);
+      // for each block within the current file, add fee info and write to db
+      for (let key in blockData) {
+        const block = blockData[key];
+
+        const indexInFeeInfo = block.height - startingBlock.height;
+        if (indexInFeeInfo >= 0 && indexInFeeInfo < numOfBlocksToProcess) {
```
thehenrytsai, line 220: In general I find the conditional checks (including ifs, whiles) require unpacking a level of indirection to understand.

For instance, `indexInFeeInfo < numOfBlocksToProcess`: would be nice if the condition was "more readable" like: `block.height < earliestKnownBlock.height`

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -176,8 +188,130 @@ export default class BitcoinProcessor {
   }
 
   /**
-   * Trims the databases to the block prior to the block the given transaction is in.
-   * @returns The last fully processed block after trimming. `undefined` if all data are deleted after trimming.
+   * A faster version of process transactions that requires access to bitcoin data directory
```
thehenrytsai, line 191: Sanity: I am a bit concerned that in the event that the node is terminated unexpectedly halfway during this bootstrapping method, when it comes back up it will incorrectly think that the node is sync to a certain transaction (based on the last transaction in DB), and incorrect pick up from there resulting in bad state.

lib/bitcoin/BitcoinProcessor.ts
isaacJChen: Yes. We currently throw the the coinbase script away because we don't need it

lib/bitcoin/BitcoinProcessor.ts
isaacJChen: Created an issue for it https://github.com/decentralized-identity/sidetree/issues/794

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -176,8 +193,144 @@ export default class BitcoinProcessor {
...
+  private async fastProcessTransactions (startingBlock: IBlockInfo) {
+    const bitcoinBlockDataIterator = new BitcoinBlockDataIterator(this.bitcoinDataDirectory!);
+    const bestHeight = await this.bitcoinClient.getCurrentBlockHeight();
+    const bestHash = await this.bitcoinClient.getBlockHash(bestHeight);
+
+    let currentHash = bestHash;
+    let currentHeight = bestHeight;
+    const numOfBlocksToProcess = bestHeight - startingBlock.height + 1;
+
+    const chainInfos: (Map<string, IChainInfo> | undefined)[] = new Array(numOfBlocksToProcess).fill(undefined);
```
thehenrytsai, line 208: Might need to change condition in line 256, but is it possible to fill the array with empty maps so:
1. No need to have `undefined` declaration in multiple places.
1. No need for `if` block on line 277.

lib/bitcoin/BitcoinBlockDataIterator.ts
```diff
@@ -0,0 +1,38 @@
+import BitcoinFileReader from './BitcoinFileReader';
+import BitcoinBlockModel from './models/BitcoinBlockModel';
+import BitcoinRawDataParser from './BitcoinRawDataParser';
+
+/**
+ * Iterates through block data by parsing raw block data file from latest file to oldest
+ */
+export default class BitcoinBlockDataIterator {
```
thehenrytsai, line 8:
```suggestion
export default class BitcoinBlockIterator {
```

lib/bitcoin/BitcoinBlockDataIterator.ts
```diff
@@ -0,0 +1,38 @@
+import BitcoinFileReader from './BitcoinFileReader';
+import BitcoinBlockModel from './models/BitcoinBlockModel';
+import BitcoinRawDataParser from './BitcoinRawDataParser';
+
+/**
+ * Iterates through block data by parsing raw block data file from latest file to oldest
```
thehenrytsai, line 6:
```suggestion
 * Iterates through blocks backwards from latest block to oldest block by parsing raw block data files.
```

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -45,6 +47,13 @@ export interface IBlockInfo {
   previousHash: string;
 }
 
+interface IChainInfo {
```
thehenrytsai, line 50: I see this interface being more `IBlockInfo` or `IBlockMetadata`

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -176,8 +193,144 @@ export default class BitcoinProcessor {
...
+  private async processBlockData (
+    blockData: {[blockHash: string]: BitcoinBlockModel},
+    chainInfos: (Map<string, IChainInfo> | undefined)[],
+    startingBlockHeight: number,
+    currentHeight: number) {
+
+    for (let key in blockData) {
+      const block = blockData[key];
+
+      const indexInChainInfos = block.height - startingBlockHeight;
```
thehenrytsai, line 237: Nit: move line 237 into if statement.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -176,8 +191,144 @@ export default class BitcoinProcessor {
...
+  }
+
+  private async processBlockData (
+    blockData: {[blockHash: string]: BitcoinBlockModel},
+    chainInfos: Map<string, IBlockInfoExtended>[],
+    startingBlockHeight: number,
+    currentHeight: number) {
+
+    for (let blockHash in blockData) {
+      const block = blockData[blockHash];
```
thehenrytsai, line 234: This feels like a round about way to iterate through block data, why not just return a map, or array directly from the parser for `blockData`?

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -176,8 +191,144 @@ export default class BitcoinProcessor {
...
+    const numOfBlocksToProcess = bestHeight - startingBlock.height + 1;
+
+    // An array of maps<hash, IBlockInfoExtended>.
+    const chainInfos = [...new Array(numOfBlocksToProcess)].map<Map<string, IBlockInfoExtended>>(() => new Map());
+
+    console.log(`Begin fast processing block ${startingBlock.height} to ${bestHeight}`);
+    // loop through files and process them until we process all blocks needed
+    while (bitcoinBlockDataIterator.hasNext() && currentHeight >= startingBlock.height) {
+      const blockData = bitcoinBlockDataIterator.next()!;
+      await this.processBlockData(blockData, chainInfos, startingBlock.height, currentHeight);
```
thehenrytsai, line 213: I start to have issue with the name `BlockData` throughout the code as I have on multiple occasions thinking it is data on a single block, but only to realize that it is multiple of them, would love to have a better name.

lib/bitcoin/BitcoinProcessor.ts
isaacJChen: Will change to Array after current review.

lib/bitcoin/BitcoinProcessor.ts
isaacJChen: just `blocks`?

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -224,65 +233,61 @@ export default class BitcoinProcessor {
...
-        const indexInChainInfos = block.height - startingBlockHeight;
-        BitcoinProcessor.addBlockToChainInfo(chainInfos, indexInChainInfos, block);
+        const blockReward = BitcoinProcessor.getBitcoinBlockReward(block.height);
+        notYetValidatedBlocks.set(
+          block.hash,
+          {
+            height: block.height,
+            hash: block.hash,
+            totalFee: block.transactions[0].outputs[0].satoshis - blockReward,
+            transactionCount: block.transactions.length - 1, // minus one because of coinbase
```
thehenrytsai, line 251: Discussion: I'd rather `transactionCount` to be the same as `transactions.length` so no context knowledge required, of course it would skew the fee by 1/3000th, but not big issue I think.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -224,65 +233,61 @@ export default class BitcoinProcessor {
...
+    notYetValidatedBlocks: Map<string, IBlockInfoExtended>,
     heightOfEarliestKnownValidBlock: number,
     hashOfEarliestKnownValidBlock: string,
     startingBlockHeight: number): Promise<[number, string]> {
 
     let index = heightOfEarliestKnownValidBlock - startingBlockHeight;
-    let currentFeeData = chainInfos[index];
-    while (currentFeeData !== undefined && currentFeeData.get(hashOfEarliestKnownValidBlock) !== undefined) {
+    let validBlock = notYetValidatedBlocks.get(hashOfEarliestKnownValidBlock);
+    while (validBlock !== undefined && index >= 0) {
```
thehenrytsai, line 272: Consider:
```suggestion
    while (validBlock !== undefined && validBlock.height >= startingBlockHeight) {
```

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -224,65 +233,61 @@ export default class BitcoinProcessor {
...
   /**
-   * Checks in the chain info to find the earliest valid block and its hash,
-   * and delete all transactions from invalid blocks in DB.
+   * Find all hashes in the notYetValidatedBlocks that are actually valid,
+   * add them to the validated list and delete them from the map.
    */
-  private async findEarliestValidBlock (
-    chainInfos: Map<string, IBlockInfoExtended>[],
+  private async findEarliestValidBlockAndAddToValidBlocks (
+    validatedBlocks: (IBlockInfoExtended | undefined)[],
```
thehenrytsai, line 264: Discussion: what about just `validatedBlocks: IBlockInfoExtended[]`?

I know it maybe probably be slightly less quick ('coz we are trusting the runtime to do memory management), but the gain is slightly less code.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -224,65 +233,61 @@ export default class BitcoinProcessor {
...
+      // delete because it is now validated
+      notYetValidatedBlocks.delete(hashOfEarliestKnownValidBlock);
+      // the previous block hash becomes valid
+      hashOfEarliestKnownValidBlock = validBlock.previousHash;
       heightOfEarliestKnownValidBlock--;
       index--;
-      currentFeeData = chainInfos[index];
+      validBlock = notYetValidatedBlocks.get(hashOfEarliestKnownValidBlock);
     }
     return [heightOfEarliestKnownValidBlock, hashOfEarliestKnownValidBlock];
```
thehenrytsai, line 283: Comment: it just pains me that we are forced to pass in AND return this tuple because of the initialization case... 'coz except for the first block, `validatedBlocks` should already give us info needed.... I have no good solution to this, but thought I should call out in case I missed something obvious.

lib/bitcoin/BitcoinProcessor.ts
isaacJChen: Fair

lib/bitcoin/BitcoinProcessor.ts
isaacJChen: So they are always defined and we do array.shift()?

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -224,65 +233,61 @@ export default class BitcoinProcessor {
...
-    chainInfos[indexInChainInto]!.set(
-      block.hash,
-      {
-        height: block.height,
-        hash: block.hash,
-        totalFee: block.transactions[0].outputs[0].satoshis - blockReward,
-        transactionCount: block.transactions.length - 1, // minus one because of coinbase
-        previousHash: block.previousHash }
-      );
+  private async processInvalidBlocks (invalidBlocks: Map<string, IBlockInfoExtended>) {
```
thehenrytsai, line 286: Nit.
```suggestion
  private async removeInvalidBlocks (invalidBlocks: Map<string, IBlockInfoExtended>) {
```

lib/bitcoin/BitcoinProcessor.ts
thehenrytsai: No no, just `validatedBlocks.push()`, I can reverse it if I need to.

lib/bitcoin/BitcoinProcessor.ts
thehenrytsai: Sounds good, but the monkey wrench is `BitcoinBlockDataIterator` would become `BitcoinBlocksIterator`.... Maybe for that class we call it.... `BitcoinBlockFileIterator`?

lib/bitcoin/BitcoinProcessor.ts
isaacJChen: Changed

lib/bitcoin/BitcoinProcessor.ts
isaacJChen: I think BlockDataIterator still makes sense because we are iterating through the raw block data, processing them and then return the processed result. Will change it if you really don't like it thought.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -199,30 +199,31 @@ export default class BitcoinProcessor {
     const lastBlockHeight = await this.bitcoinClient.getCurrentBlockHeight();
     const lastBlockInfo = await this.bitcoinClient.getBlockInfoFromHeight(lastBlockHeight);
 
-    const numOfBlocksToProcess = lastBlockHeight - startingBlock.height + 1;
-
     // a map of all blocks mapped with their hash being the key
     const notYetValidatedBlocks: Map<string, IBlockInfoExtended> = new Map();
     // An array of blocks representing the validated chain sorted by height
```
thehenrytsai, line 204:
```suggestion
    // An array of blocks representing the validated chain reverse sorted by height
```

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -199,30 +199,31 @@ export default class BitcoinProcessor {
...
-      [heightOfEarliestKnownValidBlock, hashOfEarliestKnownValidBlock] = await this.findEarliestValidBlockAndAddToValidBlocks(
+      const blocks = bitcoinBlockDataIterator.previous()!;
+      await this.processBlocks(blocks, notYetValidatedBlocks, startingBlock.height, heightOfEarliestKnownValidBlock);
+      this.findEarliestValidBlockAndAddToValidBlocks(
         validatedBlocks,
         notYetValidatedBlocks,
-        heightOfEarliestKnownValidBlock,
         hashOfEarliestKnownValidBlock,
         startingBlock.height);
+      if (validatedBlocks.length) {
```
thehenrytsai, line 219: Style nit: don't like javascript treating 0 as false.
```suggestion
      if (validatedBlocks.length > 0) {
```

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -166,7 +172,13 @@ export default class BitcoinProcessor {
     }
 
     console.info(`Starting block: ${startingBlock.height} (${startingBlock.hash})`);
-    await this.processTransactions(startingBlock);
+    // this reads into the raw block files and parse to speed to the initial startup
+    // will use this instead of the rpc processing when ready
+    if (this.bitcoinDataDirectory) {
+      await this.fastProcessTransactions(startingBlock);
```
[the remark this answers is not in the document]
isaacJChen: It should work, always, unless you don't have bitcoin data.

PR #793, opened 2020-07-09
build: add conventional changelog support

tplooker:
This PR adds support for automated changelog generation during the release process

**Note** - this will enforce a specific commit format from now on so the changelog can be automatically generated based on the changelog.

Looking for feedback on this approach and confirming it is not going to be a developer burden.

### DIFF
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -14,10 +14,13 @@ module.exports = {
 			['sentence-case', 'start-case', 'pascal-case', 'upper-case']
 		],
 		"scope-empty": [2, "never"],
-        "scope-enum": [2, "always", [
-            "spec",
-            "ref-impl"
-        ]],
+        "scope-enum": [
+			2, 
+			"always", [
+				"spec",
+				"ref-impl"
+			]
+		],
 		'type-case': [2, 'always', 'lower-case'],
 		'type-empty': [2, 'never'],
 		'type-enum': [
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -1,0 +1,41 @@
+// Configuration is extended from @commitlint/config-conventional
+// @see https://github.com/conventional-changelog/commitlint/tree/master/@commitlint/config-conventional
+module.exports = {
+	rules: {
+		'body-leading-blank': [1, 'always'],
+		'body-max-line-length': [2, 'always', 100],
+		'footer-leading-blank': [1, 'always'],
+		'footer-max-line-length': [2, 'always', 100],
+		'header-max-length': [2, 'always', 100],
+		'scope-case': [2, 'always', 'lower-case'],
+		'subject-case': [
+			2,
+			'never',
+			['sentence-case', 'start-case', 'pascal-case', 'upper-case']
+		],
+		"scope-empty": [2, "never"],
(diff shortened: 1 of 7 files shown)


tplooker:
We will need to manually generate the changelog up until this date so later version releases will generate the changelog automatically

OR13:
@tplooker what needs to be done to make this actionable?

tplooker:
@OR13 first need to get the CI user access token sorted for the build pipeline, documented in #795, which is currently blocked as I'm awaiting user being setup in the npm org correctly

OR13:
JFYI you can setup NPM Access tokens at an organization level in github...  https://github.blog/changelog/2020-05-14-organization-secrets/ @csuwildcat

commitlint.config.js
```diff
@@ -0,0 +1,41 @@
...
+		'subject-case': [
+			2,
+			'never',
+			['sentence-case', 'start-case', 'pascal-case', 'upper-case']
+		],
+		"scope-empty": [2, "never"],
+        "scope-enum": [2, "always", [
+            "spec",
+            "ref-impl"
+        ]],
```
thehenrytsai, line 20: Is this indented incorrectly?

commitlint.config.js
```diff
@@ -0,0 +1,41 @@
+// Configuration is extended from @commitlint/config-conventional
+// @see https://github.com/conventional-changelog/commitlint/tree/master/@commitlint/config-conventional
+module.exports = {
+	rules: {
```
thehenrytsai, line 4: This is probably a stretch ask. Is there a way to enforce an issue number to be associated with the commit message? like

#123 - fix(spec): ambiguity around update operation terminology

tplooker: There are integrations like this but I would have to investigate a little more

commitlint.config.js
tplooker: Fixed now

commitlint.config.js
OR13: code reviews can address this. just reject the PR if it doesn't reference issues.

package.json
```diff
@@ -33,29 +34,41 @@
...
     "lint": "tslint --fix --project .",
     "spec": "node ./scripts/spec-render.js",
-    "spec:edit": "node ./scripts/spec-edit.js"
+    "spec:edit": "node ./scripts/spec-edit.js",
+    "version": "conventional-changelog -p angular -i CHANGELOG.md -s && git add CHANGELOG.md"
   },
   "husky": {
     "hooks": {
-      "pre-commit": "npm run lint"
+      "pre-commit": "[[ -n $HUSKY_BYPASS ]] || npm run lint || pretty-quick --staged",
```
OR13, line 65: what is pretty-quick --staged doing?

thehenrytsai: I think automats the code in staged files.

tplooker: Good catch, nothing, will remove

thehenrytsai: Oh? You didn't intend to use `pretty-quick` for auto formatting?

tplooker: We can, but have to install and configure prettier too e.g see here https://github.com/mattrglobal/bls12381-key-pair/blob/master/package.json#L35

PR #801, opened 2020-07-14
Allow JWK fingerprint to be used in fragments

OR13:
> The value of the id property MAY be structured as a compound key. This is especially useful for integrating with existing key management systems and key formats such as JWK. It is RECOMMENDED that JWK kid values are set to the public key fingerprint. It is RECOMMENDED that verification methods that use JWKs to represent their public keys utilize the value of kid as their fragment identifier. See the first key in EXAMPLE 15 for an example of a public key with a compound key identifier.

https://w3c.github.io/did-core/


docs/spec/patches.md
```diff
@@ -51,7 +51,7 @@ The `add-public-keys` _Patch Action_ describes the addition of cryptographic key
 1. The object ****MUST**** include an `action` property, and its value ****MUST**** be `add-public-keys`.
 2. The object ****MUST**** include a `public_keys` property, and its value ****MUST**** be an array.
 3. Each key being added ****MUST**** be represented by an entry in the `public_keys` array, and each entry must be an object composed as follows:
-    1. The object ****MUST**** include an `id` property, and its value ****MUST**** be a string with no more than twenty (20) ASCII encoded characters. If the value is not of the correct type or exceeds the specified length, the entire _Patch Action_ ****MUST**** be discarded, without any of it being used to modify the DID's state.
+    1. The object ****MUST**** include an `id` property, and its value ****MUST**** be a string with no more than twenty (50) ASCII encoded characters. If the value is not of the correct type or exceeds the specified length, the entire _Patch Action_ ****MUST**** be discarded, without any of it being used to modify the DID's state.
```
gjgd, line 54:
```suggestion
    1. The object ****MUST**** include an `id` property, and its value ****MUST**** be a string with no more than fifty (50) ASCII encoded characters. If the value is not of the correct type or exceeds the specified length, the entire _Patch Action_ ****MUST**** be discarded, without any of it being used to modify the DID's state.
```

docs/spec/patches.md
```diff
@@ -118,7 +118,7 @@ The `add-service-endpoints` _Patch Action_ describes the addition of [Service En
...
 2. The object ****MUST**** include a `service_endpoints` property, and its value ****MUST**** be an array. If the value is not of the correct type, the entire _Patch Action_ ****MUST**** be discarded, without any of it being used to modify the DID's state.
 3. Each service being added ****MUST**** be represented by an entry in the `service_endpoints` array, and each entry must be an object composed as follows:
-    1. The object ****MUST**** include an `id` property, and its value ****MUST**** be a string with a length of no more than twenty (20) ASCII encoded characters. If the value is not of the correct type or exceeds the specified length, the entire _Patch Action_ ****MUST**** be discarded, without any of it being used to modify the DID's state.
+    1. The object ****MUST**** include an `id` property, and its value ****MUST**** be a string with a length of no more than twenty (50) ASCII encoded characters. If the value is not of the correct type or exceeds the specified length, the entire _Patch Action_ ****MUST**** be discarded, without any of it being used to modify the DID's state.
```
gjgd, line 121:
```suggestion
    1. The object ****MUST**** include an `id` property, and its value ****MUST**** be a string with a length of no more than fifty (50) ASCII encoded characters. If the value is not of the correct type or exceeds the specified length, the entire _Patch Action_ ****MUST**** be discarded, without any of it being used to modify the DID's state.
```

OR13:
@gjgd thanks for the review!

troyronda:
Example of a thumbprint can also be found here (which would be the fragment portion shown in example 15): https://tools.ietf.org/html/rfc7638#section-3.1

> NzbLsXh8uDCcd-6MNwXF4W_7noWXFZAfHkxZsRGC9Xs

Which is 43 characters for _just_ the fragment portion.

OR13:
I chose 50 because multicodec identifiers are 49.... the identifiers are mean to be just the fragment, this preserves the implied type system, it just ensures it support popular identifiers we are seeing / using today with our customers.

csuwildcat:
Can someone help me understand why this change is being done now, after the discussion, spec definition, and PRs merged with these strictures was completed long ago? Unlike 2 years ago, organizations now have code in near-production flight that these changes will cause issues for. I am not confident, given the seeming inability of some to definitively discuss, review, and make lasting decisions, that we will not find ourselves endlessly readjudicating string changes that take 10 minutes to PR for spec language modifications, but cause a cascade of dev hours to deal with.

OR13:
@csuwildcat this change is being proposed to allow for https://tools.ietf.org/html/rfc7638. and multicodec fingerprints... without support for these, sidetree strict update operations are incapable of expressing keys in ways that are common / needed for interoperability with JOSE.

csuwildcat:
I think we talked about this as only being a change to the DID Document object IDs, not something that would require duplicate long strings of the same value in both those IDs and the kids of the JWKs, correct? If so, I do not have much issue with it, and can accept the will of the group.

OR13:
@csuwildcat thats correct, I'm going to count that as a Code Review with "Approve Changes".

PR #804, opened 2020-07-15
add purpose values from spec + make property names camelCase + typos

julio-cabdu:
In many files, property names are camelCase, can we make it cohesive everywhere?
Also, fix some typos, and type names inconsistencies, e.g. call the property encodedDelta instead of delta, to leave delta for the Operation Delta Object of type DeltaModel.


csuwildcat:
These properties come from the specification (https://identity.foundation/sidetree/spec/), which we decided would use snake case, as it is far more readable for data formats and other static documents. I can't tell if these changes would break the code, but if they are coming from the places where the code is touching the JSON documents/objects the spec defines, it would. At this point, I do not believe there will be support for changing either all of the reference code to be snake, all of the spec to be camel, and v.v.

julio-cabdu:
I take your point that is more readable for data formats and other static documents, but is it camelCase such a problem in those cases? I can find some references that prefer camel for JSON, e.g. [this](https://google.github.io/styleguide/jsoncstyleguide.xml?showone=Property_Name_Format#Property_Name_Format). In the code, many property names are in camel already so I hope it wouldn't be a big issue in that regard. I'll make the missing edits to fix the npm run errors.

OR13:
Go here to discuss casing: https://github.com/decentralized-identity/sidetree/issues/805

I suggest we close this PR until we reach consensus on the issue.

PR #806, opened 2020-07-16
fix total fee calculation

isaacJChen:


lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -284,6 +283,24 @@ export default class BitcoinProcessor {
...
 
+  /**
+   * Iterate through all the outputs in the first transaction (coinbase) and add up all the satoshis
+   * then minus the block reward to get the total transaction fee.
+   * @param block The block to get the fee for
+   */
+  private static getBitcoinBlockTotalFee (block: BitcoinBlockModel) {
+    // get the total fee including block reward
+    const coinbaseTransaction = block.transactions[0];
+    let totalFee = 0;
```
thehenrytsai, line 294: Nit on style: Avoid mixing uses of a variable, consider creating two variables `totalOutput` and `totalFee`, or just return without assigning is fine.
```suggestion
    let totalOutput= 0;
```

Issue #803, opened 2020-07-15
Handle database upgrade automatically with each new release


thehenrytsai:
Currently if we move genesis block when a new build is released, we require manual deletion of database. This is error prone and not sustainable going forward especially if when we have real schema changes. We should have a feature that handles DB upgrade automatically.

thehenrytsai:
Implemented and checked in.

Issue #783, opened 2020-07-01
Simplify normalized fee calculation algorithm


thehenrytsai:
Consider simplifying existing  normalized fee calculation algorithm with a basic moving average + growth cap.

thehenrytsai:
Implemented.

Issue #667, opened 2020-05-14
Improve bitcoin startup and avoid DB deletion.


mudiali:
https://github.com/decentralized-identity/sidetree/blob/731dd56f7e2581d82ed730f2aa4839b802d67f3a/lib/bitcoin/BitcoinProcessor.ts#L464

Currently, `getStartingBlockForInitialization` functions checks whether the last transaction-time stored in DB hasn't forked. If not forked, it assumes that block as the starting point (the above if statement). We can improve the above if statement to read from the quantile DB to figure out what was the last group that we saved correctly and start from that point onwards. 

**Need to think through to make sure that if there was a fork after the last transaction-time stored in DB then is doing this ok?**

Pseudo-code to refactor the above if statement:

````javascript
if (lastSavedBlockIsValid) {
  const lastSavedGroup = quantileCalculator.getLastGroupId();
  const newGroupToStart = lastSavedGroup++;
  const firstBlockInNewGroupToStart = this.getFirstBlockInGroup(newGroupToStart);
  lastValidBlock = await this.trimDatabasesToFeeSamplingGroupBoundary(firstBlockInNewGroupToStart - 1); 
}
````
-------------------------------

**Possible Solution 2**
Another solution that came up with offline brainstorming, and is probably better due to catching the forking issue, is that we save either the lastProcessedBlock + hash or lastGroup.lastBlock + hash in the DB. Then at startup check the block + hash to ensure that there's no fork.

thehenrytsai:
Implemented and checked in.

PR #807, opened 2020-07-17
#667, #783, #803 - Multiple improvements on bitcoin initialization + storage of block metadata

thehenrytsai:
1. Implemented storage of metadata to allow implementation of normalized fee calculator in the future.
1. Avoided unnecessary DB deletion on bitcoin service initialization.
1. Added code to handle database upgrade automatically with each new release.
1. Simplified revert/trim logic to be based on 'true' last processed block, as opposed to last processed transaction.
1. Bumped TypeScript and MongoDB types version to take advantage of new features.


lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -190,6 +199,26 @@ export default class BitcoinProcessor {
     void this.periodicPoll();
   }
 
+  private async upgradeDatabaseIfNeeded () {
+    const currentServiceVersion = await this.getServiceVersion();
+    const savedServiceState = await this.serviceStateStore.get();
+    const savedServiceVersion = savedServiceState ? savedServiceState.serviceVersion : 'unknown';
+
+    // Upgrade the DB if the saved service version is different to current running service version.
+    if (savedServiceVersion !== currentServiceVersion.version) {
```
mudiali, line 208: So every time we release a new npm package it will trigger this upgrade right? If yes, then I'm not sure if that is what we want.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -601,28 +611,26 @@ export default class BitcoinProcessor {
...
-    for (let blockHeight = startBlockHeight; blockHeight <= endBlockHeight; blockHeight++) {
-      const processedBlockHash = await this.processBlock(blockHeight, previousBlockHash);
+      await this.blockMetadataStore.add([processedBlockMetadata]);
 
-      this.lastProcessedBlock = {
-        height: blockHeight,
-        hash: processedBlockHash,
-        previousHash: previousBlockHash
-      };
+      this.lastProcessedBlock = processedBlockMetadata;
```
mudiali, line 621: No explicit cast needed? This probably works because the property names are same? If that's the reason then maybe we can initialize the `this.lastProcessedBlock` explicitly in case the property names change in the future?

Another option might be to change the type of `this.lastProcessedBlock` to `BlockMetadata` but that might trigger more cascading changes (especially in the test) so ok with not doing that.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -601,28 +611,26 @@ export default class BitcoinProcessor {
     const endBlockHeight = await this.bitcoinClient.getCurrentBlockHeight();
     console.info(`Processing transactions from ${startBlockHeight} to ${endBlockHeight}`);
 
+    let blockHeight = startBlockHeight;
     let previousBlockHash = startBlock.previousHash;
+    while (blockHeight <= endBlockHeight) {
```
mudiali, line 616: Curious: any reason to change from for to while loop? Readability?

lib/bitcoin/MongoDbBlockMetadataStore.ts
```diff
@@ -0,0 +1,116 @@
...
+    let lookBackDistance = 1;
+    let currentHeight = lastBlock!.height;
+    while (currentHeight >= firstBlock.height) {
+      heightOfBlocksToReturn.push(currentHeight);
+
+      currentHeight = lastBlock!.height - lookBackDistance;
+      lookBackDistance *= 2;
+    }
+
+    const exponentiallySpacedBlocks = await this.collection!.find<BlockMetadata>(
```
mudiali, line 108: Lint warning that this `find` function w/ the 2nd option is deprecated?

lib/bitcoin/MongoDbServiceStateStore.ts
```diff
@@ -0,0 +1,32 @@
+import IServiceStateStore from './interfaces/IServiceStateStore';
+import MongoDbStore from './MongoDbStore';
+
+/**
+ * Implementation of IServiceStateStore using MongoDB database.
+ */
+export default class MongoDbServiceStateStore<T> extends MongoDbStore implements IServiceStateStore<T> {
```
mudiali, line 7: Is this templatized for future use? As in other services may also use this store for their own service version? If yes, then does this belong into the `common` folder?

lib/bitcoin/MongoDbStore.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  /** MondoDB instance. */
+  protected db: Db | undefined;
+  /** MongoDB collection */
+  protected collection: Collection<any> | undefined;
+
+  /**
+   * Constructs a `MongoDbStore`;
+   */
+  constructor (private serverUrl: string, collectionName: string, databaseName?: string) {
```
mudiali, line 22: Not sure if the `databaseName?` should be optional for this class?

lib/bitcoin/MongoDbBlockMetadataStore.ts
mudiali: Had to update the mongodb types version :)

lib/bitcoin/BitcoinProcessor.ts
thehenrytsai: Yes, just readability. Because there are actually two params being incremented/change every loop, one part of the `for` line and the other at the end of the loop. Changing it make it consistent.

lib/bitcoin/MongoDbBlockMetadataStore.ts
thehenrytsai: You need to run `npm i`. This is the reason why MongoDB type version was updated.

lib/bitcoin/MongoDbServiceStateStore.ts
thehenrytsai: This is definitely written in preparation for future service usage. But I left it in bitcoin because "right now" there isn't other usage. But no push back to move it right now.

lib/bitcoin/MongoDbStore.ts
thehenrytsai: This follows the same pattern of all other `Store` classes, are you suggesting to change the pattern just for this class or for all?

lib/bitcoin/BitcoinProcessor.ts
thehenrytsai: It is also my hope that long term we eliminate `IBlockInfo` and use just `BlockMetadata`, but as you said it has more cascading effect and bloats the PR even further.

In the meantime we could make `BlockMetadata` extend `IBlockInfo` (which is what I did prior to this PR), but I didn't want to take `IBlockInfo` out of this file.

All is to say, I will explicitly set them as you've suggested for now.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -633,6 +641,10 @@ export default class BitcoinProcessor {
     let lastValidBlock: IBlockInfo | undefined;
     if (lastProcessedBlockIsValid) {
       lastValidBlock = this.lastProcessedBlock;
+
+      // We need trim the DB data to the last processed block,
```
isaacJChen, line 645: nit: 
```suggestion
      // We need to trim the DB data to the last processed block,
```

lib/bitcoin/BitcoinProcessor.ts
thehenrytsai: Happy to discuss further. I am not suggesting that this is how we DB upgrade in the long run, but my stance is that this is fine for now given how little data we have right now. While wiping away DB is the dumbest way to DB upgrade, it is "a way that works". Keep in mind that we have the opportunity to, and we should revisit this DB upgrade code every time we release a new version.

lib/bitcoin/MongoDbStore.ts
thehenrytsai: That's +2, so I will just change this class and leave other `Store` class alone. I see that those classes will inherit from this class in the future.

lib/bitcoin/BitcoinProcessor.ts
thehenrytsai: Discussed offline, will target upgrade only on DB with unknown service version.

Issue #753, opened 2020-06-16
If LockMonitor unable to broadcast create lock txn, then the lock can become invalid


mudiali:
1. Assume the valid duration for a lock is between 90 to 100 blocks.
1. Let's say that the current block height is 200 and the above function created a lock transaction which unlocks at block 295. This means that the lock duration is`295 - 200 = 95` blocks which is valid. Once the transaction is created & signed, it is saved in the lock DB and broadcasted.
1. Let's say that the broadcast fails (network congested or satoshi client is down or the node crashed) until the current block is 211. The `LockMonitor` just tries to rebroadcast until it is successful.
1. The node is finally able to broadcast and the transaction actually gets written in block 211.
1. With the unlock block being 295, the lock duration is `295 - 211 = 84` which is considered invalid.

In the above case, the following line will throw a sidetree error with code `lock_resolver_duration_is_invalid` and the lock monitor will never be able to resolve resulting in a hung periodic poll loop. The currently coded `LockMonitor` will not release the lock either.

Line that will throw: https://github.com/decentralized-identity/sidetree/blob/8fa9f21e649a663aa771a23e81329f3dce56ef25/lib/bitcoin/lock/LockMonitor.ts#L214

**Suggested solution**
1. Maybe catch the above error, and just call `this.releaseLock(...)` to return the back to the wallet and let the next iteration to retry creating the lock?
1. If a transaction was never broadcasted (or broadcasted and is in the mempool) then maybe remove it from the mempool?
1. ??

csuwildcat:
Resolution: try using relative lock

PR #808, opened 2020-07-17
#753 Implement relative locking to avoid hard dependency on a specific unlock block number

mudiali:
See the related issue #753 for details.


lib/bitcoin/lock/LockResolver.ts
```diff
@@ -85,9 +85,11 @@ export default class LockResolver {
     const lockStartBlock = await this.calculateLockStartingBlock(lockTransaction);
 
     // (C). verify that the lock duration is valid
-    if (!this.isLockDurationValid(lockStartBlock, scriptVerifyResult.unlockAtBlock!)) {
+    const unlockAtBlock = lockStartBlock + scriptVerifyResult.lockDurationInBlocks!;
+
+    if (!this.isLockDurationValid(lockStartBlock, unlockAtBlock)) {
```
isaacJChen, line 90: Since we operate on relative lock now, would i make sense to make this function just take in the `scriptVerifyResult.lockDurationInBlocks` ?

lib/bitcoin/lock/LockResolver.ts
```diff
@@ -85,9 +85,11 @@ export default class LockResolver {
     const lockStartBlock = await this.calculateLockStartingBlock(lockTransaction);
 
     // (C). verify that the lock duration is valid
-    if (!this.isLockDurationValid(lockStartBlock, scriptVerifyResult.unlockAtBlock!)) {
+    const unlockAtBlock = lockStartBlock + scriptVerifyResult.lockDurationInBlocks!;
+
+    if (!this.isLockDurationValid(lockStartBlock, unlockAtBlock)) {
       throw new SidetreeError(ErrorCode.LockResolverDurationIsInvalid,
-                              `Lock start block: ${lockStartBlock}. Unlock block: ${scriptVerifyResult.unlockAtBlock}`);
+                              `Lock start block: ${lockStartBlock}. Unlock block: ${unlockAtBlock}`);
```
isaacJChen, line 92: And then this could be `valid duration is x, but got y duration`

lib/bitcoin/BitcoinClient.ts
```diff
@@ -150,12 +150,12 @@ export default class BitcoinClient {
    * this transaction before spending from the wallet.
    *
    * @param lockAmountInSatoshis The amount to lock.
-   * @param lockUntilBlock  The block until the amount to lock to; the amount becomes spendable AT this block.
+   * @param lockDurationInBlocks  The number of blocks to lock the amount for; the amount becomes spendable AFTER this many blocks.
```
isaacJChen, line 153: If the lock is written to the chain at block 100, and we set the lock duration to 5, it will unlock at 105? (100, 101, 102, 103, 104) Just for my own knowledge more than anything :p

mudiali: That is correct @isaacJChen

lib/bitcoin/lock/LockResolver.ts
mudiali: Good idea .. let me see how can I improve this log without having to recalculate what's inside the function.

lib/bitcoin/BitcoinClient.ts
```diff
@@ -570,36 +570,36 @@ export default class BitcoinClient {
...
     const reFreezeTransaction = await this.createSpendTransactionFromFrozenTransaction(
       previousFreezeTransaction,
-      previousFreezeUntilBlock,
+      previousFreezeDurationInBlocks,
       payToScriptAddress);
 
     return [reFreezeTransaction, freezeScript.toHex()];
   }
 
   private async createSpendToWalletTransaction (
```
thehenrytsai, line 593: Nit: I probably mentioned this earlier and forgot the reason you gave already (was it for testing?). Looks like it just a method wrapping another method.

lib/bitcoin/BitcoinClient.ts
thehenrytsai: Yes is my understanding.

lib/bitcoin/BitcoinClient.ts
mudiali: `createSpendTransactionFromFrozenTransaction` is a utility function and gets called by 2 functions: `createSpendToFreezeTransaction` & `createSpendToWalletTransaction`. The 2nd one can be removed as it is literally one line calling the utility function but want to leave this for another opportunity to get this change out (there're test changes as well that I want to avoid).

PR #822, opened 2020-07-28
chore(ref-imp): publish release v0.10.0

thehenrytsai:


CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+<a name="0.10.0"></a>
+# [0.10.0](https://github.com/decentralized-identity/sidetree/compare/v0.9.1...v0.10.0) (2020-07-28)
+
+
+### Features
```
tplooker, line 5: I wonder if we should just amend and remove this from the changelog and just add a note saying "First release using changelog"

tplooker: As most of the commit history prior to this does not apply the required conventions

thehenrytsai: I am okay with leaving the automated message as is. Since you've approved it, I am not going to wait and merge the change.

PR #828, opened 2020-08-06
feat(ref-imp): changed IPFS CAS adaptor to use IPFS HTTP API directly

thehenrytsai:


lib/ipfs/Ipfs.ts
```diff
@@ -0,0 +1,163 @@
...
+
+const multihashes = require('multihashes');
+
+/**
+ * Class that implements the `ICas` interface by communicating with IPFS.
+ */
+export default class Ipfs implements ICas {
+  private fetch = nodeFetch;
+
+  public constructor (public uri: string, private fetchTimeoutInSeconds: number) { }
```
mudiali, line 23: Any reason why this has to be `public`. Also, should we do any validation on it? I think we are letting the methods throw if the uri is invalid.

lib/ipfs/Ipfs.ts
```diff
@@ -0,0 +1,163 @@
...
+
+  public async read (base64urlEncodedMultihash: string, maxSizeInBytes: number): Promise<FetchResult> {
+    // Convert base64url to base58 multihash.
+    let base58EncodedMultihashString;
+    try {
+      const multihashBuffer = base64url.toBuffer(base64urlEncodedMultihash);
+      multihashes.validate(multihashBuffer);
+      base58EncodedMultihashString = multihashes.toB58String(multihashBuffer);
+    } catch {
+      return { code: FetchResultCode.InvalidHash };
```
mudiali, line 83: Should we log the `error` thrown w/ the `base64urlEncodedMultihash` to help with debugging?

lib/ipfs/Ipfs.ts
```diff
@@ -0,0 +1,163 @@
...
+
+    // Mark content as `not found` if any error is thrown while fetching.
+    if (fetchResult instanceof Error) {
+      // Log appropriately based on error.
+      if (fetchResult instanceof SidetreeError &&
+          fetchResult.code === IpfsErrorCode.TimeoutPromiseTimedOut) {
+        console.log(`'${base64urlEncodedMultihash}' not found on IPFS.`);
+      } else {
+        // Log any unexpected error for investigation.
+        console.error(`Unexpected error fetching '${base64urlEncodedMultihash}' not found on IPFS.`);
```
mudiali, line 97: Suggest logging `SidetreeError.stringify(fetchResult as Error)` as well.

mudiali: Wait ... or are logging it in the this.fetch?

lib/ipfs/Ipfs.ts
```diff
@@ -0,0 +1,163 @@
...
+ * Class that implements the `ICas` interface by communicating with IPFS.
+ */
+export default class Ipfs implements ICas {
+  private fetch = nodeFetch;
+
+  public constructor (public uri: string, private fetchTimeoutInSeconds: number) { }
+
+  public async write (content: Buffer): Promise<string> {
+    // A string that is cryptographically impossible to repeat as the boundary string.
+    const multipartBoundaryString = crypto.randomBytes(32).toString('hex');
```
mudiali, line 27: I don't understand the boundary string concept. Any pointers?

lib/core/Core.ts
```diff
@@ -112,8 +110,7 @@ export default class Core {
   public async handleGetVersionRequest (): Promise<ResponseModel> {
     const responses = [
       this.serviceInfo.getServiceVersion(),
-      await this.blockchain.getServiceVersion(),
-      await this.cas.getServiceVersion()
+      await this.blockchain.getServiceVersion()
```
isaacJChen, line 113: Dicussion: Do we care about cas version at all? For example, ICas can have a getVersion() function and our implementation will just return the go ipfs version. Just a thought. Don't have to do it in the pr even if we want it.

tests/common/ReadableStream.spec.ts
```diff
@@ -20,4 +22,17 @@ describe('ReadableStream', () => {
 
     expect(content).toEqual(expectedContent);
   });
+
+  it('should throw error if stream exceeds the max allowed size.', async (done) => {
+    const inputFilePath = './tests/bitcoin/testData/bitcoinTwoBlocksRawDataHex.txt';
```
isaacJChen, line 27: lol glad this file is useful for something else :p

tests/core/TransactionProcessor.spec.ts
```diff
@@ -19,9 +19,9 @@ import TransactionProcessor from '../../lib/core/versions/latest/TransactionProc
 import ValueTimeLockModel from '../../lib/common/models/ValueTimeLockModel';
 import ValueTimeLockVerifier from '../../lib/core/versions/latest/ValueTimeLockVerifier';
 
-describe('TransactionProcessor', () => {
-  const config = require('../json/config-test.json');
-  let casClient: Cas;
+describe('TransactionProcessor', async () => {
+  const config = await import('../json/config-test.json');
```
isaacJChen, line 23: await import?

lib/ipfs/Ipfs.ts
thehenrytsai: No particular reason, will change it to private. What validation are you thinking of? Instance methods will throw if URI is invalid, which is okay, no?

lib/ipfs/Ipfs.ts
```diff
@@ -0,0 +1,163 @@
...
+
+    const fetchContentPromise = this.fetchContent(base58EncodedMultihashString, maxSizeInBytes);
+    const fetchResult = await Timeout.timeout(fetchContentPromise, this.fetchTimeoutInSeconds * 1000);
+
+    // Mark content as `not found` if any error is thrown while fetching.
+    if (fetchResult instanceof Error) {
+      // Log appropriately based on error.
+      if (fetchResult instanceof SidetreeError &&
+          fetchResult.code === IpfsErrorCode.TimeoutPromiseTimedOut) {
+        console.log(`'${base64urlEncodedMultihash}' not found on IPFS.`);
```
isaacJChen, line 94: Discussion: Do we want to log both the base64 CID and the base58 CID? I think it will help make debugging easier since we won't need to convert them using some other tool, we can just copy and paste.

lib/ipfs/Ipfs.ts
thehenrytsai: Will do.

lib/ipfs/Ipfs.ts
thehenrytsai: Neither did I, was forced to implement it as the API demands it.

Official spec: https://tools.ietf.org/html/rfc7578#section-4.1

lib/ipfs/Ipfs.ts
thehenrytsai: We log in `fetchContent` for the most part, but I think your suggestion of logging it here is better, will move it out.

lib/core/Core.ts
thehenrytsai: This was written specifically for our microservices, but I can see an argument for surfacing go-ipfs version here maybe? Though that opens up the question of if we should do the same for Bitcoin Core... happy to discuss this for a separate PR.

tests/core/TransactionProcessor.spec.ts
thehenrytsai: 'coz import returns a promise.... I looked at all imports used this way in our codebase, they all await. Let me know if I am smocking something I shouldn't be!

lib/ipfs/Ipfs.ts
thehenrytsai: Good suggestion... I tried to do it but only ended up doing it in happy path after a refactoring...

lib/ipfs/Ipfs.ts
mudiali: Yes that is fine. I was wondering whether we should stop initialization if the url is incorrect but then we're trying to move away from blocking the service start so it's ok.

lib/core/Core.ts
mudiali: Yea maybe the bitcoin, ipfs can call the dependent services and return their versions as well. Not a bad idea.

lib/ipfs/Ipfs.ts
```diff
@@ -122,7 +124,12 @@ export default class Ipfs implements ICas {
     let response;
     try {
       // e.g. 'http://127.0.0.1:5001/api/v0/cat?arg=QmPPsg8BeJdqK2TnRHx5L2BFyjmFr9FK6giyznNjdL93NL&length=100000'
-      const catUrl = url.resolve(this.uri, `/api/v0/cat?arg=${base58Multihash}&length=${maxSizeInBytes}`);
+      // NOTE: we pass max size + 1 to the API because the API will return upto the size given,
```
isaacJChen, line 127: nit: up to

Issue #760, opened 2020-06-22
Long-form resolution is not verifying `delta` maximum allowed size


thehenrytsai:

thehenrytsai:
Fixed.

PR #831, opened 2020-08-12
fix(ref-imp): #760 - Fixed long-form resolution not verifying delta size

thehenrytsai:


lib/common/SidetreeError.ts
```diff
@@ -4,7 +4,7 @@
  */
 export default class SidetreeError extends Error {
   constructor (public code: string, message?: string) {
-    super(message ? `${code}: ${message}` : code);
+    super(message ? message : code);
```
mudiali, line 7: This means that the when someone calls `sidetreeError.message`, it will not have the code there right. Not sure why would we make this change?

thehenrytsai: You are right `code` would not be in `message` with this change, but I ask myself, if they want to know the `code`, the more realistic scenario would be that they would just get it from `sidetreeError.code`, so no additional parsing required. So no real benefit for having the code in both properties. Also, `message` is intended to be a more detailed, user friendly message, so when such message is given, I feel we should just use it as intended and not do additional morphing.

tests/core/TransactionSelector.spec.ts
```diff
@@ -3,6 +3,8 @@ import ITransactionStore from '../../lib/core/interfaces/ITransactionStore';
 import MockTransactionStore from '../mocks/MockTransactionStore';
 import TransactionSelector from '../../lib/core/versions/latest/TransactionSelector';
 import TransactionModel from '../../lib/common/models/TransactionModel';
+import JasmineSidetreeErrorValidator from '../JasmineSidetreeErrorValidator';
+import ErrorCode from '../../lib/core/versions/latest/ErrorCode';
```
mudiali, line 7: Nit: imports not sorted.

thehenrytsai: Fixed.

Issue #336, opened 2019-11-26
Finalize and implement the DID resolution spec with `published` field as resolution response


thehenrytsai:
Assigning @csuwildcat to get the spec finalized before implementation can begin.

kdenhartog:
@csuwildcat would it be possible to return transaction metadata as method metadata so that the integrity of the resolution can be validated without needing to launch a node? I would think this would be an optional parameter that could be passed in the URL.

Something like this for a potential strawman solution

```
GET /did:sidetree:exKwW0HjS5y4zBtJ7vYDwglYhtckdO15JDt1j5F5Q0A?options=base64URL<{"methodMetadata": true }> HTTP/1.1
```

And the resulting object could be:

```JSON
{
  "didDocument": {
  "@context": "https://w3id.org/did/v1",
  "id": "did:sidetree:EiBJz4qd3Lvof3boqBQgzhMDYXWQ_wZs67jGiAhFCiQFjw",
  "publicKey": [{
    "id": "#key1",
    "type": "Secp256k1VerificationKey2018",
    "publicKeyHex": "029a4774d543094deaf342663ae672728e12f03b3b6d9816b0b79995fade0fab23"
  }],
  "service": [{
    "id": "IdentityHub",
    "type": "IdentityHub",
    "serviceEndpoint": {
      "@context": "schema.identity.foundation/hub",
      "@type": "UserServiceEndpoint",
      "instance": ["did:bar:456", "did:zaz:789"]
    }
  }]
},
"didMetadata": [{
  "header": "ewogICJvcGVyYXRpb24iOiAiY3JlYXRlIiwKICAia2lkIjogImtleTEiLAogICJhbGciOiAiRVMyNTZLIgp9",
[...]

PR #832, opened 2020-08-13
feat(ref-imp): #336 - added published field in method metadata

thehenrytsai:


lib/core/versions/0.9.0/RequestHandler.ts
```diff
@@ -190,22 +193,24 @@ export default class RequestHandler implements IRequestHandler {
 
   /**
    * Resolves the given long-form DID by resolving using operations found over the network first;
-   * if no operations found, the given create operation will is used to construct the DID state.
+   * if no operations found, the given create operation will be used to construct the DID state.
+   * 
+   * @returns [DID state, published]
    */
-  private async resolveLongFormDid (did: Did): Promise<DidState | undefined> {
+  private async resolveLongFormDid (did: Did): Promise<[DidState | undefined, boolean]> {
```
mudiali, line 200: So can't this boolean not be part of the DidState? Sorry I couldn't be part of the design discussion so feel free to continue if you want and fill me in later.

thehenrytsai: The discussion was precise on this. TL;DR is that we could include `published` in `DidState` but it isn't necessarily cleaner, because it would have to "undefined" until higher layer finds out if it is published or not.

thehenrytsai: Further discussion is with @mudiali, he raised a good point that we can potentially special case input to apply create operation if we wanted to `published` info in the `DidState`. Bottomline is that this current approach isn't necessarily bad because the `published` info isn't needed by operation processor. Since the other change will be much more involved, we decided keep the current approach. But we can follow the suggested pattern when we have new input required by the operation apply methods in the future.

Issue #838, opened 2020-08-24
Separate lock/relock transaction signing from the sidetree-anchor transaction


mudiali:
This is so that if there is a wallet implementation which requires separate inputs for the lock transactions then they have those inputs.

PR #839, opened 2020-08-25
refactor(ref-imp): Separating the freeze txn signing from the sidetree transaction signing

mudiali:
See issue #838 for motivations behind this change.

### DIFF
--- a/lib/bitcoin/BitcoinClient.ts
+++ b/lib/bitcoin/BitcoinClient.ts
@@ -155,15 +155,15 @@ export default class BitcoinClient {
   public async createLockTransaction (lockAmountInSatoshis: number, lockDurationInBlocks: number): Promise<BitcoinLockTransactionModel> {
     const unspentCoins = await this.getUnspentOutputs(this.bitcoinWallet.getAddress());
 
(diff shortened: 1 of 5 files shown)


lib/bitcoin/BitcoinClient.ts
```diff
@@ -565,13 +565,13 @@ export default class BitcoinClient {
...
 
-    return [freezeTransaction, freezeScript.toHex()];
+    return [freezeTransaction, freezeScript];
   }
 
   private async createSpendToFreezeTransaction (
     previousFreezeTransaction: BitcoreTransactionWrapper,
     previousFreezeDurationInBlocks: number,
-    newFreezeDurationInBlocks: number): Promise<[Transaction, string]> {
+    newFreezeDurationInBlocks: number): Promise<[Transaction, Script]> {
```
thehenrytsai, line 574: For my understanding: what's the benefit for returning `Script` instead of hex string here?

mudiali: Need to pass `Script` to the `IBitpayWallet` APIs. There is an opportunity to update the `IBitcoinWallet` to remove the requirement of bitcore-lib's objects but that's a separate issue that I opened sometime ago.

Issue #766, opened 2020-06-23
SIP 1 - Efficient Operation Mapping


csuwildcat:
```yaml
  SIP: 1
  Upgrade-Type: Hard Fork (to guarantee outcomes)
  Title: Efficient DID Operation Mapping
  Author: Daniel Buchner <daniel.buchner@microsoft.com>
  Comments-Summary: No comments yet.
  Comments-URI: https://github.com/decentralized-identity/sidetree/sips/1.md
  Status: Draft
  Created: 2020-06-23
```

## Summary

By segregating the proving data contained in the operation entries currently housed in the Anchor File and Map File (for Recovery, Deactivate, and Update operations), it is possible to realize a rather dramatic ~75% reduction in the minimum dataset required to trustlessly resolve DIDs.

The effect of moving this data to a segregated Proving File is that the Anchor File and Map Files become lightweight, spam-protected operation indexes, allowing for deferred acquisition of Proving Data in a JIT fashion, for nodes of various configurations.

## Motivation

These changes would make initialization of many node types faster, more efficient, and most importantly: operationally feasible for the average user-operator. Sustainable operation of nodes across consumer hardware is a key requirement for any decentralized network of this class, thus keeping network storage growth comfortably 'under the line' of the commodity storage technology cost curve and bandwidth growth curves is essential. While such curves lack precision, when one examines the trajectory of storage and bandwidth in reference to the waning cadence of the Kryder's Law and Edholm's Law doubling conjectures, it appears that a 2-3 terabytes per annum growth in the size of the minimum required dataset for a network is the top end of sustainability for a system that features peer-based replication of data and deferral of CPU intensive tasks until a JIT compilation/resolution phase.

## Requirements

* Target a minimum required dataset upper limit of 2-3 terabytes, assuming one year at a rate of 1000 operations per second.
* Push as much data out of the primary indexing files (Anchor and Map Files) as possible.

## Technical Proposal

The primary technical changes center around moving proving data out of the Anchor File and Map File, leaving those files to act as bare minimum indexes that enable a node to have global awareness of possible operations for any DID in the system. The proposed changes include the addition of two new intermediary files between the Anchor and Chunk Files. All changes to the existing Anchor and Map Files, as well as the new Proving Files, are as follows:

### Anchor File

The Anchor File would be modified in the following ways:

1. Add a new CAS URI link to a _Retained Proving File_, which contains the signed operation data that used to exist in the `recover` and `deactivate` operation entries.
[...]

PR #840, opened 2020-08-29
feat(ref-imp): #766 - Implemented hashing of public key as reveal value

thehenrytsai:


tests/fixtures/create/create.json
```diff
@@ -1,5 +1,5 @@
 {
   "type": "create",
-  "suffix_data": "eyJkZWx0YV9oYXNoIjoiRWlCWE00b3RMdVAyZkc0WkE3NS1hbnJrV1ZYMDYzN3hadE1KU29Lb3AtdHJkdyIsInJlY292ZXJ5X2NvbW1pdG1lbnQiOiJFaUM4RzRJZGJEN0Q0Q281N0dqTE5LaG1ERWFicnprTzF3c0tFOU1RZVV2T2d3In0",
-  "delta": "eyJ1cGRhdGVfY29tbWl0bWVudCI6IkVpQ0lQY1hCempqUWFKVUljUjUyZXVJMHJJWHpoTlpfTWxqc0tLOXp4WFR5cVEiLCJwYXRjaGVzIjpbeyJhY3Rpb24iOiJyZXBsYWNlIiwiZG9jdW1lbnQiOnsicHVibGljX2tleXMiOlt7ImlkIjoic2lnbmluZ0tleSIsInR5cGUiOiJFY2RzYVNlY3AyNTZrMVZlcmlmaWNhdGlvbktleTIwMTkiLCJqd2siOnsia3R5IjoiRUMiLCJjcnYiOiJzZWNwMjU2azEiLCJ4IjoieTlrenJWQnFYeDI0c1ZNRVFRazRDZS0wYnFaMWk1VHd4bGxXQ2t6QTd3VSIsInkiOiJjMkpIeFFxVVV0eVdJTEFJaWNtcEJHQzQ3UGdtSlQ0NjV0UG9jRzJxMThrIn0sInB1cnBvc2UiOlsiYXV0aCIsImdlbmVyYWwiXX1dLCJzZXJ2aWNlX2VuZHBvaW50cyI6W3siaWQiOiJzZXJ2aWNlRW5kcG9pbnRJZDEyMyIsInR5cGUiOiJzb21lVHlwZSIsImVuZHBvaW50IjoiaHR0cHM6Ly93d3cudXJsLmNvbSJ9XX19XX0"
+  "suffix_data": "eyJkZWx0YV9oYXNoIjoiRWlDTW5jb0xrODlIeWpUYzhRbUU0VVVoNFpzMlhiOGlGTWxBak51WFlDbjZCdyIsInJlY292ZXJ5X2NvbW1pdG1lbnQiOiJFaUF5VklubUZfbnpwYWVmeVdqNkJvTkdkWV85RlJOZWYtV0kzY3JySzZ2SXlBIn0",
```
troyronda, line 3: Was there a corresponding change to the spec that has resulted in the fixtures changing?

csuwildcat: This is part of SIP1, the thing where we need to remove the revealed public key JWK value from the Anchor File, which requires hashing it twice, do you don't have to put the actual data there.

OR13 (dismissed):
Changes to protocol markdown that is noted to be deprecated mixed with breaking change to implementation and test vectors.

"Double Hash" not described in spec.

Why do we need a "double hash"? This behavior should be documented. Before being merged in a way that breaks interoperability.

csuwildcat:
@OR13 kind of a tough spot here, because this is the first thing that needs to be done to address SIP 1, and the rest of SIP 1 is about a month out. So we can either reflect this one piece in the spec in some intermediary state, or update the spec to include all of SIP 1 now. I would hate to dump a bunch of time into updating this if it's an intermediary version of the spec that should not really exist.

csuwildcat:
A better way forward would be to have the latest spec simply state that it is not yet fully reflective of the updates in master of the reference. I can get started on all the SIP 1 changes to the spec, but I don't think we should blog all check-ins until all of SIP 1's facets are completed in both.

thehenrytsai:
> Why do we need a "double hash"? This behavior should be documented. Before being merged in a way that breaks interoperability.

I will let @csuwildcat comment on documentation. But essentially, this is the first stage of implementing SIP1. This change front loads the changes needed by the client side to perform double hash scheme to get the commit value as specified by SIP1, without needing to wait until server-side SIP1 is implemented.

tests/fixtures/create/create.json
thehenrytsai: Adding to @csuwildcat, this does impact the fixture/vector because with this PR commit value needs to be double hash of public key as opposed to single hash.

troyronda:
@csuwildcat @thehenrytsai It would be helpful to have some context on why SIP1 changes to double hash. Apologies if this was already explained elsewhere.

OR13:
@csuwildcat @thehenrytsai there is nothing wrong with using a branch called SIP-1 and merging it into master when its been fully documented....

Pushing incremental breaking changes to master, especially ones that impact test vectors used for interoperability is a really bad signal to the market about the stability of sidetree.

I suggest you use a branch other than master, fully document the changes that will be merged into the reference implementation via the 'spec' and then open a PR which merges the breaking changes along with the necessary documentation changes into master.

You can publish pre-release packages from branches other than master.

master should be stable and usable, and not contain partial implementations of new features... thats what feature branches are for :)

csuwildcat:
Another option is to create a branch of this state tagged with the current spec number and let master be master.

csuwildcat:
@troyronda we talked about the reason for double hashing the revealed jwk extensively on last week's Sidetree call, but I will take a few minutes on today's call to do so again.

OR13:
> Another option is to create a branch of this state tagged with the current spec number and let master be master.

https://thenewstack.io/dont-mess-with-the-master-working-with-branches-in-git-and-github/#:~:text=The%20master%20branch%20is%20meant,to%20work%20from%20the%20master.

lib/core/versions/latest/Multihash.ts
```diff
@@ -123,17 +141,42 @@ export default class Multihash {
...
+  private static verifyDoubleHash (content: Buffer, encodedMultihash: string): boolean {
+
+    try {
+      const expectedMultihashBuffer = Encoder.decodeAsBuffer(encodedMultihash);
+      const hashAlgorithmCode = Multihash.decode(expectedMultihashBuffer).algorithm;
+
+      const intermediateHashBuffer = Multihash.hashAsNonMultihashBuffer(content, hashAlgorithmCode);
+      const actualMultihashBuffer = Multihash.hash(intermediateHashBuffer, hashAlgorithmCode);
+
+      if (Buffer.compare(actualMultihashBuffer, expectedMultihashBuffer) !== 0) {
```
isaacJChen, line 175: return Buffer.compare(actualMultihashBuffer, expectedMultihashBuffer) == 0

lib/core/versions/latest/Multihash.ts
```diff
@@ -146,12 +189,12 @@ export default class Multihash {
...
-      const multihashBuffer = Encoder.decodeAsBuffer(encodedMultihash);
+      const expectedMultihashBuffer = Encoder.decodeAsBuffer(encodedMultihash);
+      const hashAlgorithmCode = Multihash.decode(expectedMultihashBuffer).algorithm;
 
-      const hashAlgorithmCode = Multihash.getHashAlgorithmCode(multihashBuffer);
-      const actualHashBuffer = Multihash.hash(content, hashAlgorithmCode);
+      const actualMultihashBuffer = Multihash.hash(content, hashAlgorithmCode);
 
-      if (Buffer.compare(actualHashBuffer, multihashBuffer) !== 0) {
+      if (Buffer.compare(actualMultihashBuffer, expectedMultihashBuffer) !== 0) {
```
isaacJChen, line 197: Same simplification as above.

thehenrytsai:
Hey @OR13,

In the WG meeting, the WG have:

1. Approved SIP1.
1. Approved this PR to be merged in to allow downstream clients/services to immediately start generating SIP1 compatible DIDs going forward without waiting on rest of the SIP1 specification/implementation.
1. Agreed that SIP1 spec must be immediately be worked on, finalized in the next week. A separate spec PR will be sent out.
1. Agreed that remaining/actual SIP1 implementation will follow after the SIP1 spec is checked in.

Thanks,
Henry = ]

OR13:
Master should be marked as unstable for this repo, I remain opposed to incrementally merging partial implementation of features into master, especially if they are wildly out of sync with the spec... it makes everyones jobs harder.... 

However, as I am no long a chair / editor of the wg I cannot block any of this, I am just giving you my opinion, as a developer who has to maintain interop with this repo.

The CODEOWNERS for this repo, if there where any, could merge over my objection... I suggest someone set that up... here is an example from the W3C CCG https://github.com/w3c-ccg/universal-wallet-interop-spec/blob/master/CODEOWNERS

isaacJChen (approved):

thehenrytsai:
@OR13, filed #844 to add CODEOWNERS file in a separate PR. Agree with the point about not merging in broken code/features into `master` in principle, and I believe most of us are following that principle. In this particular case, since SIP1 is officially approved, the WG agreed that it is beneficial for implementer to be aware and start generating Sidetree v1.0 compatible DIDs ASAP, so we agree to check in this PR is on its own. Your point on committed spec lagging behind implementation is taken, as such a PR for SIP1 will be sent out in a week's time or so, SIP1 implementation PR will follow.

The `master` is considered unstable by default currently, in fact Tobias updated the pipeline such that an unstable release is carved out every time a merge is made into master (unstable version can be seen on repo landing page). If you have specific steps to make this more clear, please let me know, I am certainly not as well-versed as you when it comes to github repo/branch management, so could use your help there! Not to say this paradigm won't change as we stabilize and approach spec v1.0, that is certainly the goal.

With above said, I will merge this PR in.

Issue #820, opened 2020-07-27
Deactivate operation potentially not working.


gjgd:
Looking at this code, it looks like only create update and recover operations are being processed in the TransactionProcessor:
https://github.com/decentralized-identity/sidetree/blob/master/lib/core/versions/latest/TransactionProcessor.ts#L206-L266

It seems that a deactivate operation would just be ignored here. I'm trying to find an integration test that checks that deactivate operations are working but I don't see any. We need to have integrations tests that verify all operations are working. 

Tests such as the one below are not enough because they don't use the TransactionProcessor: https://github.com/decentralized-identity/sidetree/blob/05af0c56d993632311f640290a286e2e5b9df12d/tests/core/Resolver.spec.ts#L120

gjgd:
I can confirm that there are no integration tests for the deactivate operation by adding this check [here](https://github.com/decentralized-identity/sidetree/blob/master/lib/core/versions/latest/TransactionProcessor.ts#L224)

```js
    if (deactivateOperations.length > 0) {
      throw new Error('This code path was hit');
    }
```

No error is thrown when running `npm test`, so this code path was never hit.

With the same method, I was able to verify that the recover operation is not tested either.

thehenrytsai:
Fixed, added a test to prevent regression.

PR #854, opened 2020-09-10
fix(ref-imp): #820 - Deactivate not working

thehenrytsai:


lib/core/versions/0.11.0/TransactionProcessor.ts
```diff
@@ -222,37 +223,30 @@ export default class TransactionProcessor implements ITransactionProcessor {
...
+    // NOTE: The last set of `operations` are deactivates, they don't have `delta` property.
     const anchoredOperationModels = [];
     for (let i = 0; i < operations.length; i++) {
       const operation = operations[i];
+      const operationJsonString = operation.operationBuffer.toString();
+      const operationObject = await JsonAsync.parse(operationJsonString);
+      operationObject.type = operation.type;
+
+      // Add `delta` property read from chunk file to each operation if chunk file exists.
+      if (operation.type === OperationType.Deactivate) {
```
isaacJChen, line 238: nit: can't we just use 1 if statement, the comments should explain the logic well enough that we don't need an empty if block and then else if

isaacJChen (approved):

isaacJChen (approved):

PR #860, opened 2020-09-14
fix(ref-imp): make recover and create to always advance commitment

isaacJChen:


tests/core/OperationProcessor.spec.ts
```diff
@@ -459,6 +462,15 @@ describe('OperationProcessor', async () => {
...
         expect(newDidState!.nextRecoveryCommitmentHash).toEqual(nextRecoveryCommitmentHash);
       });
+
+      it('should apply the create operation by with { } document if encoded data and suffix data do not match', async () => {
+        isValidHashSpy.and.returnValue(false);
+        const createOperationData = await OperationGenerator.generateAnchoredCreateOperation({ transactionTime: 1, transactionNumber: 1, operationIndex: 1 });
+        const newDidState = await operationProcessor.apply(createOperationData.anchoredOperationModel, undefined);
+        expect(newDidState!.lastOperationTransactionNumber).toEqual(1);
+        expect(newDidState!.document).toEqual({ });
+        expect(newDidState!.nextRecoveryCommitmentHash).not.toEqual(nextRecoveryCommitmentHash);
```
thehenrytsai, line 472: This check is invalid because `nextRecoveryCommitmentHash` will never be the same as `newDidState.nextRecoveryCommitmentHash`. I think it's much more direct to check the commitment hash against the double hash of the `createOperationData`'s recovery key.

tests/core/OperationProcessor.spec.ts
```diff
@@ -459,6 +462,15 @@ describe('OperationProcessor', async () => {
         expect(newDidState!.document).toBeDefined();
         expect(newDidState!.nextRecoveryCommitmentHash).toEqual(nextRecoveryCommitmentHash);
       });
+
+      it('should apply the create operation by with { } document if encoded data and suffix data do not match', async () => {
```
thehenrytsai, line 466:
```suggestion
      it('should apply the create operation with { } as document if encoded data and suffix data do not match', async () => {
```

tests/core/OperationProcessor.spec.ts
isaacJChen: Good point

lib/core/versions/latest/OperationProcessor.ts
```diff
@@ -103,37 +103,39 @@ export default class OperationProcessor implements IOperationProcessor {
...
+      // update the commitment hash regardless
+      newDidState.nextUpdateCommitmentHash = delta.updateCommitment;
+      try {
         document = DocumentComposer.applyPatches(document, delta.patches);
+        newDidState.document = document;
+      } catch (error) {
+        const didUniqueSuffix = anchoredOperationModel.didUniqueSuffix;
+        const transactionNumber = anchoredOperationModel.transactionNumber;
+        console.debug(
+          `Unable to apply document patch in transaction number ${transactionNumber} for DID ${didUniqueSuffix}: ${SidetreeError.stringify(error)}.`);
```
thehenrytsai, line 135: The original debug log will sound like the entire operation is not applied. Can we update this it indicate that this is "partially applied" with the next commitment hash?

tests/core/OperationProcessor.spec.ts
```diff
@@ -463,13 +463,13 @@ describe('OperationProcessor', async () => {
         expect(newDidState!.nextRecoveryCommitmentHash).toEqual(nextRecoveryCommitmentHash);
       });
 
-      it('should apply the create operation by with { } document if encoded data and suffix data do not match', async () => {
+      fit('should apply the create operation with { } as document if encoded data and suffix data do not match', async () => {
```
thehenrytsai, line 466:
```suggestion
      it('should apply the create operation with { } as document if encoded data and suffix data do not match', async () => {
```

lib/core/versions/latest/OperationProcessor.ts
```diff
@@ -132,7 +132,8 @@ export default class OperationProcessor implements IOperationProcessor {
         const didUniqueSuffix = anchoredOperationModel.didUniqueSuffix;
         const transactionNumber = anchoredOperationModel.transactionNumber;
         console.debug(
-          `Unable to apply document patch in transaction number ${transactionNumber} for DID ${didUniqueSuffix}: ${SidetreeError.stringify(error)}.`);
+          `Partial update on next commitment hash applied because:
```
thehenrytsai, line 135: This will print unwanted spaces in log. Consider:

```suggestion
          `Partial update on next commitment hash applied because:` +
```

lib/core/versions/latest/OperationProcessor.ts
```diff
@@ -132,7 +132,8 @@ export default class OperationProcessor implements IOperationProcessor {
...
+          Unable to apply delta patches for transaction number ${transactionNumber} for DID ${didUniqueSuffix}: ${SidetreeError.stringify(error)}.`);
```
thehenrytsai, line 136: This will print unwanted spaces in log. Consider:
```suggestion
          `Unable to apply delta patches for transaction number ${transactionNumber} for DID ${didUniqueSuffix}: ${SidetreeError.stringify(error)}.`);
```

thehenrytsai (approved):

PR #872, opened 2020-10-01
Migrate tslint to eslint

tiago154:
#865 

Hi. I added eslint and put some rules according to the project.

So to validate the errors, run `npm run lint`

There is also (`npm run lint:fix`), where it is possible to apply corrections automatically.


thehenrytsai:
This is awesome! Thanks a lot! Taking a look now!

.eslintrc.json
```diff
@@ -0,0 +1,43 @@
...
+    "parser": "@typescript-eslint/parser",
+    "parserOptions": {
+        "ecmaVersion": 11,
+        "sourceType": "module"
+    },
+    "plugins": [
+        "@typescript-eslint"
+    ],
+    "rules": {
+        "no-unused-vars": "off",
```
thehenrytsai, line 23: Sanity: should this be on? Or the typescript version below take precedence?

.eslintrc.json
```diff
@@ -0,0 +1,43 @@
...
+        ],
+        "max-lines": [
+            "error",
+            160
+        ],
+        "func-style": [
+            "error",
+            "expression"
+        ],
+        "semi": "off",
```
thehenrytsai, line 40: Should be turned on?

package.json
```diff
@@ -31,11 +31,20 @@
...
     "conventional-changelog-cli": "2.0.35",
     "copyfiles": "2.3.0",
     "cz-conventional-changelog": "3.1.0",
+    "eslint": "^7.10.0",
+    "eslint-config-standard": "^14.1.1",
+    "eslint-plugin-import": "^2.22.1",
+    "eslint-plugin-node": "^11.1.0",
+    "eslint-plugin-prefer-arrow": "^1.2.2",
+    "eslint-plugin-promise": "^4.2.1",
+    "eslint-plugin-standard": "^4.0.1",
```
thehenrytsai, line 47: Just sanity: are all these dependencies needed?

.eslintrc.json
tiago154: This item is actually related to line 24. It is a way for the typescript not to complain about imports that are used only for typing.

Example:

```typescript
import { AnyType } from 'model'

Promise<AnyType[]>
```

.eslintrc.json
tiago154: You are right. I forgot to change here.
Now I turned on the rule.

package.json
tiago154: Yes. I used the initial eslint file configurator. It already installs all the development dependencies necessary for the rules to work in typescript.

thehenrytsai (approved):

PR #877, opened 2020-10-07
Handle estimatesmartfee errors

lionello:
This PR adds a new config setting `estimatedFeeInBitcoinsPerKB: number | undefined` to be used when the `estimatesmartfee` RPC call fails. There are plenty of occasions when that RPC call fails, most prominently when using bitcoind in `-regtest` or even on testnet where there's less traffic.

Smaller changes in this PR include renames of private APIs (`Kb` to `KB` to signify kilobyte), typos, and existing lint errors.

### DIFF
--- a/lib/bitcoin/BitcoinClient.ts
+++ b/lib/bitcoin/BitcoinClient.ts
@@ -42,7 +42,8 @@ export default class BitcoinClient {
     bitcoinWalletOrImportString: IBitcoinWallet | string,
     private requestTimeout: number,
     private requestMaxRetries: number,
-    private sidetreeTransactionFeeMarkupPercentage: number) {
+    private sidetreeTransactionFeeMarkupPercentage: number,
+    private estimatedFeeSatoshiPerKB?: number) {
 
     if (typeof bitcoinWalletOrImportString === 'string') {
       console.info('Creating bitcoin wallet using the import string passed in.');
@@ -336,7 +337,7 @@ export default class BitcoinClient {
   /**
    * Gets the transaction fee of a transaction in satoshis.
    * @param transactionId the id of the target transaction.
-   * @returns the transaction fee.
+   * @returns the transaction fee in satoshis.
    */
   public async getTransactionFeeInSatoshis (transactionId: string): Promise<number> {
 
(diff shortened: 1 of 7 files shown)


lib/bitcoin/IBitcoinConfig.ts
```diff
@@ -11,6 +11,7 @@ export default interface IBitcoinConfig {
   bitcoinRpcUsername: string | undefined;
   bitcoinRpcPassword: string | undefined;
   bitcoinWalletOrImportString: IBitcoinWallet | string;
+  estimatedFeeInBitcoinsPerKB: number | undefined;
```
thehenrytsai, line 14: I think it makes sense to be in satoshis so we are not dealing with small fraction numbers.
Also "default" is important because it conveys the message that this maybe overriden which is indeed the case in the implementation.
```suggestion
  defaultEstimatedFeePerKBInSatoshis: number | undefined;
```

lib/bitcoin/BitcoinClient.ts
```diff
@@ -418,6 +419,21 @@ export default class BitcoinClient {
     return BitcoinClient.convertBtcToSatoshis(feerateInBtc);
   }
 
+  /** Get the current estimated fee from RPC or return the stored estimate */
+  private async getEstimatedFeeInSatoshisPerKB (): Promise<number> {
```
thehenrytsai, line 423: Style: this method doesn't just "get", it also "updates" stored estimate as a side effect, so I think we should flip the method name, and make the "get" as a side effect if you really want to keep it as a return value.
```suggestion
  private async updateEstimatedFeeInSatoshisPerKB () {
```
or
```suggestion
  private async updateEstimatedFeeInSatoshisPerKB (): Promise<number> {
```

Alternatively you don't have to change the method name, but don't update the cached estimate, update it after this function call.

thehenrytsai (changes requested):
Looks good overall, a couple of suggestions.

lib/bitcoin/IBitcoinConfig.ts
lionello: I agree on both accounts, but re using satoshis over bitcoins: `valueTimeLockTransactionFeesAmountInBitcoins` is in bitcoins so rather they stay in the same unit, no?

lionello: Another reason why I think it makes sense to stick to bitcoin in the config file is that the RPC call this is referring to also returns in bitcoin. During ION setup you could use the RPC call to get the estimate and put it in the config file as is.

lionello: I finally went with `defaultFeePerKBInSatoshis` which I think better describes what this ends up doing. "Default estimated fee" sounded too vague IMHO.

lionello:
Addressed comments and rebased to latest master.

thehenrytsai (approved):

Issue #805, opened 2020-07-15
Should sidetree rely on a styleguide for JSON?


OR13:
https://github.com/json-api/json-api/issues/1255
https://www.w3schools.com/js/js_conventions.asp
https://google.github.io/styleguide/jsoncstyleguide.xml
https://jsonapi.org/recommendations/#naming

OR13:
I suggest we should just do whatever json-api says to do.... going our own way here is likely to result in endless issues... I would much rather just say.... see our style guide, we didn't invent it.

OR13:
This was discussed previously here: https://github.com/decentralized-identity/sidetree/issues/749

OR13:
I think the action here is to make sure we document why our JSON structures do or do not conform to a style guide.

csuwildcat:
We already had consensus agreement on casing with PRs reviewed and accepted long ago. It is a bad practice, and harmful to all those who actually are impacted by the code churn, to not let our decisions as a group be final, within reason, serious bugs/issues withstanding. I would oppose a change now, because there are valid arguments to be made againstUsingCamelCase vs using_snake_case, and because this represents a burden on those doing the actual work, with literally 0 technically effective advantage.

OR13:
Lets get our decision to handle JSON differently than google and json-api documented in the spec and we can close this issue :) 

The more I think about this, the more I think we made a mistake... read the original issue: https://github.com/decentralized-identity/sidetree/issues/749

We basically decided to close it, because we didn't want to break compatibility... but with things like SIP-1 / removing `replace` a possibility, I think its fair to reopen this issue, and seek a better solution.

tplooker:
+1 to having consistency, as @wyc put it beautifully in slack this morning

> i will accept a consistent abomination over abominable inconsistency

The context of the statement was around debating which is better snake_case kebab-case or camelCase, for which people have endless opinions on.

At present in this spec we use kebab-case and snake_case, I really dont care what we use but it should be just the one.

csuwildcat:
I want to be very clear: all properties in the spec are snake case. The one thing that isn't is a string value (not property) in the patch data. I didn't realize we were imposing anything on values, and really hope we don't adjudicate this with real risks in plat related to code/operation churn. I would ask that any nit-level changes from here on out be accompanied by PRs that change literally everything affected in the reference implementation. I believe requiring more skin in the game for changes that cause real code, operational, and deployment impacts will probably help focus our efforts on the things that are the most important.

[...]

PR #878, opened 2020-10-08
Ische/sip2

isaacJChen:
All SIP 2 changes excluding long form backwards compatibility.


tests/generators/TestVectorGenerator.ts
```diff
@@ -0,0 +1,55 @@
...
+        )
+        
+        // derive a recover operation from the create operation
+        const [newRecoveryPublicKey] = await Jwk.generateEs256kKeyPair();
+        const [newSigningPublicKey] = await OperationGenerator.generateKeyPair('keyAfterRecover');
+
+        const [documentKey] = await OperationGenerator.generateKeyPair('newDocumentKey');
+        const newServiceEndpoints = OperationGenerator.generateServiceEndpoints(['newId']);
+  
+        const recoverOperationRequest = await OperationGenerator.generateRecoverOperationRequest(
```
thehenrytsai, line 41: Sanity: what about deactivate operation?

tests/generators/OperationGenerator.ts
```diff
@@ -129,7 +129,8 @@ export default class OperationGenerator {
     recoveryPublicKey?: JwkEs256k,
     updatePublicKey?: JwkEs256k,
     otherPublicKeys?: PublicKeyModel[],
-    serviceEndpoints?: ServiceEndpointModel[]) {
```
thehenrytsai, line 132: Sanity: Looks like through the recent refactoring, `recoveryPublicKey` and `updatePublicKey` is no longer needed to be passed in?

lib/core/versions/latest/UpdateOperation.ts
```diff
@@ -112,9 +113,12 @@ export default class UpdateOperation implements OperationModel {
...
       }
-
-      encodedDelta = operationObject.delta;
-      delta = await Operation.parseDelta(encodedDelta);
+      Operation.validateDelta(operationObject.delta);
+      delta = {
+        patches: operationObject.delta.patches,
+        updateCommitment: operationObject.delta.update_commitment
+      };
+      encodedDelta = Encoder.encode(JsonCanonicalizer.canonicalizeAsBuffer(operationObject.delta));
```
thehenrytsai, line 121: Design: I think/hoped that all of `encodedDelta` (and `encodedSuffixData`) can now be removed with SIP2. The point of keeping these "encoded" form was really just for hash checks, but we no longer need them (with exception to perhaps the create operation where old long-form DID still needs this old way of hash validation).

lib/core/versions/latest/UpdateOperation.ts
```diff
@@ -112,9 +113,12 @@ export default class UpdateOperation implements OperationModel {
       if (operationObject.type !== OperationType.Update) {
         throw new SidetreeError(ErrorCode.UpdateOperationTypeIncorrect);
       }
-
-      encodedDelta = operationObject.delta;
-      delta = await Operation.parseDelta(encodedDelta);
+      Operation.validateDelta(operationObject.delta);
+      delta = {
+        patches: operationObject.delta.patches,
+        updateCommitment: operationObject.delta.update_commitment
```
OR13, line 119: this kind of casing makes me so sad :(

lib/core/versions/latest/models/AnchorFileModel.ts
```diff
@@ -6,7 +6,10 @@ export default interface AnchorFileModel {
   map_file_uri: string;
   operations: {
     create?: {
-      suffix_data: string;
+      suffix_data: {
```
OR13, line 9: seems like this type should be imported from somewhere common.

lib/core/versions/latest/models/ChunkFileModel.ts
```diff
@@ -2,5 +2,5 @@
  * Defines the external Chunk File structure.
  */
 export default interface ChunkFileModel {
-  deltas: string[];
+  deltas: object[];
```
OR13, line 5: this is weak typing, prefer exact type.

tests/generators/TestVectorGenerator.ts
```diff
@@ -0,0 +1,55 @@
...
+import OperationGenerator from "./OperationGenerator";
+
+export default class TestVectorGenerator {
+    /**
+     * Generate and prints out test vectors for operation requests
+     */
+    public static async generateOperationVectors() {
+
+        // generate a create operation request
+        const createOperationData = await OperationGenerator.generateCreateOperation();
```
OR13, line 12: I assume this is random, I suggest you modify your operation generator to be deterministic, see:

https://github.com/transmute-industries/sidetree.js/blob/master/packages/core/src/__test__/test-fixture-generation.test.ts

We generate all fixtures deterministically and compare them too committed values.

tests/fixtures/update/create.json
```diff
@@ -1,5 +1,40 @@
...
+  "delta": {
+    "update_commitment": "EiDuOprunhX2vM9ZOSY4M-C7QUQDGSe_g3AKJGoOPMpevA",
+    "patches": [
+      {
+        "action": "replace",
+        "document": {
+          "public_keys": [
+            {
+              "id": "signingKey",
+              "type": "EcdsaSecp256k1VerificationKey2019",
```
OR13, line 16: Prefer JsonWebKey2020

tests/fixtures/recover/recover.json
```diff
@@ -1,6 +1,38 @@
...
+  "signed_data": "eyJhbGciOiJFUzI1NksifQ.eyJkZWx0YV9oYXNoIjoiRWlEUkpZQ2NmdGxTX3V5eG52VjVQT3dKcFkzdVc3THVRdU4tSElURXJMemJhUSIsInJlY292ZXJ5X2tleSI6eyJrdHkiOiJFQyIsImNydiI6InNlY3AyNTZrMSIsIngiOiJ2Q3ZPZnUzamVGakVOM2VaYXFJSEhESmExQWNTWWNoaDFsOTd5TzdZeGE0IiwieSI6IjVmamNzV1JkalNOeENtd0VCRFpuSndLdUNnZ2lidE8xV2g3U3VfX3laZjAifSwicmVjb3ZlcnlfY29tbWl0bWVudCI6IkVpRHZOXzI2WUJrakdwWndEUFI4U2t2Y05FTGdZT3AzZk83b1pVZW51T3EzV1EifQ.nfzlXC3SD2s4K6U1dzFM5IH-vg4N5fSMCK8SQuyZRPovnMPGOBeo72RgaSd1O8q2bZktkW4aBVG5XsPiT4kNJQ",
+  "delta": {
+    "patches": [
+      {
+        "action": "replace",
+        "document": {
+          "public_keys": [
+            {
+              "id": "newDocumentKey",
+              "type": "EcdsaSecp256k1VerificationKey2019",
```
OR13, line 13: prefer to use JsonWebKey2020 for JWK keys.

tests/fixtures/legacyLongFormDid/resultingDocument.json
```diff
@@ -0,0 +1,40 @@
+{
+    "@context": "https://www.w3.org/ns/did-resolution/v1",
+    "didDocument": {
+        "id": "did:ion:EiC5-1uBg-YC2DvQRbI6eihDvk7DOYaQ08OB0I3jCe9Ydg?-ion-initial-state=eyJkZWx0YV9oYXNoIjoiRWlBbExNMC1qem1DWi1FcElVZ0laQ2piWk5yMDFfVVBMbnd5MHdfT3I0Rks0dyIsInJlY292ZXJ5X2NvbW1pdG1lbnQiOiJFaUJDNGhTMVVHeVNnTmYzbWFMdnNKRUpxX05aQUlKa0pndTNKMTJMeGNESE93In0.eyJwYXRjaGVzIjpbeyJhY3Rpb24iOiJyZXBsYWNlIiwiZG9jdW1lbnQiOnsicHVibGljX2tleXMiOlt7ImlkIjoiYW55U2lnbmluZ0tleUlkIiwiandrIjp7ImNydiI6InNlY3AyNTZrMSIsImt0eSI6IkVDIiwieCI6ImFHc01HMHU5Rlg2STU0cGVJS3FZb2tqblFQR2hMVVlUT1FOYzNuT3ZFMVEiLCJ5IjoiZmppbHFoZVdRWWtITkU3MHNoTVJ5TURyWnA4RUdDZkVfYUwzaC15Sm1RQSJ9LCJwdXJwb3NlIjpbImF1dGgiLCJnZW5lcmFsIl0sInR5cGUiOiJFY2RzYVNlY3AyNTZrMVZlcmlmaWNhdGlvbktleTIwMTkifV0sInNlcnZpY2VfZW5kcG9pbnRzIjpbeyJlbmRwb2ludCI6Imh0dHA6Ly9hbnkuZW5kcG9pbnQiLCJpZCI6ImFueVNlcnZpY2VFbmRwb2ludElkIiwidHlwZSI6ImFueVR5cGUifV19fV0sInVwZGF0ZV9jb21taXRtZW50IjoiRWlERkM2RE9Ed0JNeG5kX19oMTFSeDRObjFlOHpubFlPUjJhLVBqeUNva2NGZyJ9",
```
OR13, line 4: `did:ion` vs `did:sidetree`... I think we expect `did:sidetree` here...

tests/fixtures/create/resultingDocument.json
```diff
@@ -34,7 +34,7 @@
   },
   "methodMetadata": {
     "published": true,
-    "recoveryCommitment": "EiAyVInmF_nzpaefyWj6BoNGdY_9FRNef-WI3crrK6vIyA",
-    "updateCommitment": "EiCjkDGDDOnuZq4Kiw2icudIN95boDMrGdYx4pz8bBt2bw"
+    "recoveryCommitment": "EiAzN_y4WBLpZsjRXg8GbpRtXAhddATz-vmq3AxMOV2Huw",
```
OR13, line 37: is this casing correct?

tests/generators/TestVectorGenerator.ts
```diff
@@ -0,0 +1,55 @@
...
+          createOperationData.createOperation.didUniqueSuffix,
+          createOperationData.recoveryPrivateKey,
+          newRecoveryPublicKey,
+          newSigningPublicKey,
+          newServiceEndpoints,
+          [documentKey]
+        );
+  
+  
+        console.log(JSON.stringify(createOperationData.operationRequest));
```
OR13, line 51: You may want to just write these directly to disk, instead of logging them.

lib/core/versions/latest/UpdateOperation.ts
thehenrytsai: @OR13, are you talking about the mix of snake casing and camel casing? If so, remember I fought hard on having just camel casing and even made the case that DID core spec is using camel casing in WG meeting back in like April or March and lost?

OR13: yep, we all lost... the sidetree spec still does not have a "styleguide" which defines this.

I have already hit a bunch of casing related bugs because of this.

But I assume its too late to fix now.

csuwildcat: I cannot begin to express the level of frustration I faced when OTHERS supported the current casing, voted to use it, then later wanted to change it to something else, which I opposed simply for stability sake. But if people want to play faux-amnesia revisionist history games and willfully forget that, screw it, just change it now and break our client teams, but I'd like to set a meeting where you own the change, given you all agreed that the the current casing and later wavered after it was accepted and implemented. I'm getting our whole team on the call to witness the hat-in-hand mea culpa, because I'm tired of being the messenger for a group of people who can't vote on something and stick to it.

tests/generators/TestVectorGenerator.ts
isaacJChen: Lol good point. This change didn't touch deactivate so I forgot about it. Will add

lib/core/versions/latest/UpdateOperation.ts
OR13: https://github.com/decentralized-identity/sidetree/issues/805

I provided links to API styleguides (Google and Microsoft), you provided examples of APIs that don't use the styleguides I linked (GitHub and Slack).

Regardless of which "path we take" the WG needs a styleguide, or stuff like this is going to keep coming up.

At this point, if you are willing to use Google and Microsoft JSON API styleguides, I am ok admitting that I was wrong to ever compromise with you, and that I should never have accepted the casing convention you proposed.

I admit that changing this now will create a big headache, but IMO its worth having a style that matches google and miscrosoft.

I admit that I did accept snake case, and that I am now rejecting it

I also want to point out that sidetree has not hit a 1.0 release, and so we should be allowed to make breaking changes.

I am happy to take a call with your team at Microsoft to explain why its nice for us to follow Microsoft's JSON API styleguide.

tests/fixtures/legacyLongFormDid/resultingDocument.json
isaacJChen: Good catch, thank!

lib/core/versions/latest/UpdateOperation.ts
OR13: @csuwildcat we had many calls on this, and I have documented my position clearly many times, both verbally and in writing.... one more time:

My preferences are:

1. Sidetree has a JSON API styleguide that it uses to make decisions
2. The styleguide recommends camelCase (among other things, see Microsoft and Google JSON API style guides)

thehenrytsai: My default stance is that snake casing was decided by consensus vote, so we now need to stick with it, as multiple implementers are down the road and relying on it. My comment was more expressing my frustration from months back, but not suggesting we change it, certainly not in this PR.

tests/fixtures/create/resultingDocument.json
isaacJChen: Ya, everything in the resulting document should be camel case, unless my understanding is wrong.

OR13: This is a resolver response (external data)... this is exactly my point.

lib/core/versions/latest/UpdateOperation.ts
OR13: @thehenrytsai can you find a styleguide that aligns with this decision and add it to the README?

tests/generators/TestVectorGenerator.ts
isaacJChen: I logged them so you have to intentionally copy and paste them into the fixtures, I don't want someone to run this and accidentally change them. I see your point for convenience, but I think the safety and making it more intentional is better. What do you think?

OR13: you can set a flag, like `WRITE_FIXTURES_TO_DISK` and make someone change it to update them on disk... thats what we do.

tests/generators/TestVectorGenerator.ts
isaacJChen: Good idea. Will make an issue for it. Not going to do it in this pr since the main issue addressing here is SIP 2, it just so happen that fixture has to be updated.

isaacJChen: https://github.com/decentralized-identity/sidetree/issues/879

tests/generators/TestVectorGenerator.ts
isaacJChen: Sounds good, will add the flag.

tests/fixtures/update/create.json
isaacJChen: Not going to address this in this pr since this pr is purely for SIP 2 changes.

lib/core/versions/latest/models/ChunkFileModel.ts
isaacJChen: It is intentionally object because we don't know for sure the deltas are valid delta objects yet, we just know they need to be objects (similar to before we didn't know if they were valid base64encoded strings that can decode into delta, we just know they are strings). There are more checks later on in the process that validates they are delta objects during resolution time.

OR13: I see, that makes sense... a comment to this effect may be helpful to future readers.

isaacJChen: Note: We have internal discussion to see if we want to check the type during observing time, we are worried about the performance, but if we decide to do it, it can easily be versioned in.

OR13: hmm yep, wondering what happens when its bad... prefer to throw stuff out as soon as possible if it's bad. may want to link to an issue for future discussion related to this.

isaacJChen: Ok, will make an issue for it and add comment in code.

isaacJChen: https://github.com/decentralized-identity/sidetree/issues/880

lib/core/versions/latest/UpdateOperation.ts
isaacJChen: Since this discussion is progressing in slack, I will resolve this comment. We can make an issue for having a style guide and bring it up in a WG meeting if we choose to do so :)

lib/core/versions/latest/models/AnchorFileModel.ts
isaacJChen: This structure is not used by anything else so I think leaving it here makes sense. We can pull it out when there's use case for it in the future.

lib/core/versions/latest/OperationProcessor.ts
```diff
@@ -113,8 +114,12 @@ export default class OperationProcessor implements IOperationProcessor {
       lastOperationTransactionNumber: anchoredOperationModel.transactionNumber
     };
 
-    // Ensure actual delta hash matches expected delta hash.
-    const isMatchingDelta = Multihash.isValidHash(operation.encodedDelta, operation.suffixData.deltaHash);
+    // Verify the delta hash against the expected delta hash.
+    const expectedDeltaHash = operation.delta ? Encoder.encode(JsonCanonicalizer.canonicalizeAsBuffer({
```
thehenrytsai, line 118: Suggesting two changes:
1. Output is not a hash, rename.
2. Should not encode just to be able to use `Multihash.isValidHash`, make `Multihash.verify()` public and use that. Though the name should be better now that it's public. Or maybe even better introduce a new method `Multihash.canonicalizeAndVerifyHash()`.
```suggestion
    const deltaPayload = operation.delta ? JsonCanonicalizer.canonicalizeAsBuffer({
```

lib/core/versions/latest/AnchorFile.ts
```diff
@@ -166,8 +166,12 @@ export default class AnchorFile {
     }
 
     const createOperations = createOperationArray.map(operation => {
+      const suffixData = { delta_hash: operation.suffixData.deltaHash, recovery_commitment: operation.suffixData.recoveryCommitment, type: operation.suffixData.type };
+      if (suffixData.type === undefined) {
+        delete suffixData.type;
+      };
```
thehenrytsai, line 172: I think this is unnecessary.

thehenrytsai (approved):

PR #881, opened 2020-10-10
Implement conformance to a widely recognized API/style guide

OR13:
- [x] suffix_data -> suffixData
- [x] recovery_commitment -> recoveryCommitment
- [x] update_commitment -> updateCommitment
- [x] delta_hash -> deltaHash
- [x] public_keys -> publicKey ...consider using "verificationMethod instead, see https://w3c.github.io/did-spec-registries/#publickey
- [x] service_endpoints -> service https://w3c.github.io/did-core/#service-endpoints
- [x] purpose -> verificationRelationship https://w3c.github.io/did-core/#verification-relationships
- [x] jwk -> publicKeyJwk https://w3c.github.io/did-core/#key-types-and-formats
- [x] verificationRelationship cleanup
- [x] fix controller (for linked data proofs)


tests/generators/OperationGenerator.ts
```diff
@@ -144,23 +144,60 @@ export default class OperationGenerator {
...
+      longFormDid,
+      shortFormDid,
+      didUniqueSuffix
+    };
+  }
+
+  /**
+   * Generates a long from from create operation data.
+   */
+  public static async longFormFromCreateOperationData (createOperationData:any, network?: string) {
```
isaacJChen, line 181: Nit: Should have space between : and any

tests/core/RequestHandler.spec.ts
```diff
@@ -139,39 +138,43 @@ describe('RequestHandler', () => {
 
   it('should resolve LEGACY long form did from test vectors correctly', async () => {
```
isaacJChen, line 139: Legacy long form cannot switch to camel case. There are clients of ION that are dependent on it not breaking until they fully switch over to the new form. The legacy code path is separated out so we can duplicate functions and add TODO to delete them once everyone migrates over.

lib/core/versions/latest/CreateOperation.ts
```diff
@@ -186,7 +186,7 @@ export default class CreateOperation implements OperationModel {
       throw new SidetreeError(ErrorCode.CreateOperationMissingOrUnknownProperty);
     }
 
-    const encodedSuffixData = operationObject.suffix_data;
```
isaacJChen, line 189: Legacy long form has to remain snake case.

lib/core/versions/latest/models/DocumentModel.ts
```diff
@@ -6,6 +6,6 @@ import ServiceEndpointModel from './ServiceEndpointModel';
  * NOTE: This model should ONLY be used by the `DocumentComposer`.
  */
 export default interface DocumentModel {
-  public_keys?: PublicKeyModel[];
-  service_endpoints?: ServiceEndpointModel[];
+  publicKey?: PublicKeyModel[];
```
isaacJChen, line 9: This is an internal model so they don't have to have the same key names as DID spec, and we want array names to be plurals.

lib/core/versions/latest/VerificationRelationship.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * Sidetree public key verificationRelationship.
+ */
+enum VerificationRelationship {
```
thehenrytsai, line 4: Please revert and stash this part of the change from the PR because:

1. They are unrelated to the property casing change.
1. This has further impact to downstream clients/services which they are not aware of and did not agree to
1. The current design of separation of patch actions and DID document output is intentional.

Adding @csuwildcat and @troyronda since this would require specification changes on both the core Sidetree protocol and the REST API, and they are editors in these spec. I am asking this part to be reverted so the casing change can be merged without being stalled by discussions on this.

tests/generators/OperationGenerator.ts
thehenrytsai: eslint would've picked this up, but too bad since the switch from tslint to eslint a few days ago, we still have ~200 lint issues to resolve.

.vscode/settings.json
```diff
@@ -9,5 +9,6 @@
     "**/versions/[0-9]**/**": true
   },
   "prettier": false,
+  "editor.formatOnSave": false,
```
thehenrytsai, line 12: Looks like this is already the default setting.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -300,7 +300,7 @@ export default class BitcoinProcessor {
     hashOfEarliestKnownValidBlock: string,
     startingBlockHeight: number) {
 
-    let validBlockCount = 0; // Just for console print out purpose at the end.
+    let validBlockCount = 0; // Just for console print out verificationRelationship at the end.
```
thehenrytsai, line 303: Find and replace error.

lib/core/versions/latest/AnchorFile.ts
```diff
@@ -167,9 +167,9 @@ export default class AnchorFile {
 
     const createOperations = createOperationArray.map(operation => {
       return {
-        suffix_data: { 
-          delta_hash: operation.suffixData.deltaHash, 
-          recovery_commitment: operation.suffixData.recoveryCommitment, 
+        suffixData: { 
+          deltaHash: operation.suffixData.deltaHash, 
+          recoveryCommitment: operation.suffixData.recoveryCommitment, 
           type: operation.suffixData.type
```
thehenrytsai, line 173: Now that we don't need this transform logic, let's just assign the property directly and remove the bloat?
Do the same thing for all occurrences to remove all the bloat.
```suggestion
        suffixData: operation.suffixData
```

lib/core/versions/latest/AnchorFile.ts
```diff
@@ -167,9 +167,9 @@ export default class AnchorFile {
```
thehenrytsai, line 167: This PR missed at least a couple of snake_casing properties, `did_suffix` and `signed_data` for instance, please make sure all snake_casing properties are switched over.

thehenrytsai: `map_file_uri`

thehenrytsai: `writer_lock_id`

lib/core/versions/latest/CreateOperation.ts
```diff
@@ -126,14 +126,14 @@ export default class CreateOperation implements OperationModel {
...
-      recoveryCommitment: operationObject.suffix_data.recovery_commitment
+      deltaHash: operationObject.suffixData.deltaHash,
+      recoveryCommitment: operationObject.suffixData.recoveryCommitment
     };
 
-    if (operationObject.suffix_data.type !== undefined) {
-      suffixData.type = operationObject.suffix_data.type;
+    if (operationObject.suffixData.type !== undefined) {
+      suffixData.type = operationObject.suffixData.type;
     }
```
thehenrytsai, line 137: Similar comment from above: I think the bloat can be dealt with:
```
const suffixData = operationObject.suffixData;
```

lib/core/versions/latest/CreateOperation.ts
```diff
@@ -150,7 +150,7 @@ export default class CreateOperation implements OperationModel {
         Operation.validateDelta(operationObject.delta);
         delta = {
           patches: operationObject.delta.patches,
-          updateCommitment: operationObject.delta.update_commitment
+          updateCommitment: operationObject.delta.updateCommitment
         };
```
thehenrytsai, line 154: Remove bloat:
delta = operationObject.delta;

lib/core/versions/latest/models/DocumentModel.ts
```diff
@@ -6,6 +6,6 @@ import ServiceEndpointModel from './ServiceEndpointModel';
...
+  service?: ServiceEndpointModel[];
```
thehenrytsai, line 10: Similar comment to `VerificationRelationship`, this "misalignment" if you will, is intentional. Further discussion needed if you disagree with it, but this is not strictly related to the snake_casing change. So I'd suggest to change them to `publicKeys` and `serviceEndpoints` respectively to get the casing portion merged without further delay.

lib/core/versions/latest/DocumentComposer.ts
```diff
@@ -2,7 +2,7 @@ import DocumentModel from './models/DocumentModel';
 import DidState from '../../models/DidState';
```
thehenrytsai, line 2: I am skipping this entire file for now as it is pending a bunch of code revert requested.

lib/core/versions/latest/models/PublicKeyModel.ts
```diff
@@ -1,11 +1,11 @@
-import PublicKeyPurpose from '../PublicKeyPurpose';
+import VerificationRelationship from '../VerificationRelationship';
 
 /**
  * Data model representing a public key in the 'publicKey' array in patches.
  */
 export default interface PublicKeyModel {
   id: string;
   type: string;
-  jwk: any;
```
thehenrytsai, line 9: Pushing back on this change as this is what is currently in spec and unrelate to casing change.

lib/core/versions/latest/ErrorCode.ts
```diff
@@ -31,12 +31,12 @@ export default {
...
-  CreateOperationSuffixDataMissingOrUnknownProperty: 'create_operation_suffix_data_missing_or_unknown_property',
-  CreateOperationSuffixDataTypeInvalidCharacter: 'create_operation_suffix_data_type_invalid_character',
-  CreateOperationSuffixDataTypeIsNotString: 'create_operation_suffix_data_type_is_not_string',
-  CreateOperationSuffixDataTypeLengtGreaterThanFour: 'create_operation_suffix_data_type_length_greater_than_four',
+  CreateOperationSuffixDataIsNotObject: 'create_operation_suffixData_is_not_object',
+  CreateOperationSuffixDataMissingOrNotString: 'create_operation_suffixData_missing_or_not_string',
+  CreateOperationSuffixDataMissingOrUnknownProperty: 'create_operation_suffixData_missing_or_unknown_property',
+  CreateOperationSuffixDataTypeInvalidCharacter: 'create_operation_suffixData_type_invalid_character',
+  CreateOperationSuffixDataTypeIsNotString: 'create_operation_suffixData_type_is_not_string',
+  CreateOperationSuffixDataTypeLengtGreaterThanFour: 'create_operation_suffixData_type_length_greater_than_four',
```
thehenrytsai, line 39: Find and replace error.

lib/core/versions/latest/ErrorCode.ts
```diff
@@ -122,8 +122,8 @@ export default {
   MultihashUnsupportedHashAlgorithm: 'multihash_unsupported_hash_algorithm',
   OperationCreateInvalidDidDocument: 'operation_create_invalid_did_document',
   OperationCountLessThanZero: 'operation_count_less_than_zero',
-  OperationCreatePayloadHasMissingOrInvalidNextRecoveryCommitment: 'operation_create_payload_has_missing_or_invalid_next_recovery_commitment',
-  OperationCreatePayloadHasMissingOrInvalidNextUpdateCommitment:  'operation_create_payload_has_missing_or_invalid_next_update_commitment',
+  OperationCreatePayloadHasMissingOrInvalidNextRecoveryCommitment: 'operation_create_payload_has_missing_or_invalid_next_recoveryCommitment',
+  OperationCreatePayloadHasMissingOrInvalidNextUpdateCommitment:  'operation_create_payload_has_missing_or_invalid_next_updateCommitment',
```
thehenrytsai, line 126: Find and replace error. I realized this is painful, I am convinced that property and value being the same casing now trumps the "friendliness" to the reader of these messages, just haven't gotten around to write a script to change them all over.

package.json
```diff
@@ -58,7 +58,8 @@
   "scripts": {
     "commit": "git-cz",
     "build": "tsc && copyfiles \"lib/**/*.json\" dist && copyfiles \"tests/**/*.js*\" dist && copyfiles \"package.json\" dist",
-    "test": "npm run build && nyc jasmine --config=./tests/jasmine.json",
+    "test": "npm run build && jasmine --config=./tests/jasmine.json",
+    "test:coverage": "npm run build && nyc jasmine --config=./tests/jasmine.json",
```
thehenrytsai, line 62: Pushing back on this change. Why postpone test coverage report when you can get it in the same run with practically no impact on execution time? (I had the same way as yours, and was convinced in my own PR by another reviewer that having code coverage in the same report is the way to go)

tests/bitcoin/BitcoinClient.spec.ts
```diff
@@ -1037,7 +1037,7 @@ describe('BitcoinClient', async () => {
         labels: [
           {
             name: 'sidetree',
-            purpose: 'receive'
+            verificationRelationship: 'receive'
```
thehenrytsai, line 1040: Find and replace error.

thehenrytsai (changes requested):
A number changes needed.

package.json
OR13: when focusing tests, why look at a coverage report that is 99% irrelevant?

tests/bitcoin/BitcoinClient.spec.ts
OR13: https://github.com/decentralized-identity/sidetree/pull/881/commits/cba9d69241087b4f4b2c4803aaa9b8d3431538ca

lib/core/versions/latest/ErrorCode.ts
OR13: https://github.com/decentralized-identity/sidetree/pull/881/commits/5f3194cd745beabfdd33e4ed3717cd5d482174e7

lib/core/versions/latest/models/PublicKeyModel.ts
OR13: https://w3c.github.io/did-core/#key-types-and-formats

https://tools.ietf.org/html/rfc7517

JWK could be a private or symmetric key, luckily we use the term `privateKey` to refer to JWKs that have a `d` values in sidetree code base today....would still prefer the term `privateKeyJwk` instead... 

might as well match did core, also makes JSON production easier, 1 less internal property that needs to be mapped to did core external representation.

lib/core/versions/latest/CreateOperation.ts
OR13: legacy clients should pin their version to avoid taking this change... semver allows them to avoid taking a breaking change until they are ready to.

lib/core/versions/latest/DocumentComposer.ts
OR13: Transmute, SecureKey and everyone using Linked Data Proofs and Verifiable Presentations rely on verification relationships beyond `authentication`... see:

https://w3c.github.io/did-core/#verification-relationships

We have support them in our fork here:

https://github.com/transmute-industries/sidetree.js/blob/master/packages/common/src/enums/PublicKeyPurpose.ts#L23

You can see the go implementation supports them here:

https://github.com/trustbloc/sidetree-core-go/blob/28fabda71792f0bf59735ddbc9fa38bebcca1a32/pkg/dochandler/transformer/didtransformer/transformer.go#L130

Yet the sidetree reference implementation and spec do not align with did core on this front.

The Document compose change set most applies to public_keys vs `publicKey` and `service_endpoints` vs `service`... I did not implement support for new verification relationships in this code base, but I did correct the enum and cover the cases documented in did core today, because.... rather fix them all at once while making other breaking changes.

lib/core/versions/latest/models/DocumentModel.ts
OR13: -1 to making this reference implementation internal data model use an `s` and the external data model not. needlessly different, will result in more implementation bugs.

lib/core/versions/latest/AnchorFile.ts
OR13: `chunk_file_uri`

.vscode/settings.json
OR13: it depends on user preferences, in our mono repo projects I have my set to auto format and lint everything...

```
{
  "editor.formatOnSave": true,
  "editor.codeActionsOnSave": {
    "source.fixAll.eslint": true
  },
  "eslint.validate": ["javascript", "typescript"],
  "eslint.workingDirectories": [
      {
          "pattern": "packages/*"
      }
  ]
}

```

This reduces the amount of `chore: lint fix` commits. I can remove the explicit disable if you like.

lib/core/versions/latest/VerificationRelationship.ts
OR13: Downstream clients should pin to the latest stable version, so that we can implement many breaking changes all at once, and let them catch up in chunks.

Everyone who uses Verifiable Credentials and Encrypted Data Vaults needs these verification relationships...

https://github.com/transmute-industries/sidetree.js/blob/master/packages/common/src/enums/PublicKeyPurpose.ts#L23

https://github.com/trustbloc/sidetree-core-go/blob/28fabda71792f0bf59735ddbc9fa38bebcca1a32/pkg/dochandler/transformer/didtransformer/transformer.go#L130

https://w3c.github.io/did-core/#verification-relationships

Rather than push another set of breaking changes, I would prefer to simple correct the string enum "auth" to "authentication" and "general" to "verificationMethod" to align with did core.

I also added string enum values which match the exact bucket names in DID Core for the other missing relationships.

I'm afraid if we don't add support for these (and fix the internal naming confusion) in this PR, subsequent breaking changes will no longer be possible.

Better to get it done now.

lib/core/versions/latest/models/DocumentModel.ts
OR13: since these keys expressed in the external (client facing) API, I believe the choice to add and remove `s` will lead to significant developer pain... and since did core uses `publicKey` I think the external interface should match the external representation. If these data models were not externally facing I would happily add an `s` to resolve this.

lib/core/versions/latest/CreateOperation.ts
OR13: any clients that are using the legacy version, should pin to the last stable version until they are ready to take breaking changes.

semver allows client to take breaking changes when they would like to, and avoid taking them until they are ready to "catch up".

tests/generators/OperationGenerator.ts
OR13: I would prefer not to resolve lint fixes in this PR, it's already big enough.

OR13:
I have concerns about doing multiple rounds of breaking changes and the need to align with DID Core and the other sidetree implementations around verification relationships. I have addressed most of the PR feedback, but we should take some time on the next call to discuss SemVer and its use to protect clients from breaking changes, as well as any remaining casing issues.

tests/generators/OperationGenerator.ts
thehenrytsai: Agree, as long as we don't increase the total eslint erro count.

lib/core/versions/latest/CreateOperation.ts
thehenrytsai: That's one valid view of how a node operator can support (or not support) breaking long-form DID changes before v1, this approach would require 2 versions of Sidetree nodes to be deployed side-by-side. But the reference implementation took a different approach, ie. single version deployment that supports such breaking changes, not supporting this now would mean a lot more work from Microsoft's DID infra team and some work on clients/services teams to understand 2 versions and 2 endpoints, which would almost certainly mean push back on the casing change. The idea always was to minimize the downstream dependent client/service changes.

lib/core/versions/latest/models/DocumentModel.ts
thehenrytsai: I disagree with @OR13 here. But since this is contentious and technically unrelated to the casing change, please stash this part of the change for a separate discussion.

lib/core/versions/latest/VerificationRelationship.ts
thehenrytsai: I disagree. But since this is contentious and technically unrelated to the casing change, please stash this part of the change for a separate discussion.

.vscode/settings.json
thehenrytsai: Don't feel strongly on this one. Call is yours.

lib/core/versions/latest/models/DocumentModel.ts
thehenrytsai: I disagree. But since this is contentious and technically unrelated to the casing change, please leave this part of the change out for a separate discussion so we can at least push the agreed upon change as at least a small win.

lib/core/versions/latest/DocumentComposer.ts
thehenrytsai: I disagree. But since this is contentious and technically unrelated to the casing change, please leave this part of the change out for a separate discussion so we can at least push the agreed upon change as at least a small win.

lib/core/versions/latest/models/PublicKeyModel.ts
thehenrytsai: I buy the argument about being explicit and differentiate from "private key" for this one in principle, but have to push back because this is not part of the agreed upon change and taking this would mean more changes on clients/services and also backwards compat code changes in ref-imp.

package.json
thehenrytsai: Just don't look at it? This is more for the case that people forget to run coverage and as a result coverage decreases over time. I am not gonna block this PR over this little thing, so I'll let you make the call, but my 2 cents here is that unless you are going to touch this codebase and run these tests frequently from now on, I think it makes sense to let the current practice be the default. Maybe a good compromise would be to introduce `test:no-cc` instead.

OR13: you can set the PR to fail if coverage drops below the required threshold, and then apply that to every PR, so that nobody can choose to ignore coverage :)

But I will also happy to revert this change, to `test:only`, so thats its easy quickly run tests without seeing coverage that is going to be 0 since the tests are focused.

lib/core/versions/latest/ErrorCode.ts
```diff
@@ -122,7 +122,7 @@ export default {
   MultihashUnsupportedHashAlgorithm: 'multihash_unsupported_hash_algorithm',
   OperationCreateInvalidDidDocument: 'operation_create_invalid_did_document',
   OperationCountLessThanZero: 'operation_count_less_than_zero',
-  OperationCreatePayloadHasMissingOrInvalidNextRecoveryCommitment: 'operation_create_payload_has_missing_or_invalid_next_recovery_commitment',
+  OperationCreatePayloadHasMissingOrInvalidNextRecoveryCommitment: 'operation_create_payload_has_missing_or_invalid_next_recoveryCommitment',
```
tplooker, line 125: Find and replace error

OR13: thanks

OR13:
@thehenrytsai I am going to resolve comments that have been addressed to focus debate on any remaining issues, feel free to reopen any you feel are not resolved.

lib/core/versions/latest/CreateOperation.ts
OR13: I don't understand your comment, how does pinning the version not solve this problem? 

ION is already pinned:

https://github.com/decentralized-identity/ion/blob/master/package.json#L9

The latest unstable version is https://www.npmjs.com/package/@decentralized-identity/sidetree/v/0.11.0-unstable.8c59559

This change set will break the ability for clients to resolve long form DIDs, of the (now 2) legacy formats.... You should simply avoid taking these changes until you are willing to support them.

tests/generators/OperationGenerator.ts
OR13: You can cause PRs to fail if lint rules fail using CI, and you can also setup husky to prevent commits where lint errors occur. I am going to resolve this, as that work should happen in a subsequent PR.

lib/core/versions/latest/models/PublicKeyModel.ts
OR13: I disagree, we agreed to implement breaking changes for clients, all properties under `public_keys` are now possible to align with did core, and provide a better development experience for everyone who relies on the sidetree protocol, Transmute, SecureKey, Microsoft, ConsenSys, and others... The working group should determine the design aesthetics of the sidetree protocol, not a specific did method consuming client team.

lib/core/versions/latest/DocumentComposer.ts
OR13: Seems like the consensus on the call was to take these changes.

lib/core/versions/latest/models/DocumentModel.ts
OR13: Seems like the consensus on the call was to take these changes.

lib/core/versions/latest/VerificationRelationship.ts
OR13: Seems like the consensus on the call was to take these changes, and align with the W3C DID Core Spec.

thehenrytsai:
Thanks @OR13 for pulling this together! My overarching feedback is that the REST API property naming convention need not follow the DID Core spec convention (that's the reason and intention behind `DocumentComposer`), I would even go as far as saying that for Sidetree REST API properties, they should NOT adopt the singular `publicKey` and `service` naming convention for arrays. The guiding principle in my opinion should be KISS for the clients (Keep It Simple, Stupid), for instance, keeping pluralities in property names for ALL array properties for consistency across the API.

There are three options to take this PR across the finish line:

1. Include only snake_case -> camelCase changes in this PR.
2. Snake case changes + opportunistic name fixes WIHTOUT strictly following DID Core naming convention.
3. Snake case changes + opportunistic name fixes strictly following DID Core naming convention.

My vote is for 1 or 2, thus asking the PR to be updated accordingly.

csuwildcat:
I support Henry's preferred option #2 above, making any other changes beyond casing a logical, orthogonal decision based on specific, objective API outcomes, not just blanketly changing strings to match an intermediate pre-CR draft of props/values that are more likely than not to shift further.

OR13:
I am in favor of:

> 3. Snake case changes + opportunistic name fixes strictly following DID Core naming convention.

Reason being that all changes in this PR are underneath the breaking changes regarding casing, and I have always objected to the "data model" that Microsoft has bundled into the sidetree spec, I am more familiar with the W3C standards, and we are required to implement support for W3C standards, and therefore we prefer the external API to look more like the W3C standards:

- https://w3c.github.io/did-core/#verification-relationships
- https://w3c.github.io/vc-data-model/#proofs-signatures

I believe that alignment with the W3C Standards is in the best interests of the working group, and I don't believe we will get another change to do alignment cleanly because if we change any of the string values for:

VerificationRelationship.Authentication = 'auth', 
VerificationRelationship.Authentication = 'authentication', 
VerificationRelationship.Authentication = 'general', 
VerificationRelationship.Authentication = 'verificationMethod'

If we don't fix this now, there will either be breaking changes for the Microsoft client dev team later... or we will never align with DID Core, in our external API.

I am concerned that we are aligning the reference implementation (and therefore the spec) to an API which is actually more confusing for all sidetree implementors and we have an opportunity to fix it in 1 set of breaking changes.

thehenrytsai:
> VerificationRelationship.Authentication = 'auth',
> VerificationRelationship.Authentication = 'authentication',
> VerificationRelationship.General = 'general',
> VerificationRelationship.VerificationMethod = 'verificationMethod'

To be clear, while I am fairly opinionated on the naming of `VerificationRelationship` name , option 2 can certainly include the above enum value changes.

I still do think REST API should NOT adopt `publicKey` and `service` as array as I argue they will just end up looking consistent to the clients and thus prone to trivial bugs as they now have to juggle context knowledge in both specs, thus those changes would NOT be incompatible with option 2.

tests/core/Resolver.spec.ts
```diff
@@ -287,13 +303,13 @@ describe('Resolver', () => {
...
-      expect(document.service_endpoints).toBeDefined();
-      expect(document.service_endpoints.length).toEqual(1);
-      expect(document.service_endpoints[0].endpoint).toBeDefined();
-      expect(document.service_endpoints[0].id).toEqual('newDummyHubUri2');
+      expect(document.publicKey.length).toEqual(2);
+      expect(actualNewSigningPublicKey1!.publicKeyJwk).toEqual(newSigningPublicKey.publicKeyJwk);
+      expect(actualNewSigningPublicKey2!.publicKeyJwk).toEqual(newKey2ForUpdate1AfterRecovery.publicKeyJwk);
+      expect(document.service).toBeDefined();
+      expect(document.service.length).toEqual(1);
+      expect(document.service[0].endpoint).toBeDefined();
```
troyronda, line 311: note: this is additional difference as compared to did-core. https://www.w3.org/TR/did-core/#service-endpoints
`service.endpoint` vs `service.serviceEndpoint`

troyronda:
Thanks for the efforts in cleaning up property strings and increasing alignment with did-core:
- Usage of style guides for JSON properties and the REST API is helpful.
- Adding the remaining verification relationship enumerated types is good and helps alignment with other Sidetree implementations.
- Adjusting the casing to match the style guide is good.
- In general, aligning the Sidetree data model (and terminology) with the target data model (and terminology) helps reduce code complexity and the mental model of understanding Sidetree.

Some potentially competing interests of did-core alignment are:
  - Stability of the did-core spec.
  - The derivation of the did-core spec from JSON-LD leads to singular collection names, while JSON style guidance calls for plural.
  - Desire to simplify the "API" in Sidetree. For example, this PR doesn't address the data model delta of separating the verification method collection from the verification relationships as is done in did-core. Note: part of the tension is derived from having the data model exposed as operations in the REST API, and thus needing to treat the data model as "API".

To me, this PR could be helped by:
- Having a consensus on the singular vs plural collection issue _in the data model_ - the JSON style guide recommends plural but the target data model (did-core) has singular. Regardless of this outcome, I think the REST API endpoints and properties _unrelated to the data model_ should follow the style guide.
- Having a consensus (and justifying) which aspects of the Sidetree data model we are explicitly not aligning with the did-core data model.
- Having a consensus about dealing with future differences if did-core spec evolves.

csuwildcat:
In response to @troyronda's prompt: I vote to keep the Sidetree Patch Actions and REST API free from commingling multiple, conflicting layers of naming conventions and style guide directives, and just focus on the casing + changing things like `auth` to `authentication` where it makes sense for _creating a cohesive API_.

OR13:
> * Note: part of the tension is derived from having the data model exposed as operations in the REST API, and thus needing to treat the data model as "API".

Yep, and IMO this is not a concern for the spec, and can easily be addressed via separate APIs... the external patch api is what is of primary relevance to this working group consensus.

Regarding the consensus on styleguide for external sidetree APIs, this is why styleguides are so important to start with. I think we should ammend the styleguid which we have added in this PR to include the following:

Sidetree is protocol for building scalable decentralized identifier methods that are capable of support verifiable credentials as well as legacy identity and credentialing standards.

In order to provide the best possible developer experience, the external sidetree API is first aligned with the W3C standards to avoid potentially confusing casing or pluralization issues which might lead to significant bugs.

For external sidetree APIs we follow the property names associated the external representation, for example: 

`keyAgreement` is a verification relationship defined in did core, so we use the term without an `s`.

For any data model that is not in the external sidetree API, developer preference is allowed, and we try our best to follow the conventions documented in .... `<the styleguide in this PR>`

I am so strongly opposed to the idea of having an external API that differs from did core by changing pluralization (publicKey vs publicKeys) that I will not do the work to correct this PR to implement that, and I will formally object if that path is taken, as I believe it is harmful bot to the W3C Standards we are trying to support and the developers who will be hoping Sidetree is the best way to implement those standards.

> Having a consensus about dealing with future differences if did-core spec evolves.

I suggest we be honest about the nature of building on standards that have not been fully ratified by a real standards organization like the W3C. Until there is a stable permanent version of the spec, changes may occur.

If we are implementing these standards on behalf of customers we will have to choose between "really implementing the standard", or "shipping something to production and then never changing it"... This only applies to our external INPUT API... the patch structure... the DID Document compose is capable of aligning with did core regardless of future changes.... 

Because the current patch structure is needlessly different from did core, I believe it is needlessly confusing, and we have an opportunity to fix it while addressing the casing issue, and while keeping the external API as simple and aligned with did core as possible.

I am happy to agree that we don't make any more changes to the external APIs after this regardless of what did core does.

In conclusion, I am in favor of merging this PR as is, and won't be making additional changes to it.

mprorock:
We likewise are required by customer needs to utilize libraries and APIs that are in line W3C standards, and like @OR13 prefer the external API to look more like the W3C standards and would vote to merge this PR.

I will strongly emphasize and agree with the statements made above that:
> alignment with the W3C Standards is in the best interests of the working group, and I don't believe we will get another chance to do alignment cleanly because if we change any of the string values for:
> 
> VerificationRelationship.Authentication = 'auth',
> VerificationRelationship.Authentication = 'authentication',
> VerificationRelationship.General = 'general',
> VerificationRelationship.VerificationMethod = 'verificationMethod'
> 
> If we don't fix this now, there will either be breaking changes for the Microsoft client dev team later... or we will never align with DID Core, in our external API.

If this PR is not merged, it would be a pretty strong blocker for us to use sidetree as it exists today with a number of high value, high volume customers.

Therecanbeonlyone1969:
I am in the same boat as @mprorock. In favor of merging the PR.

bumblefudge:
> To me, this PR could be helped by:
> 
> * Having a consensus on the singular vs plural collection issue _in the data model_ - the JSON style guide recommends plural but the target data model (did-core) has singular. Regardless of this outcome, I think the REST API endpoints and properties _unrelated to the data model_ should follow the style guide.
> * Having a consensus (and justifying) which aspects of the Sidetree data model we are explicitly not aligning with the did-core data model.
> * Having a consensus about dealing with future differences if did-core spec evolves.

Knowing the working group as little as I do, I don't want to meddle too much but I will underline that these seem to me very reasonable and helpful requests for the group process-wise. The relative prioritization of data model, style guide, and did-core alignment seems to me an underlying conflict here, best dealt with in the abstract, and ideally with minimal reference to everyone's client demands and reputation issues around breaking changes.

lib/core/versions/latest/CreateOperation.ts
kdenhartog: @OR13 I think the concern is that this means they have to keep separate nodes running for multiple teams until they're all up to date.

@thehenrytsai Is this legacy code that's just going to be deprecated in the future and it's fine to make this change where it's completely removed in a later PR once the client teams can update to not using this longform anymore?

lib/core/versions/latest/DocumentComposer.ts
kdenhartog: It's my preference that we use the terms defined in the did core specification, but also adhere to the plurality rule that's mentioned in all of those style guides. 

This would leave us with 
publicKeys, 
verificationRelationships
services

The reason I'm preferring this is because the names of the properties will be easy to map as a developer that they're the similar to what's in did-core not leaving it up to interpretation and getting very close to the [principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment), but we place the style guide above all as the preferred way to implement the code without exception. This will go a long way to resolving conflicts for us because then we can always have the style guide as an independent arbitrator of truth. 

On a side note, the problem is that the did-core specification isn't using plural nouns for arrays, but given that ship has sailed, let's at least make sure we get it right here.

lib/core/versions/latest/VerificationRelationship.ts
kdenhartog: It's my preference that we get all the breaking changes in to adhere to the style guide now and then we don't need to change in the future. In this case, that would mean I'd prefer to keep `verificationRelationship`, but pluralize it to be `verificationRelationships`. See above comment for rationale.

lib/core/versions/latest/models/DocumentModel.ts
kdenhartog: Preference is towards plural here to adhere to the style guide.

lib/core/versions/latest/models/PublicKeyModel.ts
kdenhartog: I'd prefer to see all changes like this go in now and have them set in stone. In this case publicKeyJwk is preferred for consistency to the did-core spec so that it adheres to the principle of least astonishment.

tests/core/Resolver.spec.ts
kdenhartog: prefer `services.serviceEndpoint` here

tests/fixtures/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+Then you need to update the other fixtures, by toggling `OVERWRITE_FIXTURES`.
+
+For example see [Resolver.spec.ts](../core/Resolver.spec.ts).
+
+When running tests with `OVERWRITE_FIXTURES` set to true, the `received` output will be written to disk.
+
+The next time you run the test, it should pass, because `received` will match `expected`.
+
+Be sure to review the JSON that is being written to disk, its possible for test to pass and the JSON on disk to be horribly wrong.
```
kdenhartog, line 17:
```suggestion
Be sure to review the JSON that is being written to disk, its possible for test to pass and the JSON on disk to be incorrect.
```

tests/generators/OperationGenerator.ts
kdenhartog: let's add the space here because it's a change that would create a new error, and we'll go back and fix the subsequent errors in a later PR (as well as add the CI changes to make linting fail the pipeline)

tests/utils.ts
```diff
@@ -0,0 +1,21 @@
...
+};
+
+// this funtion should not be necessary if fixtures are well designed
+// however, it is useful while they remain randomly generated.
+export const fixtureDriftHelper = (recieved: any, expected: any, pathToFixture:string, overwrite:boolean = false) => {
+  const match = JSON.stringify(recieved) === JSON.stringify(expected);
+  if (!match) {
+    // console.error('Fixture Drift!');
+    // console.warn('Consider updating JSON to: ');
+    // console.warn(JSON.stringify(recieved, null, 2));
```
kdenhartog, line 16: remove comments here?

csuwildcat:
If there is a change to use more DID Core strings in certain places, I really agree and appreciated what @kdenhartog said about doing so in adherence to the style guide, pluralizing as it directs. I feel this would be a good direction, and if we went this way, we could get on board with it.

troyronda:
This PR has received substantial discussion both in the Sidetree WG and in the above PR comments. In order to make progress, we need to record a decision on this topic.

Please provide your vote by EOD Friday (Oct 16, 2020) on this PR by using one GitHub reaction on this comment.

Thumbs Up: You are of the opinion that we should merge this PR as-is. Implication: an exception to the style guide collection pluralization requirements will be created for `did-core` strings.
Rocket: You are of the opinion that we should modify this PR to follow the style guide without exception for `did-core` strings. The `did-core` collection strings will be pluralized in this PR but otherwise merged as-is.
Eyes: You are of the opinion that we should modify this PR to follow the style guide without exception for `did-core` strings AND drop the `did-core` focused string alignment aspects of this PR. We will make decisions for case-by-case string improvements during the next WG meeting (Oct 20, 2020).

The above vote is to record a final decision for the stable 1.0 strings for the Sidetree spec.

kdenhartog:
I've got a slight preference towards 🚀  still, but am more than willing to compromise for 👀 . I'm also willing to fix it up for any of these two options.

csuwildcat:
Please, all of you, read this thread and what Dave Longley (@dlongley), co-author of the DID spec, wrote in response, wherein he specifically advocates that implementers SHOULD NOT try to bind themselves to terms (especially non-normative terms) found within the spec: https://github.com/w3c/did-core/issues/318#issuecomment-709727821

I *strongly* urge you to vote EYES emoji so we can make rational decisions about properties/values instead of defying the good advice of the authors of the very spec we're working with.

kdenhartog:
I've updated to 👀 after speaking with @csuwildcat more and reading the responses on the thread. It's only slightly edged 🚀 out for me because verificationRelationships is meant as a concept and trying to bind to it is not what was intended through the terminology chosen. PublicKeyPurpose also remains accurate. Even so, 👀 allows us to decide the names of the properties on the call so we can try to find something that everyone will live it.

lib/core/versions/latest/CreateOperation.ts
OR13: my concern is that a "reference implementation" should not be used to support "legacy code paths".... it is meant to be copied by other implementations... I suggest pinning versions, and removing any "legacy code" from the sidetree reference implementation.

docs/styleguide.md
```diff
@@ -0,0 +1,12 @@
+# styleguide
+
+The sidetree specification relies on JSON and the current node implementations rely on HTTP.
+
+Unless otherwise noted, the following styleguides should be used.
+
+- [Microsoft API Guidelines](https://github.com/microsoft/api-guidelines/blob/vNext/Guidelines.md)
```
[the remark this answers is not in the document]
OR13: Those guides are not the same, I just added the 2 most popular ones I know.

OR13:
> PublicKeyPurpose also remains accurate

are "ethereumAddress" or "bitcoinAddress" public keys? NO.... there is a reason we changed did core to remove `publicKey` in favor of `verificationMethod`....

@kdenhartog `publicKey` is not even defined in did core properties anymore.... https://w3c.github.io/did-core/#core-properties

OR13:
> Please, all of you, read this thread and what Dave Longley (@dlongley), co-author of the DID spec, wrote in response, wherein he specifically advocates that implementers SHOULD NOT try to bind themselves to terms (especially non-normative terms) found within the spec: [w3c/did-core#318 (comment)](https://github.com/w3c/did-core/issues/318#issuecomment-709727821)
> 
> I _strongly_ urge you to vote EYES emoji so we can make rational decisions about properties/values instead of defying the good advice of the authors of the very spec we're working with.

@csuwildcat the enum name does not matter... it could be DIDCoreVM... the part that matters (and that is normative) is the string values associated with the enum... they are normative, and they are defined here:

https://w3c.github.io/did-core/#core-properties

mprorock:
voting Rocket on this one for sure.

csuwildcat:
@OR13 couple things:

1. Everyone is still on board with changing the enum strings to match the DID spec - you seem to be mixing different issues together.
2. The patch action in question only deals with public keys, so that's why it makes sense to name the props in a reasonable way that reflects the specificity of the patch items.

I think you need to re-review what we're saying here, because, as noted in 1 above, I think you're not parsing the different facets of common ground we're calling for.

OR13:
> @OR13 couple things:
> 
> 1. Everyone is still on board with changing the enum strings to match the DID spec - you seem to be mixing different issues together.
> 2. The patch action in question only deals with public keys, so that's why it makes sense to name the props in a reasonable way that reflects the specificity of the patch items.
> 
> I think you need to re-review what we're saying here, because, as noted in 1 above, I think you're not parsing the different facets of common ground we're calling for.

Possibly, maybe the proprosal isn't clear... I am interpretting:

> AND drop the did-core focused string alignment aspects of this PR. 

as don't care about did core strings at all... perhaps it really means `verificationRelationship` -> `purpose` (revert) ?

OR13:
Ok, seems like I did not interpret "eyes" correctly....

I understand eyes request a few additional reverts to the PR as is.

1. `verificationRelationship -> purpose` (revert in patch data model)
2. `publicKey -> publicKeys` (pluralize always)
3. `service.endpoint -> service.serviceEndpoint` (not implemented did core alignment

I am prepared to vote "eyes" if this is confirmed.... I can't vote eyes without know exactly what changes it will imply.

csuwildcat:
Aside: I wish the poll emoji were a horse and eyes, so at the end we could say either "The Nays prevailed" or "The aEyes have it"

docs/styleguide.md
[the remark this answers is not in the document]
troyronda: @kdenhartog @OR13 @csuwildcat Any further action suggested on this?

Therecanbeonlyone1969:
@csuwildcat @OR13 @troyronda @kdenhartog 

I am not 100% clear what the proposals mean in detail. I rewrote them as to my understanding, trying to add a distinction between normative and non-normative. Does my understanding of the options align with your guys understanding?

Rocket: You are of the opinion that we should modify this PR to follow the style guide without exception for **non-normative did-core** strings. The **non-normative did-core** collection strings will be pluralized in this PR but otherwise **normative did-core strings** will be merged **as spelled out in this PR**.

Eyes: You are of the opinion that we should modify this PR to follow the style guide without exception for ** normative and non-normative did-core strings** AND drop the did-core focused string alignment aspects of this PR. We will make decisions for case-by-case string improvements **including normative did-core strings** during the next WG meeting (Oct 20, 2020).

OR13:
@Therecanbeonlyone1969 that was my initial interpretation as well, it sounds like its not what they intended... imprecision here will lead to further debate.... unfortunately we are already in voting... I suggest @csuwildcat  add additional comments to clarify what he thinks "eyes" means.

csuwildcat:
In my understanding, an EYES emoji vote would mean:

1. Adhere to style guide, including camel casing and pluralization where directed
2. Do not blanketly replace all strings with things from DID Core and break the style guide conventions or reasonable API desires, but DO switch in props/values with DID Core props/values that make sense, as we determine them.
3. I believed the intent was to make the move to DID Core strings that everyone can easily agree to (auth > authetication, endpoint > serviceEndpoint, etc.), and figure out the few stragglers on Tuesday's call.

^ I think this is a very reasonable, logically sound course of action that balances the needs of all those involved.

OR13:
> 2\. Do not blanketly replace all strings with things from DID Core and break the style guide conventions or reasonable API desires, but DO switch in props/values with DID Core props/values that make sense, as we determine them.

this is very imprecise, and essentially "we'll discuss on the call".... happy to discuss on the call... but perhapse we can make progress before then by adding to my list on: https://github.com/decentralized-identity/sidetree/pull/881#issuecomment-710061703

@csuwildcat would would consider helping me get the precise list down on that comment, and removing 2 (which is confusing). ?

csuwildcat:
> @csuwildcat would would consider helping me get the precise list down on that comment, and removing 2 (which is confusing). ?

Sure, I think @troyronda was keeping a tally of the ones that had broad agreement (which looked like most of them), and others that may require further discussion on next week's call.

Therecanbeonlyone1969:
> In my understanding, an EYES emoji vote would mean:
> 
> 1. Adhere to style guide, including camel casing and pluralization where directed
> 2. Do not blanketly replace all strings with things from DID Core and break the style guide conventions or reasonable API desires, but DO switch in props/values with DID Core props/values that make sense, as we determine them.
> 3. I believed the intent was to make the move to DID Core strings that everyone can easily agree to (auth > authetication, endpoint > serviceEndpoint, etc.), and figure out the few stragglers on Tuesday's call.
> 
> ^ I think this is a very reasonable, logically sound course of action that balances the needs of all those involved.

@csuwildcat ... I understand the casing part ... that is not the issue. I want to draw a very clear distinction between normative and non-normative and would appreciate to indulge me on this one. Does eyes then mean the following below?

Eyes: You are of the opinion that we should modify this PR to follow the style guide without exception for non-normative did-core strings AND explicitly keep the PR language for **normative** did-core strings. We will make decisions for case-by-case string improvements during the next WG meeting (Oct 20, 2020) except for the **normative** did-core strings.

csuwildcat:
> Eyes: You are of the opinion that we should modify this PR to follow the style guide without exception for non-normative did-core strings AND explicitly keep the PR language for **normative** did-core strings. We will make decisions for case-by-case string improvements during the next WG meeting (Oct 20, 2020) except for the **normative** did-core strings.

Depends where and in what context the intended normative string swap is taking place - for example: we have some patch types that are more precise/targeted than the more generalized nomenclature of DID Core, so certain properties make less sense in certain places where a more targeted/precise term is desired.

Therecanbeonlyone1969:
> > Eyes: You are of the opinion that we should modify this PR to follow the style guide without exception for non-normative did-core strings AND explicitly keep the PR language for **normative** did-core strings. We will make decisions for case-by-case string improvements during the next WG meeting (Oct 20, 2020) except for the **normative** did-core strings.
> 
> Depends where and in what context the intended normative string swap is taking place - for example: we have some patch types that are more precise/targeted than the more generalized nomenclature of DID Core, so certain properties make less sense in certain places where a more targeted/precise term is desired.

@csuwildcat ... I do not care about non-normative did-core string i.e. the ones with a should or may ... I only care about normative did-core strings the ones with a must ... those imho cannot be touched no matter how bad or imprecise they are ... we either align to the normative part of did-core or we do not ... what is it?

csuwildcat:
Hit the wrong button, sorry.

troyronda:
@OR13 has created a [concrete proposal](https://github.com/decentralized-identity/sidetree/pull/881#issuecomment-710061703) for the case-by-case string changes to this PR:

verificationRelationship -> purposes
publicKey -> publicKeys
service.endpoint -> service.serviceEndpoint
service -> services

It would be helpful to see our current level of consensus on this concrete proposal. As such, could you please use one GitHub reaction on this comment as follows:

Thumbs Up: I **am** of the opinion we should merge this PR after making the above string changes.

Thumbs Down: I **am not** of the opinion we should merge this PR after making the above string changes.

peacekeeper:
I'm not sure I fully understand the difference between the "Rocket" and "Eyes" option, but definitely +1 to aligning APIs and internal names with DID Core property names. Pluralization is fine of course, but otherwise I think the terms in the code should match the spec wherever possible...

As long as the DID Core spec isn't final, the argument that "we can't change things because people already depend on it" does not feel convincing to me. Just this week I pushed several breaking changes to [did-common-java](https://github.com/decentralized-identity/did-common-java) and the Universal Resolver (e.g. `publicKey` -> `verificationMethod`) to match the latest version of DID Core.

tplooker:
I'm voting "eyes" but here is why.

The original intention of this PR was to implement a style guide and bring the reference implementation inline with that style guide, an endeavour that will vastly improve consistency across the implementation and set out easy to follow rules in how developers can continue to extend sidetree consistently in future (hats off to @OR13 for making this happen!). What is problematic and has been revealed by the proposals though, is that the style guide selected, conflicts with the decisions made by DID-core, namely around how to name JSON object members that have a value type of array. Therefore in my mind the rocket proposal is in essence proposing we create an exception list to the style guide for the current terms in DID-core that are permitted to have a JSON value type of array. Adopting an exception list approach like this to the style guide I think could bite us in future, because people could add properties to the sidetree data model (that aren't in did core) that use a plural naming convention (by simply following the styling guide), which would therefore create inconsistencies throughout the data model (some properties being named plurally and some that are not).

Essentially I'm opposed to any approach that creates the opportunities for inconsistencies to emerge within the sidetree data model, and instead would rather have to translate from one data model to the other (sidetree -> did core).

IMO the only way I see sidetree being able to fully line up to the did core data model as it stands today without an exception list approach to the style guide, would be to write our own style guide that has universal rules around JSON naming conventions that lineup with did-core.

kdenhartog:
> @OR13 has created a [concrete proposal](https://github.com/decentralized-identity/sidetree/pull/881#issuecomment-710061703) for the case-by-case string changes to this PR:
> 
> verificationRelationship -> purposes
> publicKey -> publicKeys
> service -> services
> service.endpoint -> services.serviceEndpoint
> 
> It would be helpful to see our current level of consensus on this concrete proposal. As such, could you please use one GitHub reaction on this comment as follows:
> 
> Thumbs Up: I **am** of the opinion we should merge this PR after making the above string changes.
> 
> Thumbs Down: I **am not** of the opinion we should merge this PR after making the above string changes.



> Ok, seems like I did not interpret "eyes" correctly....
> 
> I understand "eyes" request a few additional changes to the PR:
> 
>     1. `verificationRelationship -> purposes` (revert in patch data model)
> 
>     2. `publicKey -> publicKeys` (pluralize always)
> 
>     3. `service.endpoint -> service.serviceEndpoint` (not implemented, but desired did core alignment)
> 
>     4. `service` -> `services` (pluralize always)
> 
> 
> I am prepared to vote "eyes" if this is confirmed.... I can't vote eyes without knowing exactly what changes it will imply.

I'm on board with those set forth.

docs/styleguide.md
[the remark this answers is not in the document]
kdenhartog: My preference would be that we pick the Microsoft one since it looked the most comprehensive and having only one would prevent us from running into minor inconsistencies between them.

csuwildcat:
@peacekeeper there's just one issue with your comment about how easy it was for you to "just this week push several breaking changes" to your project: in our case, we're talking about an immutable set of values that we can NEVER stop supporting and can NEVER remove code from nodes/wallets to deal with, because these strings will be baked into DIDs, and subsequently the network, for all time. So while I realize making changes in the project you mentioned was easy, and even agree that it usually is/should be, this is a bit of a different animal, and we must carefully consider the ramifications.

lib/core/versions/latest/CreateOperation.ts
kdenhartog: From what I'm understanding, they want it removed, and they want these new changes to start building off ASAP, but would end up blocking other internal teams and causing lots of unnecessary, regretful work if they have to accept both at once.

It seems reasonable to me to leave them in for a bit longer until they're able to get around to updating them especially since we didn't set an official deprecation date originally from my understanding. This has been what we've done in the Aries/Indy-SDK world which has worked very well for us. As an example, (the precursor authcrypt function)[https://github.com/hyperledger/indy-sdk/blob/f9eb2cf17b51584f875c4707094256a96656e7b8/libindy/src/api/crypto.rs#L288] still exists in Indy-sdk even though everyone is well beyond that work. It's not completely unreasonable to wait a bit longer on this change until things can get updated even with version pinning. Would you be alright if we left it in for now with a comment saying "these are legacy code paths and MUST NOT be implemented by reference implementors going forward"?

peacekeeper:
> in our case, we're talking about an immutable set of values that we can NEVER stop supporting

@csuwildcat Okay yes I understand the difference now, thanks for clarifying..

OR13:
IIW is happening this week, can we get chair / editor approval for the "next steps" on the issue?

kdenhartog:
> IIW is happening this week, can we get chair / editor approval for the "next steps" on the issue?

From the looks of it, we're going with your proposal. https://github.com/decentralized-identity/sidetree/pull/881#issuecomment-710256369

@troyronda shall we declare consensus achieved on this?

lib/core/versions/latest/CreateOperation.ts
thehenrytsai: @kdenhartog, your understanding is accurate, we intend to remove legacy behavior at the first possible opportunity, certainly prior to v1, we just wanted to give whoever is dependent on current behavior (including internal MS teams) the some time to switch over instead of a full breakage. I agree in a perfect world I also would rather not have this stop-gap.

tests/generators/OperationGenerator.ts
```diff
@@ -549,7 +586,7 @@ export default class OperationGenerator {
     if (idOfServiceEndpointToAdd !== undefined) {
       const patch = {
         action: 'add-service-endpoints',
```
troyronda, line 588: It would be sensible for the action string to be `add-services` to be consistent with the rest of the rename.

lib/core/versions/latest/models/PublicKeyModel.ts
thehenrytsai: Consensus to use `publicKeyJwk`.

tests/generators/OperationGenerator.ts
thehenrytsai: Fixed.

troyronda:
> shall we declare consensus achieved on this?

yes

tests/vectors/resolution/afterDeactivate.json
```diff
@@ -0,0 +1,3 @@
+{
+  "status": "deactivated"
```
troyronda, line 2: I wasn't able to find this field in the spec.

I think we only currently note that the HTTP status is 410.

thehenrytsai (approved):

tests/vectors/resolution/longFormResponse.json
```diff
@@ -0,0 +1,43 @@
+{
+  "status": "succeeded",
```
troyronda, line 2: I see a status field in this response (and the deactivated example) but not the other ones.

tests/vectors/resolution/afterDeactivate.json
thehenrytsai: @troyronda, please file a separate bug if you think this implementation must not add anything in boy in deactivate case. But this is the existing behavior so should not block the PR IMO.

tests/vectors/resolution/longFormResponse.json
thehenrytsai: @troyronda, this is actually expected, because the is actually the output the `RequestHandler.handleResolveRequest`, others are the output of `DocumentComposer.transformToExternalDocument`

tests/vectors/resolution/afterDeactivate.json
troyronda: @thehenrytsai  I don't want to block the PR due to existing behavior.

lib/core/versions/latest/PublicKeyPurpose.ts
```diff
@@ -1,7 +1,7 @@
 /**
- * Sidetree public key verificationRelationship.
+ * DID Document public key purpose.
  */
-enum VerificationRelationship {
+enum PublicKeyPurpose {
   VerificationMethod = 'verificationMethod',
```
csuwildcat, line 5: This should not be commingled as a purpose, as it is not in the DID Core spec. This should be the array all keys are printed in, and then referenced from via IDs.

csuwildcat (changes requested):
Remove `verificationMethod` from purposes, print all keys in `verificationMethod` array by default, and reference those keys from purposes specified by the user.

lib/core/versions/latest/PublicKeyPurpose.ts
OR13:
```
{
  "@context": [
    "https://www.w3.org/ns/did/v1",
    {
      "@base": "did:key:z6MkhC1wRfHJ6SzxtTpFJGR6S5QTS7hze6B6jFZJsYBbbMf1"
    }
  ],
  "id": "did:key:z6MkhC1wRfHJ6SzxtTpFJGR6S5QTS7hze6B6jFZJsYBbbMf1",
  "verificationMethod": [
    {
      "id": "#z6MkhC1wRfHJ6SzxtTpFJGR6S5QTS7hze6B6jFZJsYBbbMf1",
      "type": "JsonWebKey2020",
      "controller": "did:key:z6MkhC1wRfHJ6SzxtTpFJGR6S5QTS7hze6B6jFZJsYBbbMf1",
      "publicKeyJwk": {
        "crv": "Ed25519",
        "x": "KKtmCrqQ4EC1B3v6ok6jGJR156NpN0a_5PgZcplxFNw",
        "kty": "OKP"
      }
    },
    {
      "id": "#z6LSrhYgWXdWJcqxniVvTD5Ve85f9E2YpW9PbvNzRZ5VpgBN",
      "type": "JsonWebKey2020",
      "controller": "did:key:z6MkhC1wRfHJ6SzxtTpFJGR6S5QTS7hze6B6jFZJsYBbbMf1",
      "publicKeyJwk": {
        "kty": "OKP",
        "crv": "X25519",
        "x": "3zk2zPWpgich_-x59_DqBujI3Pkk-LQMAghqfwBCt20"
      }
    }
  ],
  "authentication": [
    "#z6MkhC1wRfHJ6SzxtTpFJGR6S5QTS7hze6B6jFZJsYBbbMf1"
  ],
  "assertionMethod": [
    "#z6MkhC1wRfHJ6SzxtTpFJGR6S5QTS7hze6B6jFZJsYBbbMf1"
  ],
  "capabilityInvocation": [
    "#z6MkhC1wRfHJ6SzxtTpFJGR6S5QTS7hze6B6jFZJsYBbbMf1"
  ],
  "capabilityDelegation": [
    "#z6MkhC1wRfHJ6SzxtTpFJGR6S5QTS7hze6B6jFZJsYBbbMf1"
  ],
  "keyAgreement": [
    "#z6LSrhYgWXdWJcqxniVvTD5Ve85f9E2YpW9PbvNzRZ5VpgBN"
  ]
}
```

OR13: @csuwildcat I think you are confusing `purpose` as a property of the sidetree publickey model, with the concept of verification relationships, which is defined in did core:

https://w3c.github.io/did-core/#verification-relationships

OR13: `verificationMethod` is the "general" purpose... its totally valid to expose a key that is not "general" in fact its a security anti pattern to list all keys as "general"... its better to delegate keys for specific purpose, and never reuse the same key for multiple purposes.

OR13: the did:key example I provided can't handle security that way, but sidetree did methods can.

OR13: also, `verificationMethod` is in the did core spec, its what replaced public key: https://w3c.github.io/did-core/#verification-methods

since public key can't describe verification methods such as bitcoin addresses or hash based proofing mechanisms.

OR13: I agree with @csuwildcat actually, this will be more broken if we don't do what you are suggesting.

troyronda (approved):

csuwildcat (approved):
Approved on the condition that we make the following proposed changes in a follow-up PR:

1. Remove the `verificationMethod` enum from purposes.
2. Project all keys to the `verificationMethod` array.
3. Project key references to any of the DID Document locations a user specifies in the purposes array.

Issue #319, opened 2019-10-25
Implement different operation modes (resolver, observer, etc) for the nodes to support different topology


mudiali:
In order to support different deployment configurations, we should have different modes for the sidetree node. For example:

Mode | Comments
------------ | -------------
resolver | Node is going to handle the resolution requests
observer | Node is going to observe the blockchain data
writer | Node is going to accept the 'write' operation requests (create, update) from the clients
batchwriter | Node is going to batch up the write operations and anchor them to the blockchain

One node can operate in multiple modes and they can be then passed via config as:

````
mode: [resolver, writer]
````

csuwildcat:
I don't understand the writer vs batchwriter distinction. What node would ever accept ops without the intention of writing a batch?

thehenrytsai:
Done.

PR #889, opened 2020-10-15
feat(ref-imp): #319 - Added ability to turn on or off Observer and Batch Writer

thehenrytsai:
1. Added ability to turn on or off Observer and Batch Writer.
1. Fixed numerous ESlint errors.


lib/core/Core.ts
```diff
@@ -32,16 +33,18 @@ export default class Core {
...
     this.operationStore = new MongoDbOperationStore(config.mongoDbConnectionString, config.databaseName);
     this.blockchain = new Blockchain(config.blockchainServiceUri);
     this.downloadManager = new DownloadManager(config.maxConcurrentDownloads, this.cas);
     this.resolver = new Resolver(this.versionManager, this.operationStore);
-    this.batchScheduler = new BatchScheduler(this.versionManager, this.blockchain, config.batchingIntervalInSeconds);
     this.transactionStore = new MongoDbTransactionStore(config.mongoDbConnectionString, config.databaseName);
     this.unresolvableTransactionStore = new MongoDbUnresolvableTransactionStore(config.mongoDbConnectionString, config.databaseName);
+
+    this.batchScheduler = new BatchScheduler(this.versionManager, this.blockchain, config.batchingIntervalInSeconds);
     this.observer = new Observer(
```
isaacJChen, line 48: Should we conditionally create the batch scheduler and observer so we don't have to have a transaction store or blockchain service ready for it if we don't want to in the future?

lib/core/Observer.ts
```diff
@@ -6,10 +7,9 @@ import IUnresolvableTransactionStore from './interfaces/IUnresolvableTransaction
 import IVersionManager from './interfaces/IVersionManager';
 import SharedErrorCode from '../common/SharedErrorCode';
 import SidetreeError from '../common/SidetreeError';
-import timeSpan = require('time-span');
 import ThroughputLimiter from './ThroughputLimiter';
 import TransactionModel from '../common/models/TransactionModel';
-import TransactionUnderProcessingModel, { TransactionProcessingStatus } from './models/TransactionUnderProcessingModel';
+import timeSpan = require('time-span');
```
isaacJChen, line 12: Can we use `import * as a from b` ?

lib/core/Core.ts
thehenrytsai: Good suggestion, I prefer to kick this can down the road because the change right now is super simple and in one place. If we now say we will not create observer if observing is disabled, then it begs the follow up question such as "do we need transaction store?" probably not, do we then remove the instantiation of the stores too? Now `VersionManager` references it, do we then do plumb through the `VersionManager`? Just too much than what is worth right now IMO.

Also with the current approach, we leave the possibility open for the observer/writer to start polling through an API call, might be useful for live site mitigation.

lib/core/Observer.ts
thehenrytsai: Great suggestion. Done.

isaacJChen (approved):

Issue #890, opened 2020-10-16
Add ability to disable the updating off value time lock


thehenrytsai:
In the case of multiple running bitcoin service instances running in a cluster sharing the same DB, there can only be one instance that is changing/updating the lock, so this feature is to avoid the service running into concurrency and/or conflict issues.

thehenrytsai:
Done.

PR #891, opened 2020-10-16
#890 - Ability to turn off value time lock updates

thehenrytsai:
1. #890 - Ability to turn off value time lock updates.
2. Added documentation on value lock config parameters.
3. Lots and lots of ESLint error fixes.


lib/bitcoin/lock/LockMonitor.ts
```diff
@@ -31,17 +32,21 @@ interface LockState {
  * Encapsulates functionality to monitor and create/remove amount locks on bitcoin.
  */
 export default class LockMonitor {
-  private initialized: boolean;
-
   private periodicPollTimeoutId: NodeJS.Timeout | undefined;
 
   private currentLockState: LockState;
```
thehenrytsai, line 37: I think there can be a further simplification here: that we remove this in-memory `currentLockState` completely. Let me know if you agree then I will remove it.

The way I see it:
1. Polling happens only once a few minutes, it's okay if it does a full `getCurrentValueTimeLock()` call.
2. If REST API call occurs for getting the active lock, it is even more okay that we do a full `getCurrentValueTimeLock()` call also because a) the usage pattern should be that this occurs once every 10 minutes or so and b) this ensures that we are getting the latest state of the lock, not some "cached state".

lib/bitcoin/lock/LockMonitor.ts
```diff
@@ -123,45 +117,45 @@ export default class LockMonitor {
   }
 
   private async handlePeriodicPolling (): Promise<void> {
+    this.currentLockState = await this.getCurrentLockState();
```
thehenrytsai, line 120: Calling out that internally `getCurrentLockState()` will "help" broadcast the lock transaction if it is not found, so the `LockMonitor` is not in "pure read mode" even if lock update is disabled. But I don't think this is a big deal. If you have concern for this behavior, let me know!

isaacJChen (dismissed):

lib/bitcoin/lock/LockMonitor.ts
isaacJChen: I think it's fine as long as it doesn't update the lock internally. Broadcasting just lets other bitcoin nodes know it exists, and it shouldn't even happen often if we don't create new locks.

mudiali: Same here. Broadcast is not the same as 'updating'. Someone already updated it, and this node is just broadcasting it which is a noop if already broadcasted.

lib/bitcoin/lock/LockMonitor.ts
```diff
@@ -123,45 +117,45 @@ export default class LockMonitor {
...
 
     if (!lockRequired && validCurrentLockExist) {
-      currentLockUpdated =
-        await this.handleReleaseExistingLock(
-          this.currentLockState.activeValueTimeLock!,
-          this.desiredLockAmountInSatoshis);
-    }
+      console.info(LogColor.lightBlue(`Value time lock no longer needed.`));
 
-    if (currentLockUpdated) {
```
mudiali, line 163: Just wondering .. should we call the `getCurrentLockState` again at the end to make sure that we have the correct new state in case any updates to the lock were made? Any of the changes will not be reflected and the `this.currentLockState` will be in stale state for when the `getCurrentValueTimeLock` is called.

Off course if we remove the `this.currentLockState` completely then this issue goes away but that's a separate discussion above.

lib/bitcoin/lock/LockMonitor.ts
mudiali: The only concern that I have w/ the removal of this is that now an external call to the `getCurrentValueTimeLock()` function can cause reads/re-broadcasts internally. I haven't formed an opinion on that because the usage pattern of the external REST API may prevent that but something to think about.

lib/bitcoin/lock/LockMonitor.ts
thehenrytsai: Doing this twice in one method seems to be "bad design", which could be exactly why you have the initial fetch as the original code. This elevates the discussion of removing `this.currentLockState()` even more I think.

lib/bitcoin/lock/LockMonitor.ts
thehenrytsai: Good point and related to the other comment, I think we address your concern by decouple the rebroadcast by pulling it out from `getCurrentValueTimeLock()`.

mudiali (approved):

Issue #884, opened 2020-10-10
Randomness in test fixture generation increases feature development time


OR13:
When a change is made, that impacts fixtures, it should be possible to run a single comment, to update them all.

And then all the tests should for the most part pass.

Right now, you have to navigate the source, console logging and editing files, it a way that is likely to lead to bugs, and which is certainly much slower that deterministically rebuilding test vectors.

thehenrytsai:
Done.

Issue #876, opened 2020-10-06
Fix all eslint errors


thehenrytsai:

chetangawai:
@thehenrytsai Can I take this up?

thehenrytsai:
Fixed.

@chetangawai, sorry didn't realized you nominated yourself to take on this. In the future, you are welcome to take up any issue any time, really appreciate your help!

PR #900, opened 2020-10-27
Issue #884 - Made test vector generation deterministic & #876 - Fixed all eslint errors

thehenrytsai:


tests/vectors/inputs/publicKeyModel1.json
```diff
@@ -0,0 +1,13 @@
+{
```
OR13, line 1: from a types perspective, this is almost a linked data key pair, see: https://github.com/transmute-industries/did-key.js/blob/master/packages/test-vectors/src/did-core-conformance/p-256/p-256.json#L5

in the future, we might want to rename this model "SidetreeKeyPair" or something like that.

OR13 (approved):

isaacJChen (approved):

Issue #897, opened 2020-10-23
Align remove-public-keys and remove-services on use of `ids` property


csuwildcat:
Currently `remove-public-keys` uses a `publicKeys` field for its array of key ID strings to be removed, whereas `remove-services` uses `ids` for the same array of ID strings. We should change `remove-public-keys` to use the same `ids` property to normalize this.

NOTE: this has already been changed in the spec to reflect the use of `ids` as the property name in both places.

troyronda:
Normalizing with `ids` makes sense to me.

thehenrytsai:
Done.

PR #904, opened 2020-10-27
fix(ref-imp): #897 - changed remove-public-keys to use ids

thehenrytsai:


lib/core/versions/latest/DocumentComposer.ts
```diff
@@ -226,29 +226,31 @@ export default class DocumentComposer {
...
-    if (patchProperties.length !== 2) {
-      throw new SidetreeError(ErrorCode.DocumentComposerPatchMissingOrUnknownProperty);
+    const allowedProperties = new Set(['action', 'ids']);
+    for (const property in patch) {
+      if (!allowedProperties.has(property)) {
+        throw new SidetreeError(ErrorCode.DocumentComposerUnknownPropertyInRemoveServicesPatch, `Unexpected property ${property} in remove-services patch.`);
+      }
     }
 
     if (!Array.isArray(patch.ids)) {
```
isaacJChen, line 256: Do we also need a check to see if action exists? Since we removed the length check, we no longer knows if a patch has both, we just know it doesn't contain extra keys.

thehenrytsai: I think it's okay, because these private method are invoked by `validatePatch` which already does that check.

isaacJChen (approved):

csuwildcat (approved):

PR #911, opened 2020-11-03
Align purpose with spec using verification method and verification relationships

isaacJChen:


lib/core/versions/latest/DocumentComposer.ts
```diff
@@ -24,10 +24,11 @@ export default class DocumentComposer {
 
     const document = didState.document as DocumentModel;
 
-    // Only populate `publicKey` if general purpose exists.
-    // Only populate `authentication` if authentication purpose exists.
-    const authentication: any[] = [];
-    const publicKeys: any[] = [];
+    // Put each public key in verificationMethod
+    // then populate the verification relationships by reference if a key has multiple purposes,
+    // or populate the verification relattionship by full description if a key has only 1 purpose
```
thehenrytsai, line 29: Sanity: I thought Daniel has confirm that we do by reference always?

lib/core/versions/latest/DocumentComposer.ts
```diff
@@ -24,10 +24,11 @@ export default class DocumentComposer {
...
+    const verificationRelationships: Map<string, Array<object | string>> = new Map();
```
thehenrytsai, line 30:
```suggestion
    const verificationRelationships: Map<string, string[]> = new Map();
```

tests/core/Resolver.spec.ts
```diff
@@ -89,7 +89,7 @@ describe('Resolver', () => {
       const published = true;
       const didState = await resolver.resolve(didUniqueSuffix) as DidState;
       const resultingDocument = DocumentComposer.transformToExternalDocument(didState, `did:sidetree:${didUniqueSuffix}`, published);
-      fixtureDriftHelper(resultingDocument, afterUpdate, 'resolution/afterUpdate.json', OVERWRITE_FIXTURES);
+      fixtureDriftHelper(resultingDocument, afterUpdate, 'resolution/afterUpdate.json', false);
```
thehenrytsai, line 92: Sanity: did you intend to make these false, if so, why?

thehenrytsai (changes requested):
Minor changes requested.

tests/core/Resolver.spec.ts
isaacJChen: Good catch. Not intended.

lib/core/versions/latest/DocumentComposer.ts
isaacJChen: Old comments before we made the decision, will update.

thehenrytsai (approved):

PR #914, opened 2020-11-06
feat(ref-imp): #766 - SIP 1 - Added downloading of proof files

thehenrytsai:


lib/core/versions/latest/InputValidator.ts
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * Class containing generic input validation methods.
+ */
+export default class InputValidator {
+  /**
+   * Validates that the given input is of a non-array object type.
+   * @param inputContextForErrorLogging This string is used for error logging purposes only. e.g. 'document', or 'suffix data'.
+   */
+  public static validateNonArrayObject (input: any, inputContextForErrorLogging: string) {
+    if (typeof input !== 'object') {
```
isaacJChen, line 13: Why assert against object? Can't it just directly do a check for array?

lib/core/versions/latest/CoreProofFile.ts
```diff
@@ -32,4 +52,100 @@ export default class CoreProofFile {
...
+    } catch (error) {
+      throw SidetreeError.createFromError(ErrorCode.CoreProofFileNotJson, error);
+    }
+
+    if (coreProofFileModel.operations === undefined) {
+      throw new SidetreeError(ErrorCode.CoreProofFileOperationsNotFound, `Core proof file does not have any operation proofs.`);
+    }
+
+    const operations = coreProofFileModel.operations;
+    const allowedProperties = new Set(['recover', 'deactivate']);
```
isaacJChen, line 82: Can't this just use the validateObjectContainsOnlyAllowedProperties function you created?

lib/core/versions/latest/ProvisionalProofFile.ts
```diff
@@ -29,4 +46,68 @@ export default class ProvisionalProofFile {
...
+    } catch (error) {
+      throw SidetreeError.createFromError(ErrorCode.ProvisionalProofFileNotJson, error);
+    }
+
+    if (provisionalProofFileModel.operations === undefined) {
+      throw new SidetreeError(ErrorCode.ProvisionalProofFileOperationsNotFound, `Provisional proof file does not have any operation proofs.`);
+    }
+
+    const operations = provisionalProofFileModel.operations;
+    const allowedProperties = new Set(['update']);
```
isaacJChen, line 76: Same for validateObjectContainsOnlyAllowedProperties

lib/core/versions/latest/InputValidator.ts
```diff
@@ -0,0 +1,37 @@
+import ErrorCode from './ErrorCode';
+import SidetreeError from '../../../common/SidetreeError';
+
+/**
+ * Class containing generic input validation methods.
+ */
+export default class InputValidator {
```
isaacJChen, line 7: Should this be versioned? I feel like this should be a general purposed helper. Or at the very least be in util?

tests/JasmineSidetreeErrorValidator.ts
```diff
@@ -10,15 +10,21 @@ export default class JasmineSidetreeErrorValidator {
...
+  public static expectSidetreeErrorToBeThrown (functionToExecute: () => any, expectedErrorCode: string, expectedContainedStringInMessage?: string): void {
     let validated: boolean = false;
 
     try {
       functionToExecute();
     } catch (e) {
       if (e instanceof SidetreeError) {
         expect(e.code).toEqual(expectedErrorCode);
+
+        if (expectedContainedStringInMessage !== undefined) {
```
isaacJChen, line 24: Yaaaay, I like this.

lib/core/versions/latest/CoreProofFile.ts
thehenrytsai: Yes it can, I added the lib method after this and was being lazy, was hoping to do a full codebase sweep separately, but will do this one now since you asked!

lib/core/versions/latest/ProvisionalProofFile.ts
thehenrytsai: Will do.

tests/JasmineSidetreeErrorValidator.ts
thehenrytsai: Do you also like `util:generate-error-codes`? =P

lib/core/versions/latest/InputValidator.ts
thehenrytsai: I think it is fine to be versioned (Based on the guiding principle of: put as much code into versioned folder as possible).
Good ask about putting the code in `util`, I also looked at the `util` folder but decided against it because the classes in there have nothing to do with Sidetree, and this one is arguable.

isaacJChen (approved):

PR #917, opened 2020-11-10
Change kid requirements for JWS headers to a MAY

OR13:
Possibly the language regarding how to verify a JWS that contains no key identifier needs to be updated as well.


csuwildcat:
I believe we should allow for the MAY be present directive on `kid`, as the JW* specs do, because there are valid cases where various cryptographic systems (e.g. DID Methods) can securely, explicitly identify a key and validate secret material without needing a `kid`. For example: in commitment schemes, you are committing to a specific, cryptographically locked key, and then subsequently revealing the key you committed to, which means you implicitly know exactly whether or not you're dealing with the right key.

csuwildcat (approved):

sandrask (approved):

troyronda (approved):

PR #929, opened 2020-11-18
Moving window normalized fee

isaacJChen:
- Calculate normalized fee
- Version protocol parameters


lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -144,15 +141,15 @@ export default class BitcoinProcessor {
...
+      this.versionManager
     );
   }
 
   /**
    * Initializes the Bitcoin processor
    */
   public async initialize () {
-    await this.versionManager.initialize();
+    await this.versionManager.initialize(this.blockMetadataStore);
```
thehenrytsai, line 152: Style: inject all dependencies from constructor.

lib/bitcoin/VersionManager.ts
```diff
@@ -1,37 +1,47 @@
...
     // NOTE: In principal each version of the interface implementations can have different constructors,
     // but we currently keep the constructor signature the same as much as possible for simple instance construction,
     // but it is not inherently "bad" if we have to have conditional constructions for each if we have to.
     for (const versionModel of this.versionsReverseSorted) {
       const version = versionModel.version;
+      const initialNormalizedFee = versionModel.protocolParameters.initialNormalizedFee;
+
+      this.protocolParameters.set(version, versionModel.protocolParameters);
 
       /* tslint:disable-next-line */
```
thehenrytsai, line 42: Opportunistic: remove tslint:disable (all remaining occurrences if you care to).

lib/bitcoin/BitcoinProcessor.ts
isaacJChen: I think it's better to do it in initialize because `VersionManager` doesn't need an instance of the store, it only takes it and use it to construct `Feecalculator`.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -405,7 +405,7 @@ export default class BitcoinProcessor {
    * Fetches Sidetree transactions in chronological order from since or genesis.
    * @param since A transaction number
    * @param hash The associated transaction time hash
-   * @returns Transactions in complete blocks since given transaction number.
+   * @returns Transactions in complete blocks since given transaction number, with normaliedFee.
```
thehenrytsai, line 408: Nit: typo. normalized

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -435,6 +435,26 @@ export default class BitcoinProcessor {
...
+        while (transaction.transactionTime !== blockMetaData[blockMetadataIndex].height) {
+          blockMetadataIndex++;
+          // This means that it has reached the end of block metadata and still cannot find the data for the current transaction.
+          // This can happen when a block reorg happens, and some block metadata are getting deleted.
+          if (blockMetadataIndex === blockMetaData.length) {
+            throw new RequestError(ResponseStatus.ServerError, ErrorCode.BitcoinBlockMetadataNotFound);
+          }
+        };
+        transaction.normalizedTransactionFee = blockMetaData[blockMetadataIndex].normalizedFee;
+      }
```
thehenrytsai, line 455: Optional: consider a  less performant but easier to read implementation that involves 2 un-nested loops: first loop constructs a map, 2nd loop uses the map.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -524,19 +544,42 @@ export default class BitcoinProcessor {
...
   /**
-   * Return proof-of-fee value for a particular block.
+   * Modifies the given array and update the normalized fees, then write to block metadata store.
+   * @param blocks the ordered block metadata to set the normalized fee for.
    */
-  public async getNormalizedFee (block: number): Promise<TransactionFeeModel> {
+  private async writeBlocksToMetadataStoreWithFee (blocks: BlockMetadataWithoutNormalizedFee[]) {
+    for (const block of blocks) {
+      const normalizedFee = (await this.getNormalizedFee(block.height)).normalizedTransactionFee;
+      this.blockMetadataStore.add([{
```
thehenrytsai, line 553: Maybe it's worthwhile to bulk insert just once at the end. (I say maybe 'coz I worry what would happen to CosmosDB rate limiting......)

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -524,19 +544,42 @@ export default class BitcoinProcessor {
...
 
   /**
-   * Return proof-of-fee value for a particular block.
+   * Modifies the given array and update the normalized fees, then write to block metadata store.
+   * @param blocks the ordered block metadata to set the normalized fee for.
    */
-  public async getNormalizedFee (block: number): Promise<TransactionFeeModel> {
+  private async writeBlocksToMetadataStoreWithFee (blocks: BlockMetadataWithoutNormalizedFee[]) {
+    for (const block of blocks) {
+      const normalizedFee = (await this.getNormalizedFee(block.height)).normalizedTransactionFee;
```
thehenrytsai, line 552: Should we file a improvement issue so that we have `getNormalizedFees` so that we make one range query to DB instead of many?

lib/bitcoin/versions/latest/NormalizedFeeCalculator.ts
```diff
@@ -1,19 +1,66 @@
...
+    // https://github.com/decentralized-identity/sidetree/issues/926
+    const unadjustedFee = Math.floor(totalFee / totalTransactionCount);
+
+    const previousFee = blocksToAverage[blocksToAverage.length - 1].normalizedFee;
+
+    return this.limitTenPercentPerYear(unadjustedFee, previousFee);
+  }
+
+  private limitTenPercentPerYear (unadjustedFee: number, previousFee: number): number {
+    const previousFeeAdjustedUp = Math.floor(previousFee * (1 + this.fluctuationRate));
```
thehenrytsai, line 53: Optional nit: consider `max/minAllowedFee`.

thehenrytsai (changes requested):
Some comments.

lib/bitcoin/lock/LockMonitor.ts
```diff
@@ -353,11 +355,12 @@ export default class LockMonitor {
     const currentLockIdentifier = LockIdentifierSerializer.deserialize(currentValueTimeLock.identifier);
     const currentLockDuration = currentValueTimeLock.unlockTransactionTime - currentValueTimeLock.lockTransactionTime;
 
+    const newLockDuration = this.versionManager.getLockDurationInBlocks(currentValueTimeLock.unlockTransactionTime);
```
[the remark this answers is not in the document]
isaacJChen: Let's talk about this.

lib/bitcoin/lock/LockResolver.ts
```diff
@@ -87,15 +85,17 @@ export default class LockResolver {
     // (C). verify that the lock duration is valid
     const unlockAtBlock = lockStartBlock + scriptVerifyResult.lockDurationInBlocks!;
 
+    const lockDurationInBlocks = this.versionManager.getLockDurationInBlocks(lockStartBlock);
+
     if (!this.isLockDurationValid(lockStartBlock, unlockAtBlock)) {
```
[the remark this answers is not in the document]
isaacJChen: The logic technically hasn't changed. Before, we were still doing the same thing where lock monitor uses the MAX value in protocol parameter to lock. The only difference it that now we can get a different number depending on version.

We still need this because
1. This allows you to pass in any lock and resolve it, even if a non sidetree node wrote it. It's more general purposed. (MY personal opinion)
2. In LockMonitor, we catch this specific error to set the lock state. It is set to pending status and no active lock.

lib/bitcoin/models/ProtocolParameters.ts
```diff
@@ -2,9 +2,20 @@
...
+  /** 
+   * The look back window for normalized fee calculation
+   * If this number is 10, then to calculate block X's normalized fee, it will look at blocks X - 10 to x - 1 to calculate.
+   */
+  lookBackWindowInterval: number;
+
+  /**
+   * The fluctuation rate cap. The normalized fee fluctuation cannot exceed this percentage. 1 being 100%.
+   */
+  fluctuationRate: number;
```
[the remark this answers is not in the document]
isaacJChen: I'm fine with per block. Per year requires another layer of thinking/math I think.

lib/common/models/TransactionModel.ts
```diff
@@ -7,6 +7,10 @@ export default interface TransactionModel {
   transactionTimeHash: string;
   anchorString: string;
   transactionFeePaid: number;
-  normalizedTransactionFee: number;
+
+  // Normalized fee sohuld always be populated in core layer when core makes call to transactions endpoint.
```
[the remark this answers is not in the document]
isaacJChen: Fair

lib/bitcoin/versions/latest/NormalizedFeeCalculator.ts
```diff
@@ -1,19 +1,66 @@
...
     console.log(`Initializing normalized fee calculator.`);
   }
 
-  public getNormalizedFee (_block: number): number {
-    // TODO: Issue #783 - Simplify normalized fee calculation algorithm.
-    return 10;
+  public async getNormalizedFee (block: number): Promise<number> {
+    if (block < this.genesisBlockNumber) {
+      // No normalized fee for blocks that exist before genesis
+      return 0;
```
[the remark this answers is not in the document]
isaacJChen: I think it should be handled because the it is a reachable path from the API's standpoint. I can ping bitcoin service manually with that request and break it. I don't think it should be treated as an unreachable path just because Core can't reach it.

lib/bitcoin/BitcoinProcessor.ts
isaacJChen: ok, I'm fine with moving config to init

lib/common/models/TransactionModel.ts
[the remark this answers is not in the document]
isaacJChen: Will create an issue then reference in code.

lib/bitcoin/BitcoinProcessor.ts
isaacJChen: Opened issue
https://github.com/decentralized-identity/sidetree/issues/936

lib/common/models/TransactionModel.ts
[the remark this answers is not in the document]
isaacJChen: https://github.com/decentralized-identity/sidetree/issues/937

lib/bitcoin/versions/latest/NormalizedFeeCalculator.ts
```diff
@@ -21,17 +22,26 @@ export default class NormalizedFeeCalculator implements IFeeCalculator {
   }
 
   public async getNormalizedFee (block: number): Promise<number> {
+    // DB call optimization
+    // https://github.com/decentralized-identity/sidetree/issues/936
```
thehenrytsai, line 26: Nit: standard patterns is to append `TODO` to enable searching and automated tooling (such as extension in VSC).
```suggestion
    // TODO: https://github.com/decentralized-identity/sidetree/issues/936
```

lib/common/models/TransactionModel.ts
```diff
@@ -8,8 +8,13 @@ export default interface TransactionModel {
   anchorString: string;
   transactionFeePaid: number;
 
-  // Normalized fee sohuld always be populated in core layer when core makes call to transactions endpoint.
-  // It may not be populated in blockchain service. This allows flexibility for the value to be computed on the spot or stored.
+  /**
+   * Normalized fee sohuld always be populated in core layer when core makes call to transactions endpoint.
+   * It may not be populated in blockchain service. This allows flexibility for the value to be computed on the spot or stored.
+   * To remove potentially dangerous behavior. Make a seperate model
+   * https://github.com/decentralized-identity/sidetree/issues/937
```
thehenrytsai, line 15: Nit: add TODO for tooling/searching.
```suggestion
   * TODO: https://github.com/decentralized-identity/sidetree/issues/937
```

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -149,8 +149,8 @@ export default class BitcoinProcessor {
   /**
    * Initializes the Bitcoin processor
    */
-  public async initialize () {
-    await this.versionManager.initialize(this.blockMetadataStore);
+  public async initialize (versionModels: BitcoinVersionModel[]) {
```
thehenrytsai, line 152: I guess `BitcoinProcessor` was the example I was looking for for dependency injection, so I still want to make sure we have an agreed upon pattern. Let's discuss but won't block the PR on this.

tests/JasmineSidetreeErrorValidator.ts
```diff
@@ -63,7 +65,7 @@ export default class JasmineSidetreeErrorValidator {
     }
 
     if (!validated) {
-      fail(`Expected error '${expectedErrorCode}' did not occur.`);
+      fail(`Expected error '${expectedErrorCode}' did not occur. Instead got '${actualError.code}'`);
```
thehenrytsai, line 68: LOL!

thehenrytsai (approved):

PR #938, opened 2020-11-25
feat(ref-imp): #766 - introduced create references to align with rest of operation references

thehenrytsai:


lib/core/versions/latest/models/SuffixDataModel.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * Internal data structure of the delta for each operation.
+ */
+export default interface SuffixDataModel {
+  deltaHash: string;
```
isaacJChen, line 5: Just realized... why is this different color than the line below...

isaacJChen (approved):

lib/core/versions/latest/models/SuffixDataModel.ts
thehenrytsai: There must be a reason, we just don't know.

PR #939, opened 2020-11-25
normalized fee perf improvement

isaacJChen:


lib/bitcoin/versions/latest/NormalizedFeeCalculator.ts
```diff
@@ -21,19 +26,46 @@ export default class NormalizedFeeCalculator implements IFeeCalculator {
...
-    // https://github.com/decentralized-identity/sidetree/issues/936
+    const shouldCalculate = this.shouldCalculateNormalizedFee(block);
+    if (!shouldCalculate.shouldCalculate) {
+      return shouldCalculate.fee!;
+    }
+    const blocksToAverage = await this.getLookBackBlocks(block);
+    return this.calculateNormalizedFee(blocksToAverage);
+  }
+
+  private shouldCalculateNormalizedFee (block: number): { shouldCalculate: boolean, fee: number | undefined } {
```
isaacJChen, line 60: Please give me a better name (cry)

lib/bitcoin/interfaces/IFeeCalculator.ts
```diff
@@ -6,4 +9,9 @@ export default interface IFeeCalculator {
    * Returns the fee for a particular block height.
    */
   getNormalizedFee (block: number): Promise<number>;
+
+  /**
+   * Returns the block with normalized fee added
+   */
+  addNormalizedFeeToBlock (blockMetadataWithoutFee: BlockMetadataWithoutNormalizedFee): Promise<BlockMetadata>;
```
thehenrytsai, line 16: Naming nit:
```suggestion
  addNormalizedFeeToBlockMetadata (blockMetadataWithoutFee: BlockMetadataWithoutNormalizedFee): Promise<BlockMetadata>;
```

lib/bitcoin/versions/0.10.0/NormalizedFeeCalculator.ts
```diff
@@ -1,10 +1,14 @@
+import BlockMetadata from '../../models/BlockMetadata';
```
thehenrytsai, line 1: Just a thought, should we detroy 0.10.0 folder altogether?

lib/bitcoin/versions/latest/NormalizedFeeCalculator.ts
```diff
@@ -21,19 +26,46 @@ export default class NormalizedFeeCalculator implements IFeeCalculator {
...
-      return this.initialNormalizedFee;
+      return { shouldCalculate: false, fee: this.initialNormalizedFee };
     }
-
-    const blocksToAverage = await this.getLookBackBlocks(block);
-    return this.calculateNormalizedFee(blocksToAverage);
+    return { shouldCalculate: true, fee: undefined };
   }
 
   private async getLookBackBlocks (block: number): Promise<BlockMetadata[]> {
```
thehenrytsai, line 71: Naming nit:
```suggestion
  private async getBlocksInLookBackWindow (block: number): Promise<BlockMetadata[]> {
```

lib/bitcoin/versions/latest/NormalizedFeeCalculator.ts
thehenrytsai: LOL. I'd suggest eliminating this method, and just implement the logic in both callers. I double checked, you don't need to handle pre-genesis case in both cases (`getNormalizedFee()` in `BitcoinProcessor` already does the same check and actually throws error which I think is the better behavior than returning 0), so code there would be more or less the same even without this method.

thehenrytsai (changes requested):
Minor changes requested.

lib/bitcoin/versions/latest/NormalizedFeeCalculator.ts
isaacJChen: Oh, true. Ok.

thehenrytsai (approved):

PR #942, opened 2020-11-26
feat(ref-imp): #783 - Added unit to initialNormalizedFee param + various renames

thehenrytsai:


lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -449,7 +449,7 @@ export default class BitcoinProcessor {
       for (const transaction of transactions) {
         const block = blockMetaDataMap.get(transaction.transactionTime);
         if (block !== undefined) {
-          transaction.normalizedTransactionFee = block.normalizedFee;
+          transaction.normalizedTransactionFee = Math.floor(block.normalizedFee);
```
isaacJChen, line 452: Why do we floor here instead of in the fee calculator?

lib/bitcoin/versions/latest/NormalizedFeeCalculator.ts
```diff
@@ -8,55 +8,76 @@ import IFeeCalculator from '../../interfaces/IFeeCalculator';
...
     }
-    const newFee = this.calculateNormalizedFee(this.blockMetadataCache);
-    const newBlockWithFee = Object.assign({ normalizedFee: newFee }, blockMetadata);
-    this.blockMetadataCache.push(newBlockWithFee);
-    this.blockMetadataCache.shift();
-    this.expectedBlockHeight++;
+
+    const normalizedFee = this.calculateNormalizedFee(this.cachedLookBackWindow);
+    const newBlockWithFee = Object.assign({ normalizedFee }, blockMetadata);
+    console.log(normalizedFee);
```
isaacJChen, line 61: Debug log? Either remove or make it more informative, I think.

lib/bitcoin/versions/latest/NormalizedFeeCalculator.ts
```diff
@@ -8,55 +8,76 @@ import IFeeCalculator from '../../interfaces/IFeeCalculator';
...
 
   public async getNormalizedFee (block: number): Promise<number> {
     if (block < this.genesisBlockNumber + this.feeLookBackWindowInBlocks) {
-      return this.initialNormalizedFee;
+      return this.initialNormalizedFeeInSatoshis;
     }
     const blocksToAverage = await this.getBlocksInLookBackWindow(block);
-    return this.calculateNormalizedFee(blocksToAverage);
+    
+    // TODO: @Isaac, why are we not getting the content fee from DB?
```
isaacJChen, line 74: We can. I don't think it will happen, but I was thinking if core asks for 1 block ahead, then we will need to calculate it, or else it will be undefined. Probably will never happen.

lib/bitcoin/versions/latest/NormalizedFeeCalculator.ts
thehenrytsai: Yeah... was left over from debugging... will remove!

lib/bitcoin/versions/latest/NormalizedFeeCalculator.ts
```diff
@@ -69,16 +90,16 @@ export default class NormalizedFeeCalculator implements IFeeCalculator {
       totalTransactionCount += blockToAverage.transactionCount;
     }
 
-    // TODO: #926 investigate potential rounding differences between languages and implemetations
+    // TODO: #926 investigate potential rounding differences between languages and implementations
     // https://github.com/decentralized-identity/sidetree/issues/926
     const unadjustedFee = Math.floor(totalFee / totalTransactionCount);
```
isaacJChen, line 95: We need to remove this floor as well.

lib/bitcoin/BitcoinProcessor.ts
thehenrytsai: Discussed offline.

lib/bitcoin/versions/latest/NormalizedFeeCalculator.ts
isaacJChen: Will change in my pr

thehenrytsai: Synced offline.

isaacJChen (approved):

PR #962, opened 2020-12-04
feat(ref-imp): #766 - updated API, resolver, batch writer to support revealValue in requests

thehenrytsai:


lib/core/versions/latest/ProvisionalIndexFile.ts
```diff
@@ -128,8 +128,7 @@ export default class ProvisionalIndexFile {
     chunkFileUri: string, provisionalProofFileUri: string | undefined, updateOperationArray: UpdateOperation[]
   ): Promise<Buffer> {
     const updateReferences = updateOperationArray.map(operation => {
-      const revealValue = Multihash.canonicalizeThenHashThenEncode(operation.signedData.updateKey);
```
sandrask, line 131: What is the rationale for adding reveal value to update, recover and deactivate requests instead of just calculating it here?

isaacJChen: Correct me if I'm wrong, Henry. It is because the commitment hash can be calculated using a different hashing algorithm. If we add support for another hashing algorithm, then we won't know which one to use unless we try all of them, which is not efficient. By supplying the reveal value, we can derive the hashing algorithm and use it to hash the update key for verification and add backward and forward compatibility between commit reveal values.

thehenrytsai: @isaacJChen is correct, but I will add that it is not necessarily the calculating hashes of all supported type that is not ideal (granted it is definitely sub-optimal), it is that the index files will then be forced to store all versions of the hashes, bloating the index files significantly.

sandrask: When client is assembling let's say recover request are they supposed to use same implementors multi-hashing algorithm for generating both commitment and reveal value?

thehenrytsai: Each implementer will/may add support for different hash overtime, lets say the commitment is made with SHA2 in an earlier operation, but a requester enacts the next operation at a future time when SHA3 is supported, if the batch writer is to calculate the reveal value, it wouldn't know to use SHA2 or SHA3, so the best it can do is do both, which leads to index file size bloat.

(This also leads to a separate but mostly unrelated realization that one can never deprecate a hash algorithm due to long form DIDs, but we can discuss that separately, I will make a final PR for this)

sandrask: @thehenrytsai I am not sure if this approach will work. Do you have a test case for when protocol switches between multihash codes? For example,  operation #1 was created with 18(sha2_256) and operation #2 was created with 19(sha2-512).

thehenrytsai: Current test lib hardcodes hash algorithm when generating operations, but let me see try to do what it takes to have this test that you are asking for.

Meanwhile please do share why you think it will not work so I don't waste valuable time, but regardless, the batch writer as is without this PR will definitely not work correctly when hash algorithm support changes (see the TODO portion in the `BatchWriter`).

sandrask: For scenario I stated above it will not work because of double hashing e.g. op 1 commitment value is sha256+sha256 and op 2 reveal value if calculated with current implementors protocol is sha512. Are you expecting that client will know that previous operation was done in code sha256 and it will send correct reveal value. In this case the client needs to  manage reveal value too, not just the key.

sandrask: Also, you still need to take that op 2 reveal value and calculate commitment hash with old multihash code (18) in order to match to op 1 commitment value. It will work if the client with op2 sends you correct reveal value (calculated with code 18).

thehenrytsai: Using your scenario, reveal value in op 2 will be stored using the reveal value given in request which is SHA256, NOT SHA512. This happens at this exact line you commented on.

Yes on expectation of client, client would know which algorithm to use because it is specified by the commitment they created (since it is a multihash).

sandrask: Ok, thanks. The client is responsible to create proper reveal value for current operation based on previous operation commitment.

isaacJChen (approved):

sandrask (approved):

PR #967, opened 2020-12-05
feat(ref-imp): #766 - added support to validate reveal value as a hash

thehenrytsai:
1. Added support to validate reveal value as a hash.
1. Added test for applying hash algorithm to operation different from the hash algorithm used by the reveal value.


lib/core/versions/latest/Multihash.ts
```diff
@@ -65,8 +65,9 @@ export default class Multihash {
     const contentBuffer = JsonCanonicalizer.canonicalizeAsBuffer(content);
 
     // Double hash.
-    const intermediateHashBuffer = Multihash.hashAsNonMultihashBuffer(contentBuffer, ProtocolParameters.hashAlgorithmInMultihashCode);
-    const multihashEncodedString = Multihash.hashThenEncode(intermediateHashBuffer, ProtocolParameters.hashAlgorithmInMultihashCode);
+    const hashAlgorithmInMultihashCode = 18; // Default to SHA256.
```
isaacJChen, line 68: Should we make this an optional param to make future test changes potentially easier?

thehenrytsai: I considered it but decided against it, since it there is no need for it right now and it is easy to change when we need it.

isaacJChen (approved):

PR #972, opened 2020-12-08
Increase test coverage to over 99%

isaacJChen:


tests/core/DownloadManager.spec.ts
```diff
@@ -12,39 +12,90 @@ describe('DownloadManager', async () => {
 
   const originalDefaultTestTimeout = jasmine.DEFAULT_TIMEOUT_INTERVAL;
 
-  beforeAll(() => {
-    jasmine.DEFAULT_TIMEOUT_INTERVAL = 20000; // These asynchronous tests can take a bit longer than normal.
+  describe('constructor', () => {
+    it('should use default values if maxConcurrentDownloads is NaN', () => {
```
thehenrytsai, line 16: Do you think we should just remove this check in code?

thehenrytsai (approved):

tests/core/Blockchain.spec.ts
```diff
@@ -10,6 +10,33 @@ import TransactionModel from '../../lib/common/models/TransactionModel';
 import ValueTimeLockModel from '../../lib/common/models/ValueTimeLockModel';
 
 describe('Blockchain', async () => {
+  describe('startPeriodicCachedBlockchainTimeRefresh', () => {
+    beforeEach(() => {
+      // Freeze time
+      jasmine.clock().install();
```
thehenrytsai, line 16: Sanity: what's this jasmine.clock().install(); witchcraft? Was it left over during debug or something?

tests/core/DownloadManager.spec.ts
isaacJChen: I can. The only usage of this I can think of is if the service admin puts in a bad value, this will correct it. I can make a low prio issue and if we want to remove it, we can.

Issue #978, opened 2020-12-11
Add DB upgrade support to core service


thehenrytsai:

thehenrytsai:
Implemented.

PR #979, opened 2020-12-11
feat(ref-imp): #978 - Added DB upgrade support to core service

thehenrytsai:


lib/core/models/Config.ts
```diff
@@ -4,7 +4,7 @@
 export default interface Config {
   batchingIntervalInSeconds: number;
   blockchainServiceUri: string;
-  databaseName: string | undefined;
```
isaacJChen, line 7: No default db name?

isaacJChen (approved):

lib/core/models/Config.ts
thehenrytsai: Great call. Made all `databaseName` required.

isaacJChen (approved):

Issue #440, opened 2020-03-12
Revisit, finalize, and implement logging strategy


thehenrytsai:
Currently there are two camps for the logging strategy:
1. Use a dedicated logger.
1. Use console logging with support for hooks to pick up the logs as needed.

OR13:
I think I prefer a dedicated logger, but not strongly.

tplooker:
Element uses a custom logger implementation @gjgd will link a reference

gjgd:
This is what we are doing for element: https://github.com/decentralized-identity/element/blob/405488e92867c318b1fe40c3af3b8d7a64dea99e/packages/element-lib/src/sidetree/logger.js

thehenrytsai:
Implemented.

PR #986, opened 2020-12-16
feat(ref-imp): #440 - added custom logger support

thehenrytsai:


lib/bitcoin/BitcoinBlockDataIterator.ts
```diff
@@ -1,6 +1,7 @@
 import BitcoinBlockModel from './models/BitcoinBlockModel';
 import BitcoinFileReader from './BitcoinFileReader';
 import BitcoinRawDataParser from './BitcoinRawDataParser';
+import logger from '../common/Logger';
```
mudiali, line 4: Curious about casing: any reason to use lower case `logger` instead of `Logger`?

lib/core/Core.ts
```diff
@@ -65,7 +66,14 @@ export default class Core {
    * The initialization method that must be called before consumption of this core object.
    * The method starts the Observer and Batch Writer.
    */
-  public async initialize () {
+  public async initialize (customLogger?: ISidetreeLogger) {
```
isaacJChen, line 69: Ali suggested using singleton and delegating functionality to improve type safety. Let's talk about it. I think it's not a bad idea.

lib/common/ConsoleLogger.ts
```diff
@@ -0,0 +1,18 @@
+import ILogger from './interfaces/ILogger';
```
mudiali, line 1: As per our offline conversation, think about this pattern:

```` javascript
class ILogger {
 ...
}

=============

class ConsoleLogger implements ILogger {

}

=============

export class Logger{
  private static singleton: ILogger;
  
  static initilize(loggerImpl: ILogger) {
    singleton= loggerImpl;
  }  

  private checkSingleton() { 
    if (!singleton) { singleton = new ConsoleLogger(); }
  }

  static info (..) { 
    checkSingleton();
    singleton.info ... 
  }
}

===================

class Core {

  constructor(logger?: ILogger)
  
    if (logger) {
      Logger.initilize(logger);
      
    else {
      Logger.initialize(new ConsoleLogger());
     }
}
````

lib/bitcoin/BitcoinBlockDataIterator.ts
thehenrytsai: Trying to use the same pattern like `console.info()` (e.g. `logger.info()`). i.e. `Logger` is a class, `logger` is an instance.

lib/core/Core.ts
thehenrytsai: Took Ali's suggestion.

lib/bitcoin/BitcoinBlockDataIterator.ts
thehenrytsai: Took Ali's suggestion.

isaacJChen (approved):

PR #987, opened 2020-12-16
chore(ref-imp): more tests to cover branches

isaacJChen:


tests/core/Observer.spec.ts
```diff
@@ -438,6 +438,56 @@ describe('Observer', async () => {
     expect(processedTransactions[3].anchorString).toEqual('4thTransaction');
   });
 
+  it('should log error if blockchian throws', async () => {
```
isaacJChen, line 441: Explicit test that tests the weirdly missing test.

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -131,8 +131,8 @@ export default class BitcoinProcessor {
 
     this.mongoDbLockTransactionStore = new MongoDbLockTransactionStore(config.mongoDbConnectionString, config.databaseName);
 
-    const valueTimeLockTransactionFeesInBtc = config.valueTimeLockTransactionFeesAmountInBitcoins === 0 ? 0
-      : config.valueTimeLockTransactionFeesAmountInBitcoins || 0.25;
+    const valueTimeLockTransactionFeesInBtc = config.valueTimeLockTransactionFeesAmountInBitcoins === undefined ? 0.25
```
thehenrytsai, line 134: Sanity: This hard coded value looks wrong to me. I don't think this should be hardcoded at all?!

tests/core/Observer.spec.ts
thehenrytsai: Sanity: Shouldn't this test be 'should rethrow if unexpected error when fetching transaction from blockchain`?

tests/bitcoin/BitcoinClient.spec.ts
```diff
@@ -1100,6 +1102,44 @@ describe('BitcoinClient', async () => {
       done();
     });
 
+    it('should call retry-fetch without authohrization', async (done) => {
```
thehenrytsai, line 1105: Just a comment: I am also inclined to say the case of not having authorization is unnecessary bloat as from memory Bitcoin Client requires auth for you to use its API, and this is trying to handle a a hypothetical case where an alternative implementation does not require it.

tests/core/Observer.spec.ts
isaacJChen: There's a catch and log in `processTransactions`

lib/bitcoin/BitcoinProcessor.ts
isaacJChen: Added todo

thehenrytsai (approved):

Issue #997, opened 2021-01-06
Increment `updateCommitment` if an update patch fails to be applied


thehenrytsai:

thehenrytsai:
Implemented.

PR #998, opened 2021-01-06
fix(ref-imp): #997 - increment updateCommitment if failed to apply an update patch

thehenrytsai:


lib/core/versions/latest/OperationProcessor.ts
```diff
@@ -163,26 +163,24 @@ export default class OperationProcessor implements IOperationProcessor {
...
+    const newDidState = {
+      nextRecoveryCommitmentHash: didState.nextRecoveryCommitmentHash,
+      document: didState.document,
+      nextUpdateCommitmentHash: operation.delta.updateCommitment,
+      lastOperationTransactionNumber: anchoredOperationModel.transactionNumber
+    };
+
     let resultingDocument;
     try {
       resultingDocument = await DocumentComposer.applyUpdateOperation(operation, didState.document);
```
isaacJChen, line 176: I think the second argument needs to be a deep copy, something like Object.assign({}, didState.docuemnt) should work.

tests/core/OperationProcessor.spec.ts
```diff
@@ -693,10 +694,16 @@ describe('OperationProcessor', async () => {
           operationIndex: 2
         };
 
+        // Intentionally failing update patches.
         spyOn(DocumentComposer, 'applyUpdateOperation').and.throwError('Expected test error');
```
isaacJChen, line 698: Instead of directly throwing error, we can change some attribute in the document, then throw. The expect on line 707 can then be a test to make sure documents come out the same.

thehenrytsai: Thanks for the great suggestion!

lib/core/versions/latest/OperationProcessor.ts
thehenrytsai: Thanks for the great catch!

tests/core/OperationProcessor.spec.ts
```diff
@@ -693,10 +694,19 @@ describe('OperationProcessor', async () => {
...
 
+        // The expected outcome of patch application failure (but all integrity checks including schema checks are passed )is:
+        // 1. Operation is considered successful.
+        // 2. Update commitment is updated/incremented.
+        // 3. DID document state remains the unchanged (same as prior to the patches being applied).
         const newDidState = await operationProcessor.apply(anchoredUpdateOperationModel, didState);
-        expect(newDidState!.lastOperationTransactionNumber).toEqual(1);
+        expect(newDidState!.lastOperationTransactionNumber).toEqual(2);
+        expect(newDidState!.nextUpdateCommitmentHash).toEqual(nextUpdateCommitment);
         expect(newDidState!.document).toBeDefined();
```
isaacJChen, line 710: We should probably expect it to be the same as a deep copy of the original document in didState.

lib/core/versions/latest/util/JsObject.ts
```diff
@@ -0,0 +1,33 @@
+/**
+ * Class containing JavaScript object operations.
+ */
+export default class JsObject {
+  /**
+   * Deep copies the given input.
+   */
+  public static deepCopyObject (input: any): any {
```
isaacJChen, line 8: Maybe it's good to have specific tests for this. Something simple like

```
const a = {number: 1, object: {x: 'x'}, array: [1,2,3]};
const b = deepCopyObject(a);

a.number = 2;
a.object.x = 'y';
a.array[0] = 4;

expect(a.number).not.toEqual(b.number);
expect(a.object).not.toEqual(b.object);
expect(a.array).not.toEqual(b.array);
```

But this is indirectly tested so I'm ok if you don't want to add it.

lib/core/versions/latest/util/JsObject.ts
```diff
@@ -0,0 +1,33 @@
...
+      deepCopy[key] = JsObject.deepCopyObject(value);
+    }
+
+    return deepCopy;
+  }
+
+  /**
+   * Clears all the properties in the given object.
+   */
+  public static clearObject (input: any) {
```
isaacJChen, line 28: we can have a test that makes sure the input turns into an empty object

thehenrytsai: I added a test already to make sure the usage of this method shows that it is working which in my opinion is more meaningful. But sure, I'll add a unit test specifically for this method.

lib/core/versions/latest/util/JsObject.ts
thehenrytsai: Sure, I'll add it.

isaacJChen (approved):

Issue #995, opened 2021-01-05
Generated DID Documents should use "@base" consistently


jmandel:
In 0.12.1, when I generate a DID Document I get something like the example I've pasted below. In this example, there's some inconsistency about when the JSON-LD `@base` is used:

* `@base` is defined as part of the document context
* Sometimes the base semantics are leveraged, like:
  * my `service` has `"id":"#linked-domain"`
* Sometimes the base semantics are not leveraged, like
  *  my `authenticationMethod`s have `"id": "did:ion:EiAal0h<snip>#signing-key-1` instead of `"id": "#signing-key-1"`
  *  my `controller`s have `"id": "did:ion:EiAal0h<snip>"` instead of `"id": ""`
 
Note that the *semantics* seem right here (i.e., parsing this as a JSON-LD doc yields the right RDF triples); it's just that the serialized form is longer than it needs to be *and* I'm confused about why we're only sometimes leveraging the `@base` semantics. It'd be good to have more consistency and shorter payloads here, in my view. I think this is especially important because most developers *won't* be parsing these as JSON-LD documents. (I expect https://www.w3.org/TR/did-core/#json will help avoid any confusion here, but as far as I can tell this isn't supported in the sidetree core library -- and yet the sidetree core library isn't explicitly using a JSON-LD processor either.)

```json
{
[...]

PR #1000, opened 2021-01-08
Align resolution object with did core spec

isaacJChen:
Also addresses https://github.com/decentralized-identity/sidetree/issues/995


lib/core/versions/latest/DocumentComposer.ts
```diff
@@ -91,13 +105,27 @@ export default class DocumentComposer {
     const didResolutionResult: any = {
       '@context': 'https://www.w3.org/ns/did-resolution/v1',
```
OR13, line 106: https://w3id.org/did-resolution/v1 would be better than a URL that is 404 :)

tests/core/DocumentComposer.spec.ts
```diff
@@ -27,13 +27,17 @@ describe('DocumentComposer', async () => {
       };
 
       const published = true;
-      const result = DocumentComposer.transformToExternalDocument(didState, 'did:method:suffix', published);
+      const isShortForm = true;
+      const result = DocumentComposer.transformToExternalDocument(didState, 'did:method:suffix', published, isShortForm);
 
       expect(result['@context']).toEqual('https://www.w3.org/ns/did-resolution/v1');
```
OR13, line 33: https://w3id.org/did-resolution/v1

lib/core/versions/latest/DocumentComposer.ts
isaacJChen: OH I didn't even check it, thanks for catching it!

lib/core/versions/latest/DocumentComposer.ts
```diff
@@ -18,10 +18,24 @@ export default class DocumentComposer {
   /**
    * Transforms the given DID state into a DID Document.
    */
-  public static transformToExternalDocument (didState: DidState, did: string, published: boolean): any {
+  public static transformToExternalDocument (didState: DidState, did: string, published: boolean, isShortForm: boolean): any {
     // If the DID is deactivated.
+    // Return required metadata and a document with only context and id if so
     if (didState.nextRecoveryCommitmentHash === undefined) {
-      return { status: 'deactivated' };
+      const didDocument = {
```
thehenrytsai, line 25: Consider move this block into its own method perhaps `generateDeactivatedResolutionObject(...)` since this method is becoming large.

lib/core/versions/latest/DocumentComposer.ts
```diff
@@ -91,13 +105,27 @@ export default class DocumentComposer {
...
+      didDocumentMetadata: {
+        method: {
+          published,
+          recoveryCommitment: didState.nextRecoveryCommitmentHash,
+          updateCommitment: didState.nextUpdateCommitmentHash
+        }
       }
     };
 
+    if (!isShortForm) {
```
thehenrytsai, line 117: Nit: consider just do `if (isShortForm) {` first.

lib/core/versions/latest/RequestHandler.ts
```diff
@@ -177,8 +177,8 @@ export default class RequestHandler implements IRequestHandler {
       // We reach here it means there is a DID Document to return.
 
       // If DID is published, use the short-form DID; else use long-form DID in document.
-      const didStringToUseInDidDocument = published ? did.shortForm : did.longForm!;
-      const document = DocumentComposer.transformToExternalDocument(didState, didStringToUseInDidDocument, published);
+      const didStringToUseInDidDocument = shortOrLongFormDid;
+      const document = DocumentComposer.transformToExternalDocument(didState, didStringToUseInDidDocument, published, did.isShortForm);
```
thehenrytsai, line 181: Design consideration, what if we do:
```suggestion
      const document = DocumentComposer.transformToExternalDocument(didState, did, published);
```
Since
1. `didStringToUseInDidDocument ` is no longer 100% true since we transform it to short form in some cases. and
2. You won't need to repeat long-form to short-form logic.

isaacJChen: Good point. I like it.

thehenrytsai (approved):

sandrask:
@isaacJChen @thehenrytsai Couple of questions regarding this PR: 
1. Why is top level context ("@context": "https://w3id.org/did-resolution/v1") missing in https://raw.githubusercontent.com/decentralized-identity/sidetree/master/tests/vectors/resolution/afterDeactivate.json
2. Test vectors have an empty string for 'controller' in 'verificationMethod'. Does empty string for “controller” violate the spec? Should we omit “controller” if it is empty?
Relevant sections of did core spec:  "A DID document MAY include a controller property to indicate the DID controller(s)."  and "When a controller property is present in a DID Document, its value expresses one or more DIDs."

PR #1006, opened 2021-01-20
chore(ref-imp): remove controller and add context to deactivated docu…

isaacJChen:
…ment

### DIFF
--- a/tests/vectors/resolution/afterUpdate.json
+++ b/tests/vectors/resolution/afterUpdate.json
@@ -18,7 +18,6 @@
     "verificationMethod": [
       {
         "id": "#publicKeyModel1Id",
-        "controller": "",
         "type": "EcdsaSecp256k1VerificationKey2019",
         "publicKeyJwk": {
           "kty": "EC",
@@ -29,7 +28,6 @@
       },
       {
         "id": "#additional-key",
-        "controller": "",
         "type": "EcdsaSecp256k1VerificationKey2019",
         "publicKeyJwk": {
           "kty": "EC",
--- a/lib/core/versions/latest/DocumentComposer.ts
+++ b/lib/core/versions/latest/DocumentComposer.ts
@@ -16,6 +16,9 @@ import SidetreeError from '../../../common/SidetreeError';
  */
 export default class DocumentComposer {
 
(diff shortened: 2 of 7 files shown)


thehenrytsai (approved):

tests/vectors/resolution/afterUpdate.json
```diff
@@ -18,7 +18,6 @@
     "verificationMethod": [
       {
         "id": "#publicKeyModel1Id",
-        "controller": "",
```
OR13, line 21: please read did core before making changes to rest vectors.... https://w3c.github.io/did-core/#verification-methods

> The properties MUST include the id, type, controller, and specific verification method properties, and MAY include additional properties.

I ideally you would also have tests that showed that JSON-LD was valid.

sandrask: This change is my fault. I read the spec but I looked at wrong section (controller at top level of the document). Sorry about this. Can the controller be empty though?

OR13: the controller should equal `didDocument.id` in order for verifiable credentials to be valid.... see: https://github.com/OR13/did-params-and-you/blob/master/examples/case-0-legal-working.json#L35

and our version of sidetree: https://github.com/transmute-industries/sidetree.js/blob/main/packages/core/src/DocumentComposer.ts#L65

If his repo had CI tests that used the result of test vectors, with verifiable credentials, it would be obvious when breaking changes like this were introduced.

btw, the reason that "" is legal here, is because of `@base` / relative IRIs in sidetree did documents....

You can check the nquads to see that its the same...

if the iri is blank, it will get mapped to the didDocument.id... making what we had before legal (but only for relative ref based did documents).

sandrask: So it is legal because it will be populated when processing @base - thank you for this clarification.

OR13: I strongly recommend against returning the empty string for the controller property.... if makes sidetree look different than every other did method, and its not guaranteed to work accross JSON-LD implementations... the controller property should match the didDocument.id.

isaacJChen: Thanks for catching it. I will have a PR out fixing this soon.

PR #1009, opened 2021-01-20
chore(ref-imp): add controller back into verification methods

isaacJChen:


tests/vectors/resolution/afterCreate.json
```diff
@@ -18,6 +18,7 @@
     "verificationMethod": [
       {
         "id": "#publicKeyModel1Id",
+        "controller": "",
```
troyronda, line 21: Controller should not be an empty string. Prefer to omit property rather than have an empty string.

troyronda (changes requested):

tests/vectors/resolution/afterCreate.json
isaacJChen: It has to be there according to did-core spec
image

isaacJChen: We have a base in @context, which fills it in with JSON-LD

troyronda: I see.

Note: personally, I think this style makes the resolution result less clear to the human reader (and also by simpler processors.)

isaacJChen: Ya, I agree. I removed it in my last PR based on WG discussion results. We can further discuss it if we want the plain DID in there and not infer it separately. Having it there as empty string at least is spec compliant and was the previous state so I decided to revert it. Feel free to bring this up in the next WG meeting though :)

thehenrytsai (approved):

tests/vectors/resolution/afterCreate.json
thehenrytsai: @troyronda, this is a follow up PR for discussion that occurred in https://github.com/decentralized-identity/sidetree/pull/1006, thus @sandrask is aware of this change and I believe we have consensus, thus I am approving this PR to unblock.

troyronda: @thehenrytsai I have opened a followup https://github.com/decentralized-identity/sidetree/issues/1010 as I do not think using an empty string is the best outcome.

PR #1030, opened 2021-02-10
Add concept of operation anchor timing and validity

troyronda:
This PR introduces the concept of a stale operation.

A Sidetree-based DID Method may reject stale operations. An operation is considered stale relative to the timing information provided by the underlying anchoring system. When an operation is stale according to the DID method’s parameters, the operation is deemed as invalid.


troyronda:
@csuwildcat Also made it clear in this PR that an implementation MAY define these properties.

csuwildcat (approved):

Issue #1034, opened 2021-02-12
Allow for a partitioned anchoring system


troyronda:
In some anchoring systems, anchors can be partitioned among multiple systems rather than being held in a single common system. In other words, the overall system becomes a network of anchoring systems.

In these situations, observers need to know which system(s) know the most recent anchor for a DID. This property should be set by the DID controller/owner during `Create` and `Recover`. When an observer has previously processed a Sidetree suffix from the network of anchoring systems, that observer can also determine if they hold the most recent anchor based on that property.

To enable a Sidetree-based DID method that has a partitioned anchoring system, we should add an optional property to `Create` and `Recover` that enables an observer to find the head anchor for a DID. E.g., the property could be called `anchorSources`, `anchorHeadSources` or similar. DID methods that have a single common system can omit this property as the source is implicitly already defined.

csuwildcat:
Could we go with `anchorOrigin` or `anchorOrigins`? feel like that might be a bit more self-explanatory.

troyronda:
@csuwildcat updated PR with anchorOrigins

PR #1035, opened 2021-02-12
Add concept of anchor origin

troyronda:
See #1034 for background.


OR13 (dismissed):

csuwildcat (changes requested):
We should probably specify that the properties are MAY implement as well as MAY be present, because in implementations that don't implement the props, they will likely (should?) throw on the presence of unrecognized properties.

troyronda:
@csuwildcat Sure - will push an update.

troyronda:
@csuwildcat Added clarifying language that the property is implementation-defined and MAY be included if the implementation defines it.

csuwildcat (approved):

PR #1047, opened 2021-02-17
Add JWK nonce concept to spec

troyronda:


docs/spec/common-functions.md
```diff
@@ -28,4 +28,11 @@ The following steps define the [commitment scheme](#commitment-scheme) for gener
 2. Canonicalize the [JWK](https://tools.ietf.org/html/rfc7517) encoded public key using the implementation's [`JSON_CANONICALIZATION_SCHEME`](#json-canonicalization-scheme).
 3. Use the implementation's [`HASH_PROTOCOL`](#hash-protocol) to the hash the canonicalized public key to generate the [`REVEAL_VALUE`](#reveal-value), then hash the resulting hash value again using the implementation's [`HASH_PROTOCOL`](#hash-protocol) to produce the [public key commitment](#public-key-commitment).
 
-Implementers ****MUST NOT**** re-use public key JWK payloads across different commitment invocations.
\ No newline at end of file
+Implementers ****MUST NOT**** re-use public key JWK payloads across different commitment invocations.
```
csuwildcat, line 31: Seems like we should change this to something like: "For maximum forward cryptographic security, implementers SHOULD NOT re-use public keys across different commitment invocations", given we are allowing them a way to do exactly that. I am fine with adding the nonce, I just think we might want to mod the normative directive so it doesn't confuse people/

csuwildcat (changes requested):

docs/spec/common-functions.md
troyronda: I'll add the sentence. I assume the existing sentence should remain as the JWK, itself, MUST still change across invocations.

troyronda: Added. @csuwildcat

csuwildcat (approved):

PR #1053, opened 2021-02-23
Add DID URI parsing note and Guidelines section

csuwildcat:


docs/spec/did-uri.md
```diff
@@ -16,6 +16,10 @@ Example of Short-Form DID URI:
 did:sidetree:EiDahaOGH-liLLdDtTxEAdc8i-cfCz-WUcQdRJheMVNn3A
 ```
 
+::: note
+Many implementations have multiple active network instances of their DID Method (e.g. mainnet and testnet). How different network instances of a DID Method are represented in the DID URI string is method-specific. Many methods choose to use the base format above (`did:METHOD`) as their primary/mainnet network, and add an additional segment after the `:METHOD` segment to denote other network instances, for example: `did:METHOD:testnet`. DID Methods ****SHOULD**** clearly describe parsing rules for distinguishing between their different network instances.
```
troyronda, line 20: It would be good to mention that:

some methods add multiple segments after the `:METHOD` segment.

troyronda (approved):

Issue #561, opened 2020-04-14
Sidetree Operation Test Vectors in Spec


OR13:
JSON Create Payload => did 
JSON Update Payload  (contains a compact JWS) kid is in header
JSON Recovery Payload  (contains a compact JWS) kid is in header
JSON Deactivate Payload  (contains a compact JWS) kid is in header

OR13:
related https://github.com/w3c/did-core/issues/259

troyronda:
We likely will have some canonicalization discussion when thinking about the inputs to the test vectors.

troyronda:
(if there are test vectors on how the client produced the encoded value from JSON).

OR13:
Targeting May 19th for completion.

julio-cabdu:
I've also been wondering about the use of kid, is it gonna be supported in the reference implementation code?

OR13:
@thehenrytsai @csuwildcat @isaacJChen Can one / all of you add examples of the JWS used by sidetree here: https://github.com/decentralized-identity/did-jose-extensions

@julio-cabdu short answer is that sidetree reference implementation has fixtures like this: https://github.com/decentralized-identity/sidetree/blob/06152ae3a7062f64192ecd180e137c5ad5a76e5d/tests/fixtures/update/update.json

`signed_data` is is a normal JWS... it looks like [this](https://jwt.io/#debugger-io?token=eyJhbGciOiJFUzI1NksifQ.eyJ1cGRhdGVfa2V5Ijp7Imt0eSI6IkVDIiwiY3J2Ijoic2VjcDI1NmsxIiwieCI6IjFVZE5yVmFwZHVQdDhIVlNhSmpRZGk3WWQyVjR0YmRlajF5bkdKbmdSZXciLCJ5IjoicXpyd0dSWmdLbEwyTU5UVElJLWFmUE9sQWhiM2VhVkNzanBGWWtCNEFrVSJ9LCJkZWx0YV9oYXNoIjoiRWlDR2daN015cFNXcUdxOTRLOTh0ZkV1aGc2cmw1Z1Z4aVBaQnRWQUNHS20ydyJ9.7R4IbXxphOhaJgp-b3zvnQsrJFJCkXBRANQKSsURaTQWB71cJ9-klwze9Q5DXyNg4_UAaxXrmuQf_j7Y0RJA6g)

You can see the JWS header does not contain a `kid`...

```
{
  "alg": "ES256K"
}
```

[...]

Issue #550, opened 2020-04-10
Add precise language on DID unique suffix generation and test vector 


thehenrytsai:
Example test vectors can be found at: https://tools.ietf.org/html/rfc3394

troyronda:
Related note:

https://identity.foundation/sidetree/docs/spec/#did-suffix-composition refers to the
> Create operation’s Suffix Data Object

but this object isn't defined in that section.

troyronda:
Another note: multi-hash isn't mentioned in the spec.

csuwildcat:
**WIP:**

DID Methods based on the Sidetree protocol all share the same identifier format. The unique identifier string is a hash of a set of initialization values the controller generates when they create a Sidetree-based DID. This composite hash of initial state values enables Sidetree DIDs to be _self-certifying_, meaning the person or entity who generates a Sidetree-based DID knows their unique identifier immediately, and it is cryptographic secured for instant use (for more on instant use capabilities of Sidetree DIDs see the [Long-Form Sidetree Format](#) section below).

The process below MUST be used to correctly generate a Sidetree-based DID identifier string:

1. From the values generated during the [Create](#create) operation process, assemble an [_Anchor File Create Entry_](#anchor-file-create-entry) object, composed as follows:

    ```
    {
      "delta_hash": DELTA_HASH,
      "recovery_key": JWK_OBJECT,
      "recovery_commitment": COMMITMENT_HASH
    }
    ```
    - The object MUST contain a `delta_hash` property, and its value MUST be a hash of the [Create Operation Delta Object](create-data-object).
    - The object MUST contain a `recovery_key` property, and its value MUST be the _Initial Recovery Public Key_.
    - The object MUST contain a `recovery_commitment` property, and its value MUST be the _Initial Recovery Commitment_

OR13:
suggested change:

HTTP POST http://example.com/operations { body } -H { accept: application/json }

body : {
 type: 'create',
suffix_data: base64url ( {
  "delta_hash": DELTA_HASH,
  "recovery_key": JWK_OBJECT,
  "recovery_commitment": COMMITMENT_HASH
})
delta: base64url({
 "patches": PATCH_ARRAY,
  "update_commitment": COMMITMENT_HASH
})

}

Suffix = ASCII( Multihash( SHA256,  suffix_data ) )

troyronda:
Recommend a note that JSON Canonicalization MAY be performed: https://github.com/decentralized-identity/sidetree/issues/564

[...]

PR #1069, opened 2021-02-25
docs(spec): add short description of input test vector sequence

troyronda:
- adds description of the input test vector sequence to achieve the resolution results.
- extracts the operation inputs from generated.json into individual files.
- adds the extracted operation inputs to the appendix.

Related to #550 and #561.


OR13 (approved):

OR13:
@troyronda thanks! test vectors are the most valuable part of any software specification. These will help tremendously.

PR #1112, opened 2021-04-07
feat(ref-imp): added error code metadata for batch writer loop failure event

thehenrytsai:


lib/core/BatchScheduler.ts
```diff
@@ -55,9 +57,15 @@ export default class BatchScheduler {
 
       EventEmitter.emit(EventCode.SidetreeBatchWriterLoopSuccess, { batchSize });
     } catch (error) {
-      EventEmitter.emit(EventCode.SidetreeBatchWriterLoopFailure);
-      Logger.error('Unexpected and unhandled error during batch writing, investigate and fix:');
-      Logger.error(error);
+      let loopFailureEventData;
+      if (error instanceof SidetreeError && error.code !== ErrorCode.BlockchainWriteUnexpectedError) {
+        loopFailureEventData = { code: error.code };
```
isaacJChen, line 62: I think we want to emit unexpected code. Monitors and alerts can be setup to specifically detect unexpected error events.

thehenrytsai: Good point. Will have an unexpected error code for remaining case.

isaacJChen (approved):

PR #1116, opened 2021-04-19
fix(ref-imp): fix missing transaction when bitcoin rpc call fails

isaacJChen:


lib/bitcoin/SidetreeTransactionParser.ts
```diff
@@ -119,14 +122,27 @@ export default class SidetreeTransactionParser {
...
+        shouldRetry = false;
+      } catch (e) {
+        if (e instanceof FetchError && e.type === 'request-timeout') {
+          // timeout should retry until success.
+          shouldRetry = true;
+          EventEmitter.emit(EventCode.BitcoinTransactionParserTransactionFetchFailure);
+          Logger.warn(`Transaction fetch timed out for transaction id: ${transactionId}`);
+        } else {
+          // other errors mean the transaction is bad
+          shouldRetry = false;
```
thehenrytsai, line 140: Let's discuss, not sure this is the optimal thing to do. We can never be sure that "every other error condition" means the transaction is definitely "bad" right?

So I still feel throwing is still the way to go.

lib/bitcoin/SidetreeTransactionParser.ts
```diff
@@ -121,11 +123,11 @@ export default class SidetreeTransactionParser {
   private async fetchOutput (transactionId: string, outputIndexToFetch: number): Promise<BitcoinOutputModel | undefined> {
     try {
       const transaction = await this.bitcoinClient.getRawTransaction(transactionId);
-
       return transaction.outputs[outputIndexToFetch];
     } catch (e) {
       Logger.warn(`Error while trying to get outputIdx: ${outputIndexToFetch} from transaction: ${transactionId}. Error: ${SidetreeError.stringify(e)}`);
```
thehenrytsai, line 128:
```suggestion
      Logger.error(`Error while trying to get outputIdx: ${outputIndexToFetch} from transaction: ${transactionId}. Error: ${SidetreeError.stringify(e)}`);

```

lib/bitcoin/SidetreeTransactionParser.ts
```diff
@@ -121,11 +123,11 @@ export default class SidetreeTransactionParser {
   private async fetchOutput (transactionId: string, outputIndexToFetch: number): Promise<BitcoinOutputModel | undefined> {
     try {
       const transaction = await this.bitcoinClient.getRawTransaction(transactionId);
-
       return transaction.outputs[outputIndexToFetch];
     } catch (e) {
       Logger.warn(`Error while trying to get outputIdx: ${outputIndexToFetch} from transaction: ${transactionId}. Error: ${SidetreeError.stringify(e)}`);
-      return undefined;
+      EventEmitter.emit(EventCode.BitcoinTransactionParserFailure);
```
thehenrytsai, line 129: I'd argue that there is no need in throwing an event here, since a generic transaction processing failure event in the outer loop will be emitted up on this throw. If we really want this info to be doubled up, I'd say the thing to do is to throw a specific `SidetreeError`, so the error code is included as part of the failure label.

thehenrytsai (dismissed):

lib/bitcoin/EventCode.ts
```diff
@@ -9,5 +9,6 @@ export default {
   BitcoinLockMonitorLoopFailure: `bitcoin_lock_monitor_loop_failure`,
   BitcoinLockMonitorLoopSuccess: `bitcoin_lock_monitor_loop_success`,
   BitcoinObservingLoopFailure: 'bitcoin_observing_loop_failure',
-  BitcoinObservingLoopSuccess: `bitcoin_observing_loop_success`
+  BitcoinObservingLoopSuccess: `bitcoin_observing_loop_success`,
+  BitcoinTransactionParserFailure: 'bitcoin_transaction_parser_failure'
```
thehenrytsai, line 13: Remove.

thehenrytsai (approved):

PR #1137, opened 2021-06-11
feat(ref-imp): make core read blockchain time from db

isaacJChen:


lib/core/Observer.ts
```diff
@@ -69,6 +78,22 @@ export default class Observer {
     this.continuePeriodicProcessing = false;
   }
 
+  private async pullLatestBlockChainTime () {
```
thehenrytsai, line 81: Nit: casing, `Blockchain`

tests/core/Observer.spec.ts
```diff
@@ -404,7 +410,7 @@ describe('Observer', async () => {
 
     // Force blockchain time to be higher than the latest known transaction time by core,
```
thehenrytsai, line 411:
```suggestion
    // Force blockchain time to be higher than the cursor transaction time used in Observer,
```

lib/core/Core.ts
```diff
@@ -109,18 +112,32 @@ export default class Core {
...
+    const newApproximateTime = (await this.serviceStateStore.get()).approximateTime;
+    Logger.info(`Core approximateTime updated to: ${newApproximateTime}`);
+    this.approximateTime = newApproximateTime;
+
+    // shouldContinueTimePull is only used in tests to stop the pulling
+    if (this.shouldContinueTimePull) {
+      setTimeout(async () => this.startPeriodicTimePull(), this.approximateTimeUpdateIntervalInSeconds * 1000);
+    }
+  }
+
```
thehenrytsai, line 134: Design: can we move this logic into a separate class perhaps called `BlockchainClock`? `Core` is intentionally kept minimal to avoid becoming like another `BitcoinProcessor`.

lib/core/Observer.ts
```diff
@@ -69,6 +78,22 @@ export default class Observer {
...
+      EventEmitter.emit(EventCode.SidetreeBlockchainTimeChanged, { time: latestBlockchainTime.time });
+    } catch (e) {
+      Logger.error(`Error occured while updating blockchain time, investigate and fix: ${e}`);
+    } finally {
+      if (this.continuePeriodicProcessing) {
+        setTimeout(async () => this.pullLatestBlockChainTime(), this.blockchainTimePullIntervalInSeconds * 1000);
+      }
+    }
+  }
+
```
thehenrytsai, line 96: Optional design idea: what if this logic also goes into the `BlockchainClock` class since there is already a timer there? Then we decide whether to trigger time update based on a switch we passed in to the constructor? This way we remove the odd dependency that request handling has with Observer (through the service state store), and abstract it with `BlockchainClock`.

lib/core/Core.ts
```diff
@@ -102,41 +101,31 @@ export default class Core {
...
     if (this.config.observingIntervalInSeconds > 0) {
       await this.observer.startPeriodicProcessing();
+      // Only pull blockchain time when observer is enabled.
+      await this.blockchainClock.startPeriodicPullLatestBlockchainTime();
     } else {
       Logger.warn(LogColor.yellow(`Transaction observer is disabled.`));
     }
 
+    // Performance optimization to cache the time so requests don't have to go to db
+    this.blockchainClock.startPeriodicCacheTime();
```
thehenrytsai, line 111: What if we make it more generic by doing something like this:

```typescript
const enableRealBlockchainTimePull = this.config.observingIntervalInSeconds > 0;
this.blockchainClock.initialize(enableRealBlockchainTimePull);
```

With this we can then further simplify the two `setTimeout()` inside `BlockchainClock` into just one, I don't see them need to be run on different schedule.

lib/core/Core.ts
```diff
@@ -149,7 +138,7 @@ export default class Core {
    *   2. An encoded DID Document prefixed by the DID method name. e.g. 'did:sidetree:<encoded-DID-Document>'.
    */
   public async handleResolveRequest (didOrDidDocument: string): Promise<ResponseModel> {
-    const currentTime = this.approximateTime!;
+    const currentTime = this.blockchainClock.getApproximateTime()!;
```
thehenrytsai, line 141: Optional nit: we can consider rename `getApproximateTime()` to just `getTime()`, 'coz arguably how time is calculated now is an internal detail to `BlockchainClock`.

lib/core/BlockchainClock.ts
```diff
@@ -16,45 +16,44 @@ export default class BlockchainClock {
      * The interval which to pull and update blockchain time
      */
     private blockchainTimePullIntervalInSeconds = 60;
-
-    /**
-     * The interval which to cache the approximate time
-     */
-    private approximateTimeUpdateIntervalInSeconds: number = 60;
-
     private cachedApprocimateTime?: number;
```
thehenrytsai, line 19: Sorry missed this.
```suggestion
    private cachedApproximateTime?: number;
```

lib/core/BlockchainClock.ts
```diff
@@ -63,10 +62,16 @@ export default class BlockchainClock {
...
+      }
+    }
+
+    private async cacheTime () {
+      try {
+        const newApproximateTime = (await this.serviceStateStore.get()).approximateTime;
+        Logger.info(`Core approximateTime updated to: ${newApproximateTime}`);
+        this.cachedApprocimateTime = newApproximateTime;
+      } catch (e) {
+        Logger.error(`Error occured while caching blockchain time, investigate and fix: ${e}`);
```
thehenrytsai, line 74: Nit
```suggestion
        Logger.error(`Error occurred while caching blockchain time, investigate and fix: ${e}`);
```

lib/core/BlockchainClock.ts
```diff
@@ -16,45 +16,44 @@ export default class BlockchainClock {
...
+      await this.cacheTime();
+
+      if (this.continuePulling) {
+        setTimeout(async () => this.startPeriodicPullLatestBlockchainTime(), this.blockchainTimePullIntervalInSeconds * 1000);
+      }
+    }
+
+    private async pullRealBlockchainTime () {
       try {
         const latestBlockchainTime = await this.blockchain.getLatestTime();
```
thehenrytsai, line 58: Sanity: we should only emit this event (and optionally store) if time actually changed right?

lib/core/BlockchainClock.ts
```diff
@@ -63,10 +62,16 @@ export default class BlockchainClock {
...
-      } finally {
-        if (this.continuePulling) {
-          setTimeout(async () => this.startPeriodicPullLatestBlockchainTime(), this.blockchainTimePullIntervalInSeconds * 1000);
-        }
+      }
+    }
+
+    private async cacheTime () {
+      try {
+        const newApproximateTime = (await this.serviceStateStore.get()).approximateTime;
```
thehenrytsai, line 70: Both `cacheTime()` and `pullRealBlockchainTime` need this, so maybe just do it once? To me the main logic is more of an if-else thing:

```typescript
if (this.enableRealBlockchainTimePull) {
  // Update in-memory time if needed by pulling from bitcoin.
} else {
 // Update in-memory time from DB.
}

thehenrytsai (approved):
Approved with a minor PR on top this one, feel free to take it or leave it.

thehenrytsai (approved):

PR #1138, opened 2021-06-15
fix(ref-imp): automatically create bitcoin wallet

isaacJChen:


lib/bitcoin/BitcoinClient.ts
```diff
@@ -227,6 +229,23 @@ export default class BitcoinClient {
     };
   }
 
+  private async createWallet () {
+    const walletNameToUse = 'sidetreeDefaultWallet';
```
thehenrytsai, line 233: Sanity: this name can be anything and subsequent calls to add address as watch-only does not need to reference a wallet by name etc?

thehenrytsai (approved):

lib/bitcoin/BitcoinClient.ts
isaacJChen: Correct

PR #1140, opened 2021-06-17
fix(ref-imp): make bitcoin load wallet and specify wallet

isaacJChen:


lib/bitcoin/BitcoinClient.ts
```diff
@@ -230,20 +234,31 @@ export default class BitcoinClient {
...
-    } else {
-      // for the list of possible errors: https://github.com/bitcoin/bitcoin/blob/master/src/rpc/protocol.h
-      Logger.error(`Error code ${response.error.code} occured while attempting to create bitcoin wallet: ${response.error.message}`);
-    }
+    try {
+      await this.rpcCall(request, true, false);
+      Logger.info(`Wallet created with name "${this.walletNameToUse}".`);
+    } catch (e) {
+      Logger.error(`Error occured while attempting to create bitcoin wallet: ${e}`);
+    };
```
thehenrytsai, line 247: Sanity: I miss the point on why eating the error here instead of throw?

lib/bitcoin/BitcoinClient.ts
```diff
@@ -794,7 +809,13 @@ export default class BitcoinClient {
...
   }
 
-  private async rpcCall (request: any, timeout: boolean): Promise<any> {
+  /**
+   *
+   * @param request The request for the rpc call
+   * @param timeout Should timeout or not
+   * @param isWalletRPC Is wallet rpc or not. Should pass in true if the rpc call is called on specific wallet
+   */
+  private async rpcCall (request: any, timeout: boolean, isWalletRPC: boolean): Promise<any> {
```
thehenrytsai, line 818: Nit: camelCasing
```suggestion
  private async rpcCall (request: any, timeout: boolean, isWalletRpc: boolean): Promise<any> {

```

lib/bitcoin/BitcoinClient.ts
```diff
@@ -813,7 +834,15 @@ export default class BitcoinClient {
       };
     }
 
-    const response = await this.fetchWithRetry(this.bitcoinPeerUri.toString(), requestOptions, timeout);
+    let rpcUrl = this.bitcoinPeerUri.toString();
+    /**
+     * Specify the wallet to use if it is a wallet rpc call
+     * List of rpc calls categorized by type: https://developer.bitcoin.org/reference/rpc/
+     */
```
thehenrytsai, line 841: Style nit: use `//` for in-method documentation.

lib/bitcoin/BitcoinClient.ts
```diff
@@ -335,7 +350,7 @@ export default class BitcoinClient {
       method: 'getblockcount'
     };
 
-    const response = await this.rpcCall(request, true);
+    const response = await this.rpcCall(request, true, false);
```
thehenrytsai, line 353: Style rant: just FYI, generally hate this kinda innocent looking call. Normally prefer `const isWalletRpc = false` then pass that in, else no clue what all these true false things are. But since there is already an existing precedence... will let it go!

lib/bitcoin/BitcoinClient.ts
```diff
@@ -813,7 +834,15 @@ export default class BitcoinClient {
       };
     }
 
-    const response = await this.fetchWithRetry(this.bitcoinPeerUri.toString(), requestOptions, timeout);
+    let rpcUrl = this.bitcoinPeerUri.toString();
```
thehenrytsai, line 837: Sanity: isn't `this.bitcoinPeerUri` already a string?

lib/bitcoin/BitcoinClient.ts
isaacJChen: Will add it.

thehenrytsai (approved):

PR #1153, opened 2021-08-18
fix(ref-imp): fixes controller property to be the short form of the did

### LINKED ISSUE #1152: API data model should support passing in the "controller" property
Just was reviewing the DID Core test suite and I noticed that there's a few tests that are failing related to the controller property. As far as I can tell the API doesn't allow this to be set and the node when performing the transformation from the internal data model to the DID Document is where the controller property is being set.

Could we allow the API to accept definition of the controller property so that this can be set properly during the resolution process?

kdenhartog:
This is a bit of a hack that's potentially going to create issues when resolving
a key that's not controlled by the controller of the did document. At this point
though that's an exceptional case that doesn't need to be handled immediately and
this fix allows DID Documents resolved by the sidetree node to be compliant with
the did-test-suite.

BREAKING CHANGE: The controller is no longer an empty string and instead matches
the document id property (e.g. the did being resolved)

fixes #1152

### DIFF
--- a/lib/core/versions/latest/DocumentComposer.ts
+++ b/lib/core/versions/latest/DocumentComposer.ts
@@ -40,7 +40,7 @@ export default class DocumentComposer {
         const id = '#' + publicKey.id;
         const didDocumentPublicKey = {
           id: id,
-          controller: '',
+          controller: did.isShortForm ? did.shortForm : did.longForm,
           type: publicKey.type,
           publicKeyJwk: publicKey.publicKeyJwk
         };
@@ -51,7 +51,7 @@ export default class DocumentComposer {
         verificationMethod.push(didDocumentPublicKey);
 
         if (purposeSet.size > 0) {
-          const reference = didDocumentPublicKey.controller + didDocumentPublicKey.id;
+          const reference = didDocumentPublicKey.id;
 
           for (const purpose of purposeSet) {
             if (!verificationRelationships.has(purpose)) {
--- a/tests/core/DocumentComposer.spec.ts
+++ b/tests/core/DocumentComposer.spec.ts
@@ -13,7 +13,7 @@ import SidetreeError from '../../lib/common/SidetreeError';
 describe('DocumentComposer', async () => {
 
   describe('transformToExternalDocument', () => {
-    it('should output the expected resolution result given key(s) across all purpose types.', async () => {
+    it('should output the expected resolution result with a shortForm identifier given key(s) across all purpose types.', async () => {
       const [anySigningPublicKey] = await OperationGenerator.generateKeyPair('anySigningKey'); // All purposes will be included by default.
       const [noPurposePublicKey] = await OperationGenerator.generateKeyPair('noPurposePublicKey', []);
       const [authPublicKey] = await OperationGenerator.generateKeyPair('authPublicKey', [PublicKeyPurpose.Authentication]);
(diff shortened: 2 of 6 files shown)


csuwildcat:
@kdenhartog I will loop in @thehenrytsai for review, who returns from vacation on the 21st

lib/core/versions/latest/DocumentComposer.ts
```diff
@@ -40,7 +40,7 @@ export default class DocumentComposer {
         const id = '#' + publicKey.id;
         const didDocumentPublicKey = {
           id: id,
-          controller: '',
+          controller: did.shortForm,
```
thehenrytsai, line 43: For accuracy we can't have this always be hardcoded to short form (in case of a long form DID). I'd imagine you'll have to do something like:

```
did.isShortForm ? did.shortForm : did.longForm;
```

thehenrytsai:
Tests are failing, so need to fix those tests. Did a quick dig, looks like it's this piece of code:

```
          const reference = didDocumentPublicKey.controller + didDocumentPublicKey.id;
```

Seems like the quickest fix is that you'll want to remove the prefix of the `controller` value, while it will remove the future proofing intended by the code, but since we are hard coding the value to DID owner, no harm done.

thehenrytsai (changes requested):
Change requested and need to fix broken tests.

lib/core/versions/latest/DocumentComposer.ts
kdenhartog: I was thinking about switching to this, but this would imply that the long form identifier is different than the short form identifier based on which state is being resolved. This seemed off to me, so I opted to keep it always set to the shortForm which as I understood it was always available since it's just a shorter instance of the long form. What's the advantage you see to having the long form identifier available here?

isaacJChen: Long form DIDs will indicate that the short form is an equivilant DID in its metadata

```   
 "didDocumentMetadata": {
      "deactivated": true,
      "canonicalId": "did:sidetree:EiDyOQbbZAa3aiRzeCkV7LOx3SERjjH93EXoIM3UoN4oWg",
      "equivalentId": ["did:sidetree:EiDyOQbbZAa3aiRzeCkV7LOx3SERjjH93EXoIM3UoN4oWg"],
      "method": {
        "published": true,
        "recoveryCommitment": "EiBfOZdMtU6OBw8Pk879QtZ-2J-9FbbjSZyoaA_bqD4zhA",
        "updateCommitment": "EiDOrcmPtfMHuwIWN6YoihdeIPxOKDHy3D6sdMXu_7CN0w"
      }
    }
```

I think including the long form ID here when the resolving DID is long form is better, since someone could try to resolve the controller's DID and the short form can potentially be not found. The resolving party can always look at the meta data of the long form DID document and decide to resolve long form or short form so I don't think we will have circular dependency. I imagine something like

1. resolve a long form DID
2. controller is long form
3. try to resolve short form
4.a. short form exists, use short form to continue and use new controller which is short form
4.b. short form does not exist, use long form and the controller is still valid

tests/core/DocumentComposer.spec.ts
```diff
@@ -47,19 +47,19 @@ describe('DocumentComposer', async () => {
         verificationMethod: [
           {
             id: '#anySigningKey',
-            controller: '',
+            controller: did.shortForm,
```
isaacJChen, line 50: If we use long form controller as mentioned in the comment threads above, will need a new test for composing long form. I think simply copying the format of the `transformToExternalDocument` test and call `DocumentComposer.transformToExternalDocument(didState, did, published);` with published set to false should do the trick.

kdenhartog:
@isaacJChen and @thehenrytsai I've addressed your concerns in the latest update

kdenhartog:
> Tests are failing, so need to fix those tests. Did a quick dig, looks like it's this piece of code:
> 
> ```
>           const reference = didDocumentPublicKey.controller + didDocumentPublicKey.id;
> ```
> 
> Seems like the quickest fix is that you'll want to remove the prefix of the `controller` value, while it will remove the future proofing intended by the code, but since we are hard coding the value to DID owner, no harm done.

I thought about how we could keep this future proofing in here, but I found too many edge cases that probably weren't accounted for so went with your suggested solution of just removing the controller portion of that for now.

OR13 (approved):
This PR is required to show that ION  / Sidetree based DID Methods conforms to W3C DID Core TR.

thehenrytsai (approved):

thehenrytsai:
@kdenhartog appreciate that you kept the code coverage at 100%!

PR #1157, opened 2021-09-30
fix(ref-imp): integrate custom logger with mongodb

xinaxu:
Add mongodb logging to the default logger.

### DIFF
--- a/lib/bitcoin/lock/MongoDbLockTransactionStore.ts
+++ b/lib/bitcoin/lock/MongoDbLockTransactionStore.ts
@@ -1,5 +1,6 @@
 import { Collection, Db, Long, MongoClient } from 'mongodb';
 import Logger from '../../common/Logger';
+import MongoDbLogger from '../../common/MongoDbLogger';
 import SavedLockModel from './../models/SavedLockedModel';
 
 /**
@@ -26,7 +27,9 @@ export default class MongoDbLockTransactionStore {
    * Initializes this object by creating the required collection.
    */
   public async initialize () {
-    const client = await MongoClient.connect(this.serverUrl, { useNewUrlParser: true }); // `useNewUrlParser` addresses nodejs's URL parser deprecation warning.
+    // `useNewUrlParser` addresses nodejs's URL parser deprecation warning.
+    const client = await MongoClient.connect(this.serverUrl, { useNewUrlParser: true, logger: MongoDbLogger.customLogger });
+    MongoDbLogger.setCommandLogger(client);
     this.db = client.db(this.databaseName);
     this.lockCollection = await MongoDbLockTransactionStore.creatLockCollectionIfNotExist(this.db);
   }
--- a/lib/common/MongoDbLogger.ts
+++ b/lib/common/MongoDbLogger.ts
@@ -1,0 +1,27 @@
+import { LoggerState, MongoClient } from 'mongodb';
+import Logger from '../common/Logger';
+
+/**
+ * Base class that contains the common logging functions for mongodb.
+ */
+export default class MongoDbLogger {
+  static setCommandLogger (client: MongoClient) {
+    client.on('commandSucceeded', (event: any) => {
+      Logger.info(event);
+    });
+    client.on('commandFailed', (event: any) => {
+      Logger.warn(event);
+    });
+  }
+
+  static customLogger = function (_message: string | undefined, state: LoggerState | undefined): void {
+    switch (state?.type) {
+      case 'debug':
+      case 'info':
+        Logger.info(state);
+        break;
+      case 'error':
+        Logger.error(state);
+    }
+  };
+}
--- a/lib/common/MongoDbTransactionStore.ts
+++ b/lib/common/MongoDbTransactionStore.ts
@@ -1,6 +1,7 @@
 import { Collection, Cursor, Db, Long, MongoClient } from 'mongodb';
 import ITransactionStore from '../core/interfaces/ITransactionStore';
 import Logger from '../common/Logger';
+import MongoDbLogger from './MongoDbLogger';
 import TransactionModel from './models/TransactionModel';
 
 /**
@@ -19,7 +20,9 @@ export default class MongoDbTransactionStore implements ITransactionStore {
    * Initialize the MongoDB transaction store.
    */
   public async initialize (serverUrl: string, databaseName: string): Promise<void> {
-    const client = await MongoClient.connect(serverUrl, { useNewUrlParser: true }); // `useNewUrlParser` addresses nodejs's URL parser deprecation warning.
+    // `useNewUrlParser` addresses nodejs's URL parser deprecation warning.
+    const client = await MongoClient.connect(serverUrl, { useNewUrlParser: true, logger: MongoDbLogger.customLogger });
+    MongoDbLogger.setCommandLogger(client);
     this.db = client.db(databaseName);
     this.transactionCollection = await MongoDbTransactionStore.createTransactionCollectionIfNotExist(this.db);
   }
--- a/lib/core/MongoDbUnresolvableTransactionStore.ts
+++ b/lib/core/MongoDbUnresolvableTransactionStore.ts
@@ -1,6 +1,7 @@
 import { Collection, Db, Long, MongoClient } from 'mongodb';
 import IUnresolvableTransactionStore from './interfaces/IUnresolvableTransactionStore';
 import Logger from '../common/Logger';
+import MongoDbLogger from '../common/MongoDbLogger';
 import TransactionModel from '../common/models/TransactionModel';
 import UnresolvableTransactionModel from './models/UnresolvableTransactionModel';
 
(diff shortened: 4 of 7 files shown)


lib/common/MongoDbTransactionStore.ts
```diff
@@ -19,7 +20,9 @@ export default class MongoDbTransactionStore implements ITransactionStore {
    * Initialize the MongoDB transaction store.
    */
   public async initialize (serverUrl: string, databaseName: string): Promise<void> {
-    const client = await MongoClient.connect(serverUrl, { useNewUrlParser: true }); // `useNewUrlParser` addresses nodejs's URL parser deprecation warning.
+    // `useNewUrlParser` addresses nodejs's URL parser deprecation warning.
+    const client = await MongoClient.connect(serverUrl, { useNewUrlParser: true, logger: MongoDbLogger.customLogger });
```
thehenrytsai, line 24: Any reason why this construction doesn't do the below like others:
```
      monitorCommands: true,
      loggerLevel: 'debug'
```

lib/bitcoin/lock/MongoDbLockTransactionStore.ts
```diff
@@ -26,7 +27,14 @@ export default class MongoDbLockTransactionStore {
    * Initializes this object by creating the required collection.
    */
   public async initialize () {
-    const client = await MongoClient.connect(this.serverUrl, { useNewUrlParser: true }); // `useNewUrlParser` addresses nodejs's URL parser deprecation warning.
+    // `useNewUrlParser` addresses nodejs's URL parser deprecation warning.
```
thehenrytsai, line 30: I think this could be a good time to refactor all store class to inherit from base `MongoDbStore` so we can avoid all the duplicate code for instantiation.

The reason that this was not done earlier was just combination of  resourcing + `MongoDBStore` was introduced later than some of these store classes + 'don't touch classes if there is no need". So I opportunistically migrated some of these store classes overtime and it wasn't so bad from memory. It seems like the right time to do this for the remainder of the store classes.

lib/common/MongoDbLogger.ts
```diff
@@ -0,0 +1,31 @@
+import { LoggerState, MongoClient } from 'mongodb';
+import Logger from '../common/Logger';
+
+/**
+ * Base class that contains the common logging functions for mongodb.
+ */
+export default class MongoDbLogger {
+  static setCommandLogger (client: MongoClient) {
```
thehenrytsai, line 8: Sanity: did you run the tests against local mongoDB and CosmosDB?

lib/common/MongoDbLogger.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  static customLogger = function (_message: string | undefined, state: LoggerState | undefined): void {
+    switch (state?.type) {
+      case 'warn':
+        Logger.warn(state);
+        break;
+      case 'error':
+        Logger.error(state);
+        break;
+      case undefined:
```
thehenrytsai, line 25: Style: prefer to take this out as a check separately for readability.

lib/common/MongoDbLogger.ts
```diff
@@ -0,0 +1,31 @@
...
+  static setCommandLogger (client: MongoClient) {
+    client.on('commandSucceeded', (event: any) => {
+      Logger.info(event);
+    });
+    client.on('commandFailed', (event: any) => {
+      Logger.warn(event);
+    });
+  }
+
+  static customLogger = function (_message: string | undefined, state: LoggerState | undefined): void {
```
thehenrytsai, line 17: 1. Style: use consistent method syntax (like what you have in line 8).
2. Sanity: why do you use not to look message?
3. Style: explicitly define access modifiers (`public`).

```suggestion
  public static customLogger (_message: string | undefined, state: LoggerState | undefined): void {
```

lib/common/MongoDbLogger.ts
```diff
@@ -0,0 +1,31 @@
+import { LoggerState, MongoClient } from 'mongodb';
+import Logger from '../common/Logger';
+
+/**
+ * Base class that contains the common logging functions for mongodb.
+ */
+export default class MongoDbLogger {
```
thehenrytsai, line 7: Design: this class probably won't be necessary when you are done with `MongoDbStore` refactoring as you can probably hide these methods within it.

lib/common/MongoDbLogger.ts
```diff
@@ -0,0 +1,31 @@
...
+  static setCommandLogger (client: MongoClient) {
+    client.on('commandSucceeded', (event: any) => {
```
thehenrytsai, line 9: For my learning: why do you need to do both explicit event logging AND set custom logger? Is it because the custom logger doesn't actually log these events even though you've set it on `debug` level? If so, what does custom log? Do you have an example for either cases?

lib/common/MongoDbLogger.ts
thehenrytsai:
```suggestion
  public static enableCommandResultLogging (client: MongoClient) {
```

lib/common/MongoDbTransactionStore.ts
xinaxu: Fixed after refactoring

lib/bitcoin/BitcoinProcessor.ts
```diff
@@ -165,7 +165,7 @@ export default class BitcoinProcessor {
     await this.versionManager.initialize(versionModels, this.config, this.blockMetadataStore);
     await this.serviceStateStore.initialize();
     await this.blockMetadataStore.initialize();
-    await this.transactionStore.initialize(this.config.mongoDbConnectionString, this.config.databaseName);
```
thehenrytsai, line 168: Ironically, reading this PR I just remembered that I was convinced a while back that taking params in `initialize()` is a better pattern pattern taking it in constructor. So if you agree, let's flip it around.

lib/common/MongoDbLogger.ts
```diff
@@ -14,15 +18,25 @@ export default class MongoDbLogger {
...
+    switch (state.type) {
       case 'warn':
         Logger.warn(state);
         break;
       case 'error':
         Logger.error(state);
         break;
-      case undefined:
+      case 'debug':
+        Logger.debug(state);
```
thehenrytsai, line 39: Consider just reuse `.info()` because:
1. There we have not had the need to introduce yet another layer of logging, if something is generally useful, it's fine just let it be logged.
2. Our internal custom logger doesn't have concept of `.debug()` either, so more plumbing will be needed. Personally I don't think it's worth it.
```suggestion
        Logger.info(state);
```

lib/core/Core.ts
```diff
@@ -75,7 +75,7 @@ export default class Core {
       config.observingIntervalInSeconds
     );
 
-    this.monitor = new Monitor();
+    this.monitor = new Monitor(config);
```
thehenrytsai, line 78: Design: similar to earlier comment. This class is relatively "new", thus took the pattern of param-less constructor.

lib/core/MongoDbUnresolvableTransactionStore.ts
```diff
@@ -1,67 +1,44 @@
...
+   * @param databaseName The database name where the collection should be saved.
    * @param retryExponentialDelayFactor
    *   The exponential delay factor in milliseconds for retries of unresolvable transactions.
    *   e.g. if it is set to 1 seconds, then the delays for retries will be 1 second, 2 seconds, 4 seconds... until the transaction can be resolved.
    */
-  constructor (private serverUrl: string, private databaseName: string, retryExponentialDelayFactor?: number) {
+  public constructor (
+    serverUrl: string,
+    databaseName: string,
+    retryExponentialDelayFactor?: number) {
```
thehenrytsai, line 29: `retryExponentialDelayFactor` seems to be useless, remove it if it is bothering you.

tests/core/Monitor.spec.ts
```diff
@@ -3,24 +3,13 @@ import MongoDbTransactionStore from '../../lib/common/MongoDbTransactionStore';
 import { SidetreeMonitor } from '../../lib';
 import TransactionModel from '../../lib/common/models/TransactionModel';
 
-describe('Monitor', async () => {
+fdescribe('Monitor', async () => {
```
thehenrytsai, line 6: undo

lib/common/MongoDbLogger.ts
thehenrytsai: Still prefer to remove this class and just have the static methods in `MongoDbStore` as it only makes sense in that context. Happy to discuss.

lib/common/MongoDbLogger.ts
thehenrytsai: Still waiting on the comment regarding `_message`.

lib/common/MongoDbLogger.ts
thehenrytsai: Can we take the rename suggestion?

PR #1163, opened 2021-12-03
fix(ref-imp): refactored getWriterValueTimeLock() to handle error better

thehenrytsai:
Existing code parses too early, preventing other error handling code to take place.

### DIFF
--- a/lib/core/Blockchain.ts
+++ b/lib/core/Blockchain.ts
@@ -210,19 +210,17 @@ export default class Blockchain implements IBlockchain {
   }
 
   public async getWriterValueTimeLock (): Promise<ValueTimeLockModel | undefined> {
-
     const response = await this.fetch(this.writerLockUri);
-    const responseBodyString = await ReadableStream.readAll(response.body);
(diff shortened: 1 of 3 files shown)


csuwildcat (approved):

lib/core/Blockchain.ts
```diff
@@ -210,19 +210,17 @@ export default class Blockchain implements IBlockchain {
   }
 
   public async getWriterValueTimeLock (): Promise<ValueTimeLockModel | undefined> {
-
     const response = await this.fetch(this.writerLockUri);
-    const responseBodyString = await ReadableStream.readAll(response.body);
```
xinaxu, line 215: worth adding another negative test where the response.body is not readable or is not JSON string. Before the change, it will throw other exceptions. After the change, it no longer throws that exception since the reading/parsing logic moved to later place.

xinaxu (commented):
done review.

xinaxu (approved):

lib/core/Blockchain.ts
thehenrytsai: Sync offline.

PR #1167, opened 2022-01-06
fix(ref-imp): fixed an integer precision bug in how transaction number is constructed

thehenrytsai:
This was discovered in https://github.com/decentralized-identity/ion/issues/240

As a result, rewrote how transaction number is constructed to produce a smaller integer. Also took the opportunity to make the transaction number much more human-friendly also, the current transaction number format had been annoying me for a long time.

Specifically, the current transaction number is simply a large ugly looking number that doesn't make any sense to a human. It would be much easier to make sense of (thus debug), now that a transaction on index position 500 in block 777777 is 777777000500, as opposed to 3340526778581492. Notice the number is smaller AND easier to make sense of.

Also, due to this internal transaction number scheme change, both core and blockchain service will be required to be upgraded to the same build at the same time, a bit painful, but isn't worth more engineering time IMO to mitigate this inconvenience.

### DIFF
--- a/lib/bitcoin/BitcoinClient.ts
+++ b/lib/bitcoin/BitcoinClient.ts
@@ -928,7 +928,7 @@ export default class BitcoinClient {
       } else {
         networkError = new SidetreeError(
           ErrorCode.BitcoinClientFetchHttpCodeWithNetworkIssue,
-          `Network issue ith HTTP response: [${response.status}]: ${bodyBuffer}`
+          `Network issue with HTTP response: [${response.status}]: ${bodyBuffer}`
         );
 
         // Retry-able if one of these HTTP codes.
--- a/lib/bitcoin/BitcoinClient.ts
+++ b/lib/bitcoin/BitcoinClient.ts
@@ -928,7 +928,7 @@ export default class BitcoinClient {
       } else {
         networkError = new SidetreeError(
           ErrorCode.BitcoinClientFetchHttpCodeWithNetworkIssue,
-          `Unexpected HTTP response: [${response.status}]: ${bodyBuffer}`
+          `Network issue ith HTTP response: [${response.status}]: ${bodyBuffer}`
         );
 
         // Retry-able if one of these HTTP codes.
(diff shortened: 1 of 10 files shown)


lib/bitcoin/BitcoinClient.ts
```diff
@@ -928,7 +928,7 @@ export default class BitcoinClient {
       } else {
         networkError = new SidetreeError(
           ErrorCode.BitcoinClientFetchHttpCodeWithNetworkIssue,
-          `Unexpected HTTP response: [${response.status}]: ${bodyBuffer}`
+          `Network issue ith HTTP response: [${response.status}]: ${bodyBuffer}`
```
BenjaminMoe, line 931: "with"?

lib/bitcoin/TransactionNumber.ts
```diff
@@ -1,42 +1,68 @@
...
+
+    if (transactionIndexInBlock > TransactionNumber.maxTransactionIndexInBlock) {
+      throw new SidetreeError(
+        ErrorCode.TransactionNumberTransactionIndexInBlockTooLarge,
+        `Position index ${transactionIndexInBlock} given exceeds max allowed value of ${TransactionNumber.maxTransactionIndexInBlock}`
+      );
+    }
+
+    // Choosing a nice round number as long as it is less than `9007199254`.
+    const maxBlockNumber = 9000000000;
```
BenjaminMoe, line 34: Maybe something like `const JAVASCTIPT_MAX_INT = Math.pow(2, 53);` and then `const maxBlockNumber = JAVASCTIPT_MAX_INT` would describe what the comments say with code.

BenjaminMoe:
Looks good to me. I'm guessing that the scheme of using `777777000500` doesn't look like it would cause collision issues. And it looks like it allocates enough digits that there shouldn't be overflow problems with transaction number or transaction  number.

BenjaminMoe (dismissed):

xinaxu (approved):
LGTM

lib/bitcoin/BitcoinClient.ts
thehenrytsai: Fixed!

lib/bitcoin/TransactionNumber.ts
thehenrytsai: Will keep it as is, because we can't assign `JAVASCTIPT_MAX_INT_PRECISION` to `maxBlockNumber`, we would need to round it down to LESS than the nearest million, what we have now is simpler of the two.

xinaxu (approved):

PR #1168, opened 2022-01-10
Specify resolution result Content-Type

clehner:
This adds a requirement for a DID Resolution Result HTTP response to set a Content-Type header to indicate that it is a DID Resolution Result, as specified in DID Resolution:
- https://w3c-ccg.github.io/did-resolution/#did-resolution-result
- https://w3c-ccg.github.io/did-resolution/#bindings-https

Setting this Content-Type header is proposed in a PR to the ION implementation: https://github.com/decentralized-identity/ion/pull/244

### DIFF
--- a/docs/api/content.md
+++ b/docs/api/content.md
@@ -14,6 +14,8 @@ Sidetree defines `published`, `updateCommitment`, and `recoveryCommitment` metho
    - `updateCommitment` is the commitment for the next update operation as defined in [commitment schemes](https://identity.foundation/sidetree/spec/#commitment-schemes).
    - `recoveryCommitment` is the commitment for the next recover or deactivate operation as defined in [commitment schemes](https://identity.foundation/sidetree/spec/#commitment-schemes).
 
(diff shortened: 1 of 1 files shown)


docs/api/content.md
```diff
@@ -14,6 +14,8 @@ Sidetree defines `published`, `updateCommitment`, and `recoveryCommitment` metho
    - `updateCommitment` is the commitment for the next update operation as defined in [commitment schemes](https://identity.foundation/sidetree/spec/#commitment-schemes).
    - `recoveryCommitment` is the commitment for the next recover or deactivate operation as defined in [commitment schemes](https://identity.foundation/sidetree/spec/#commitment-schemes).
 
+The DID Resolution Result response ****MUST**** contain a `Content-Type` HTTP header with value `application/ld+json;profile="https://w3id.org/did-resolution"`.
```
csuwildcat, line 17: I am wondering why we would set LD here, as we were hoping to use just plain JSON. The output doc should parse as either, but I think folks were inclined to use plain JSON as the default.

Issue #1169, opened 2022-01-13
Update operations commit-reveal chain may lead to an unresolvable DID


shakreiner:
The current iterations over Update operation using the commit-reveal chain may lead to an infinite loop when trying to resolve a DID.
The thing that can cause this state is having a DID with 2 Update operations, where the second operation commits to the first one again.
This situation should not be allowed based on the [specification](https://identity.foundation/sidetree/spec/#update) (`updateCommitment` must be a _new_ commitment in every operation), but the code doesn't seem to enforce that.

I successfully created this state locally (by writing operations to the node's local mongoDB) using the following code (it uses [ion-tools](https://github.com/decentralized-identity/ion-tools)):
```Javascript
let RawIonSdk = require('@decentralized-identity/ion-sdk');
const Multihash_1 = require("./ion-tools/node_modules/@decentralized-identity/ion-sdk/dist/lib/Multihash");
const ION = require('./ion-tools')

async function createUpdateOp(didSuffix, updateKeys, nextUpdatePublicKey, patches){
    const hashAlgorithmInMultihashCode = ION.SDK.IonSdkConfig.hashAlgorithmInMultihashCode;

    const revealValue = Multihash_1.default.canonicalizeThenHashThenEncode(updateKeys.publicJwk, hashAlgorithmInMultihashCode);
    const nextUpdateCommitmentHash = Multihash_1.default.canonicalizeThenDoubleHashThenEncode(nextUpdatePublicKey, hashAlgorithmInMultihashCode);
    const delta = {
        patches,
        updateCommitment: nextUpdateCommitmentHash
    };
    const deltaHash = Multihash_1.default.canonicalizeThenHashThenEncode(delta, hashAlgorithmInMultihashCode);
    const dataToBeSigned = {
        updateKey: updateKeys.publicJwk,
        deltaHash: deltaHash
    };
    const compactJws = await ION.SDK.LocalSigner.create(updateKeys.privateJwk).sign({ alg: 'ES256K' }, dataToBeSigned);
    return {
        type: 'update',
        didSuffix: didSuffix,
        revealValue,
        delta,
        signedData: compactJws
    };
}

async function main() {

    let didSuffix = 'EiCd1__Gfn_YkYXjWq62HkowAXq9d2YWON_di4BQnqTDbQ'

    let firstUpdateKeys = JSON.parse('{"publicJwk":{[REDACTED]},"privateJwk":{[REDACTED]}')
[...]

PR #1172, opened 2022-01-19
fix(ref-imp): #1169 - fixed bug on commit-reveal chain leading to infinite loop in resolution

thehenrytsai:
Fix to issue raised: https://github.com/decentralized-identity/sidetree/issues/1169

### DIFF
--- a/lib/core/Resolver.ts
+++ b/lib/core/Resolver.ts
@@ -104,15 +104,18 @@ export default class Resolver {
    */
   private async applyRecoverAndDeactivateOperations (startingDidState: DidState, commitValueToOperationMap: Map<string, AnchoredOperationModel[]>)
     : Promise<DidState> {
-    let didState = startingDidState;
+    // This stores all commitment hash of the corresponding reveal value of operations that have been successfully applied,
+    // such that no infinite loop of logical operation chain can be created.
+    const commitValuesUsed = new Set<string>();
 
+    let didState = startingDidState;
     while (commitValueToOperationMap.has(didState.nextRecoveryCommitmentHash!)) {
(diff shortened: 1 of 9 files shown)


xinaxu (dismissed):

OR13 (approved):

lib/core/Resolver.ts
```diff
@@ -104,28 +104,34 @@ export default class Resolver {
    */
   private async applyRecoverAndDeactivateOperations (startingDidState: DidState, commitValueToOperationMap: Map<string, AnchoredOperationModel[]>)
     : Promise<DidState> {
-    let didState = startingDidState;
+    // This stores all commitment hash of the corresponding reveal value of operations that have been successfully applied,
```
tplooker, line 107:
```suggestion
    // This stores all commitment hashes of the corresponding reveal values of the operations that have been successfully applied,
```

lib/core/Resolver.ts
```diff
@@ -136,22 +142,27 @@ export default class Resolver {
    */
   private async applyUpdateOperations (startingDidState: DidState, commitValueToOperationMap: Map<string, AnchoredOperationModel[]>)
     : Promise<DidState> {
-    let didState = startingDidState;
+    // This stores all commitment hash of the corresponding reveal value of operations that have been successfully applied,
```
tplooker, line 145:
```suggestion
    // This stores all commitment hashes of the corresponding reveal values of the operations that have been successfully applied,
```

tplooker (approved):

lib/core/Resolver.ts
thehenrytsai: Fixed.

lib/core/Resolver.ts
thehenrytsai: Fixed.

xinaxu (approved):

PR #1174, opened 2022-01-28
fix(ref-imp): allow setting controller property

clehner:
Allow setting the controller property of a public key entry (verification method object), as described here: https://identity.foundation/sidetree/spec/v1.0.0/#add-public-keys

### DIFF
--- a/lib/core/versions/latest/DocumentComposer.ts
+++ b/lib/core/versions/latest/DocumentComposer.ts
@@ -40,7 +40,7 @@ export default class DocumentComposer {
         const id = '#' + publicKey.id;
         const didDocumentPublicKey = {
           id: id,
-          controller: did.isShortForm ? did.shortForm : did.longForm,
+          controller: publicKey.controller || (did.isShortForm ? did.shortForm : did.longForm),
           type: publicKey.type,
           publicKeyJwk: publicKey.publicKeyJwk
         };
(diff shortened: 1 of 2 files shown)


OR13 (approved):

csuwildcat:
@thehenrytsai this would be a breaking consensus change for existing implementations. Can we accept this and plan an ION release separately?

lib/core/versions/latest/DocumentComposer.ts
```diff
@@ -218,7 +218,7 @@ export default class DocumentComposer {
 
     const publicKeyIdSet: Set<string> = new Set();
     for (const publicKey of publicKeys) {
-      const allowedProperties = new Set(['id', 'type', 'purposes', 'publicKeyJwk']);
+      const allowedProperties = new Set(['id', 'type', 'purposes', 'publicKeyJwk', 'controller']);
```
thehenrytsai, line 221: The linked spec says the controller must be a DID and optionally not exceed some maximum length. Need to add such checks and tests accordingly.

thehenrytsai (changes requested):
Recommend (also responding to @csuwildcat) to just copy the changes to `latest` folder to `1.0` also so that we don't need to snapshot an additional version folder just for this change. I'd be shocked if there were any operations that were discarded because of inclusion of a `controller` property.

zkwzk:
hello, any update for this change? does sidetree plan to support controller property

thehenrytsai:
Was waiting for the contributor to update the PR to merge this. I'd love to know more about the scenarios/use cases of this property from anyone who's an advocate for the feature, I can probably make the requested changes myself and merge it in!

PR #1175, opened 2022-02-04
Wait for 6 blocks after last transaction confirmation before writting another transaction

xinaxu:


lib/core/MongoDbConfirmationStore.ts
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  public async confirm (anchorString: string, confirmedAt: number | undefined): Promise<void> {
+    await this.collection.findOneAndUpdate({ anchorString }, { confirmedAt });
+  }
+
+  public async getLastSubmitted (): Promise<{ submittedAt: number; confirmedAt: number | undefined } | null> {
+    const response: ConfirmationModel[] = await this.collection.find().sort({ submittedAt: -1 }).limit(1).toArray();
+    if (response.length === 0) {
+      return null;
```
thehenrytsai, line 27: Use `undefined` as opposed to `null` if you can.

lib/core/versions/latest/BatchWriter.ts
```diff
@@ -49,6 +52,14 @@ export default class BatchWriter implements IBatchWriter {
       return 0;
     }
 
+    const lastSubmitted = await this.confirmationStore.getLastSubmitted();
+    if (lastSubmitted !== null &&
+      (lastSubmitted.confirmedAt === undefined || currentTime.time - lastSubmitted.confirmedAt < this.minConfirmationBetweenWrites - 1)
```
thehenrytsai, line 57: Why the -1 at the end?

lib/core/versions/latest/BatchWriter.ts
```diff
@@ -101,6 +113,8 @@ export default class BatchWriter implements IBatchWriter {
     // Remove written operations from queue after batch writing has completed successfully.
     await this.operationQueue.dequeue(numberOfOperations);
 
+    await this.confirmationStore.submit(stringToWriteToBlockchain, currentTime.time);
```
thehenrytsai, line 116: Do it before line 114, else might still write quicker than required.

lib/core/MongoDbConfirmationStore.ts
```diff
@@ -0,0 +1,45 @@
...
+interface ConfirmationModel {
+  anchorString: string;
+  submittedAt: number;
+  confirmedAt: number | undefined;
+}
+
+/**
+ * Implementation of LastWriteStore that stores the last update per write
+ */
+export default class MongoDbConfirmationStore extends MongoDbStore implements IConfirmationStore {
```
thehenrytsai, line 13: Should we add index for both `submittedAt` and `anchorString`?

lib/core/Observer.ts
```diff
@@ -124,6 +126,7 @@ export default class Observer {
           this.transactionsUnderProcessing.push(transactionUnderProcessing);
           // Intentionally not awaiting on downloading and processing each operation batch.
           this.processTransaction(transaction, transactionUnderProcessing);
+          await this.confirmationStore.confirm(transaction.anchorString, transaction.transactionTime);
```
thehenrytsai, line 129: Probably should do this within `processTransaction()` method given that `processTransaction()` is called in another place.

lib/core/Observer.ts
```diff
@@ -340,6 +343,10 @@ export default class Observer {
 
     await this.unresolvableTransactionStore.removeUnresolvableTransactionsLaterThan(bestKnownValidRecentTransactionNumber);
 
+    for (const transaction of await this.transactionStore.getTransactionsLaterThan(bestKnownValidRecentTransactionNumber, undefined)) {
```
thehenrytsai, line 346: Can we do this by calling one method and pass in the transaction time? Something like `this.confirmationStore.removeConfirmationsLaterThan(bestKnownValidRecentTransactionTime)`?

lib/core/versions/latest/BatchWriter.ts
xinaxu: Say comfirmedAt = 100, currentTime = 105, then the message have already been confirmed 6 times in block 100 to 105. Hence it's okay to proceed. It only needs to wait if currentTime - confirmedAt < 5. What do you think?

lib/core/MongoDbConfirmationStore.ts
xinaxu: Yes

lib/core/Observer.ts
xinaxu: That may be complicated. To do that, I'll need to add transaction time to the data model. And instead of removing the entry, I need to reset the confirmedAt field. Since the reorg will change the transactionTime, I will need to update the transactionTime as well.

lib/core/MongoDbConfirmationStore.ts
```diff
@@ -0,0 +1,58 @@
...
+  constructor (serverUrl: string, databaseName: string) {
+    super(serverUrl, MongoDbConfirmationStore.collectionName, databaseName);
+  }
+
+  public async confirm (anchorString: string, confirmedAt: number): Promise<void> {
+    await this.collection.updateMany({ anchorString }, { $set: { confirmedAt } });
+  }
+
+  public async resetAfter (confirmedAt: number | undefined): Promise<void> {
+    await this.collection.updateMany({ confirmedAt: { $gt: confirmedAt } }, { $set: { confirmedAt: null } });
```
thehenrytsai, line 25: Add a unit test for `undefined`, I think it's edge case, but since it can be passed, should clear the entire collection in which case.

lib/core/versions/latest/BatchWriter.ts
```diff
@@ -27,11 +28,14 @@ import ValueTimeLockVerifier from './ValueTimeLockVerifier';
  * Implementation of the `IBatchWriter`.
  */
 export default class BatchWriter implements IBatchWriter {
```
thehenrytsai, line 30: For some reason branch coverage for BatchWriter is no longer 100%, bring it back please!

tests/json/config-test.json
```diff
@@ -6,5 +6,6 @@
   "maxConcurrentDownloads": 20,
   "mongoDbConnectionString": "mongodb://localhost:27017/",
   "observingIntervalInSeconds": 10,
+  "minConfirmationBetweenWrites": 6,
```
thehenrytsai, line 9: Remove.

lib/core/MongoDbConfirmationStore.ts
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  public async getLastSubmitted (): Promise<{ submittedAt: number; confirmedAt: number | undefined } | undefined> {
```
thehenrytsai, line 28: Consider return `ConfirmationModel` so you can log even more like what you've done with `submittedAt`.

xinaxu:
All addressed.
@csuwildcat would you mind take a look and approve?

tests/core/BatchWriter.spec.ts
```diff
@@ -1,29 +1,32 @@
...
 describe('BatchWriter', () => {
-  let blockchain: IBlockchain;
+  let blockchain: MockBlockchain;
   let cas: ICas;
   let operationQueue: IOperationQueue;
   let batchWriter: BatchWriter;
+  let confimrationStore: MockConfirmationStore;
 
   beforeAll(() => {
+    jasmine.DEFAULT_TIMEOUT_INTERVAL = 1000000;
```
thehenrytsai, line 23: What's the purpose of this?

thehenrytsai (approved):

Issue #1052, opened 2021-02-18
deactivated test vector response missing deactivated flag


troyronda:
`tests/vectors/resolution/afterDeactivate.json` is missing the `deactivated` flag in the metadata.

csuwildcat:
This hasn't been implemented in the reference yet, so we should just add this later. I don't think this should be a v1 blocker.

troyronda:
Updated in spec version of text vectors.

Issue #1015, opened 2021-01-26
Align implementation with resolution metadata portion of the spec


thehenrytsai:
The implementation deviated from the latest changes to the resolution metadata portion of the spec. e.g. `deactivated` property is newly added in the spec.

OR13:
You may want to add ION tests to did core test suite, at some point. for example: https://github.com/w3c/did-test-suite/pull/19

These would help did core cover cases like `deactivated`

OR13:
Provide concrete examples....

```
"didDocumentMetadata": {
    "deactivated": true,
    "canonicalId": "did:sidetree:EiDyOQbbZAa3aiRzeCkV7LOx3SERjjH93EXoIM3UoN4oWg",
    "method": {
      "published": true,
      "recoveryCommitment": "EiBfOZdMtU6OBw8Pk879QtZ-2J-9FbbjSZyoaA_bqD4zhA",
      "updateCommitment": "EiDOrcmPtfMHuwIWN6YoihdeIPxOKDHy3D6sdMXu_7CN0w"
    }
  }
```

csuwildcat:
Action: Daniel update spec to reflect empty DID Doc with id prop if deactivated.

troyronda:
Spec aspect is completed.

troyronda:
@csuwildcat @thehenrytsai the test vector for `after deactivate` response is missing the `deactivated` flag.

PR #1178, opened 2022-02-14
fix(ref-imp): set deactivated metadata property

clehner:
Set `deactivated` DID document metadata property in resolution result of deactivated DID.

Re: #1015, #1052

### DIFF
--- a/lib/core/versions/1.0/DocumentComposer.ts
+++ b/lib/core/versions/1.0/DocumentComposer.ts
@@ -128,7 +128,8 @@ export default class DocumentComposer {
       method: {
         published
       },
-      canonicalId: did
+      canonicalId: did,
+      deactivated: true
     };
     return {
       '@context': DocumentComposer.resolutionObjectContextUrl,
--- a/lib/core/versions/latest/DocumentComposer.ts
+++ b/lib/core/versions/latest/DocumentComposer.ts
@@ -128,7 +128,8 @@ export default class DocumentComposer {
       method: {
         published
       },
-      canonicalId: did
+      canonicalId: did,
+      deactivated: true
     };
     return {
       '@context': DocumentComposer.resolutionObjectContextUrl,
--- a/tests/core/DocumentComposer.spec.ts
+++ b/tests/core/DocumentComposer.spec.ts
@@ -239,7 +239,8 @@ describe('DocumentComposer', async () => {
         method: {
           published
         },
-        canonicalId: 'did:method:suffix'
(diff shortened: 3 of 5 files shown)


clehner:
@csuwildcat I don't think that is the case. The change is to set the [deactivated](https://www.w3.org/TR/did-core/#dfn-deactivated) property in the DID resolution result (in the DID document metadata object) when resolving a DID that has had a deactivate operation applied.

lib/core/versions/latest/RequestHandler.ts
```diff
@@ -181,7 +181,7 @@ export default class RequestHandler implements IRequestHandler {
       const document = DocumentComposer.transformToExternalDocument(didState, did, published);
 
       // Status is different if DID is deactivated.
-      const didDeactivated = didState.nextRecoveryCommitmentHash === undefined;
+      const didDeactivated = document.didDocumentMetadata.deactivated;
```
clehner, line 184: Note: the change on this line is not really necessary, but I thought it might be useful for readability to make use of the newly added `deactivated` property. This should be equivalent as before, since `transformToExternalDocument` (called above) returns `createDeactivatedResolutionResult(...)` when `didState.nextRecoveryCommitmentHash === undefined`:
https://github.com/decentralized-identity/sidetree/blob/c3171462b7ee945ac6209cb3a21a607332953299/lib/core/versions/latest/DocumentComposer.ts#L25-L30
... and `createDeactivatedResolutionResult` would set `deactivated: true` in the `didDocumentMetadata` object:
https://github.com/decentralized-identity/sidetree/pull/1178/files#diff-a5730fc426536637a041ba8b67308520ed5fdce4c785ebcfcfa8496c5c22ef4cR132

thehenrytsai: I agree it would be more readable, but I think the intent here is not to depend on knowledge of the returned external document and rely on `didState as much as possible, to minimize chance of breakage of this code in case for whatever reason the data model of the external document is changed. Which is why document type is `any`, as the code does not make any assumption about its schema.

thehenrytsai (changes requested):
Thanks for fixing the long lingering item! The change looks good overall, but do reverse the deactivate conditional check change!

lib/core/versions/latest/RequestHandler.ts
clehner: Makes sense. Amended commit to remove this change: 8744f65f36b206b090c8e41c793c4d8ae3a8aa05

PR #1183, opened 2022-04-07
Fix undefined when calculating normalized fee

xinaxu:


lib/bitcoin/ErrorCode.ts
```diff
@@ -25,5 +25,6 @@ export default {
   TransactionNumberBlockNumberTooLarge: 'transaction_number_block_number_too_large',
   TransactionNumberTransactionIndexInBlockTooLarge: 'transaction_number_transaction_index_in_block_too_large',
   ValueTimeLockInPendingState: 'value_time_lock_in_pending_state',
-  VersionManagerVersionStringNotFound: 'version_manager_version_string_not_found'
+  VersionManagerVersionStringNotFound: 'version_manager_version_string_not_found',
+  NormalizedFeeCalculatorBlockNotFound: 'normalized_fee_calculator.block_not_found'
```
thehenrytsai, line 29: Keep style consistent
```suggestion
  NormalizedFeeCalculatorBlockNotFound: 'normalized_fee_calculator_block_not_found'
```

lib/bitcoin/lock/LockMonitor.ts
```diff
@@ -213,7 +213,8 @@ export default class LockMonitor {
 
     } catch (e) {
 
-      if (e instanceof SidetreeError && e.code === ErrorCode.LockResolverTransactionNotConfirmed) {
+      if (e instanceof SidetreeError &&
+        (e.code === ErrorCode.LockResolverTransactionNotConfirmed || e.code === ErrorCode.NormalizedFeeCalculatorBlockNotFound)) {
         // This means that the transaction was broadcasted but hasn't been written on the blockchain yet.
```
thehenrytsai, line 218:
```suggestion
        // This means that the transaction was broadcasted but hasn't been written on the blockchain yet, or
        // transaction was broadcasted, but hasn't been observed by `BitcoinProcessor` yet.
```

lib/bitcoin/versions/latest/NormalizedFeeCalculator.ts
```diff
@@ -72,6 +74,9 @@ export default class NormalizedFeeCalculator implements IFeeCalculator {
     // TODO: #943 This is left here because it may be versioned. Move it out if we confirm it will not be versioned.
     // https://github.com/decentralized-identity/sidetree/issues/943
     const blockMetadata = await this.blockMetadataStore.get(block, block + 1);
+    if (blockMetadata.length === 0) {
```
thehenrytsai, line 77: Will need to copy the change to `1.0` folder also.

lib/bitcoin/versions/latest/NormalizedFeeCalculator.ts
```diff
@@ -72,6 +74,9 @@ export default class NormalizedFeeCalculator implements IFeeCalculator {
...
+      throw new SidetreeError(ErrorCode.NormalizedFeeCalculatorBlockNotFound);
```
thehenrytsai, line 78: Please add reference to `NormalizedFeeCalculatorBlockNotFound` as an expected error thrown in the `IFeeCalculator.getNormalizedFee()` method because this now become part of the contract between lock monitor and versioned code.

lib/bitcoin/interfaces/IFeeCalculator.ts
```diff
@@ -7,6 +7,8 @@ import BlockMetadataWithoutNormalizedFee from '../models/BlockMetadataWithoutNor
 export default interface IFeeCalculator {
   /**
    * Returns the fee in satoshis for a particular block height.
+   * @throws SidetreeError with ErrorCode.NormalizedFeeCalculatorBlockNotFound if the block is not observed by the
+   * BitcoinProcessor yet
```
thehenrytsai, line 11: It's an interface, shouldn't know about `BitcoinProcessor`.
```suggestion
   * @throws SidetreeError with ErrorCode.NormalizedFeeCalculatorBlockNotFound if fee is not found for the given block.
```

lib/bitcoin/versions/1.0/NormalizedFeeCalculator.ts
```diff
@@ -72,6 +74,9 @@ export default class NormalizedFeeCalculator implements IFeeCalculator {
     // TODO: #943 This is left here because it may be versioned. Move it out if we confirm it will not be versioned.
     // https://github.com/decentralized-identity/sidetree/issues/943
     const blockMetadata = await this.blockMetadataStore.get(block, block + 1);
+    if (blockMetadata.length === 0) {
+      throw new SidetreeError(ErrorCode.NormalizedFeeCalculatorBlockNotFound);
```
thehenrytsai, line 78: Sanity: Please make sure code coverage remains at 100%.

thehenrytsai:
@xinaxu gentle ping regarding the 100% code coverage comment.

xinaxu:
> @xinaxu gentle ping regarding the 100% code coverage comment.

Done. Check again?

thehenrytsai:
Hey @xinaxu, can you also add a test to confirmation store so that it also hits 100% coverage? I know it is technically not part of this PR, but since the test file is already touched, which pointed to coverage that was missed.

You can probably just search and copy existing test `should create collection on initialization if it does not exist.`

xinaxu:
@thehenrytsai

thehenrytsai (changes requested):
Really like the idea of always being able to test against a mongodb, but there is one small problem with this change: these tests are intended to be tested against both official mongodb releases AND cosmosDB because their behaviors are actually different, I am guilty for not having the pending test message being more specific, but it was meant to call attention to whomever is modifying/reviewing the code to make sure those tests work against their own database implementation, in your case, cosmosDB, I don't have the account to test against cosmosbd going forward. By always adding a in-memory db now, no one would really know/remember that these tests need to run against cosmosdb, thus it probably means that regressions/incompabilities in cosmosdb will not be detected until too late.

I kinda still hope that the need to run the tests against real DB is somehow messaged to the code modifiers/reviewers somehow.

Happy to chat on slack directly and hear more suggestions.

xinaxu:
> but do remember again, that MS (currently you) will be responsible for testing against cosmosDB 'coz probably no one else will.

That's fine. If anything breaks, we will make a pull request.

thehenrytsai (approved):

xinaxu:
@thehenrytsai another option is to use a cosmosdb connection string as part of the github action and run the npm test against that cosmosdb. But let's do that when we have time.

Issue #692, opened 2020-05-27
Consider to not requiring bitcoin service to wait on full sync during initialization


thehenrytsai:
Today the bitcoin service is required to perform full transaction processing until the current block height during initialization, and if an unexpected error is thrown (such as dependent service being down), the system process exits. It would be more ideal that once the service is started, it can enter the polling mode quickly such it does not depend on the bitcoin client's availability for initialization.

PR #1194, opened 2022-06-22
feat(ref-imp): added built-in polling for Bitcoin Core readiness

thehenrytsai:
This has been an ask from many people for a while and finally got around to implement it:

- https://github.com/decentralized-identity/sidetree/issues/692
- https://github.com/decentralized-identity/ion/issues/150

In summary, up until now, the Sidetree Bitcoin service expects Bitcoin Core to be in a fully sync-ed and ready state when it starts up, it exits if Bitcoin Core isn't ready to serve requests.

This behavior had been fine for deployment frameworks like kubernetes which restarts the service automatically. But painful for manual setups. This will also help simplify dockerization, because we can now fire-and-forget after starting the container.

Also used this opportunity to add better logging and more documentation on the "fast initialization" code.

### DIFF
--- a/lib/bitcoin/BitcoinClient.ts
+++ b/lib/bitcoin/BitcoinClient.ts
@@ -841,17 +841,17 @@ export default class BitcoinClient {
 
   /**
    *
-   * @param request The request for the rpc call
+   * @param request The request for the RPC call
    * @param timeout Should timeout or not
-   * @param isWalletRpc Is wallet rpc or not. Should pass in true if the rpc call is called on specific wallet
+   * @param isWalletRpc Must set to `true` if the RPC is wallet-specific; `false` otherwise.
    */
   private async rpcCall (request: any, timeout: boolean, isWalletRpc: boolean): Promise<any> {
-    // append some standard jrpc parameters
+    // Append some standard RPC parameters.
     request.jsonrpc = '1.0';
     request.id = Math.round(Math.random() * Number.MAX_SAFE_INTEGER).toString(32);
 
     const requestString = JSON.stringify(request);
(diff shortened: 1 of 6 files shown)


lib/bitcoin/BitcoinClient.ts
```diff
@@ -68,20 +69,66 @@ export default class BitcoinClient {
...
+
+        const isWalletRpc = false;
+        const allowTimeout = true;
+        const response = await this.rpcCall(request, allowTimeout, isWalletRpc);
+        const blockHeight = response.headers;
+        const syncedBlockHeight = response.blocks;
+
+        Logger.info(LogColor.lightBlue(`Bitcoin sync progress: block height ${LogColor.green(blockHeight)}, sync-ed: ${LogColor.green(syncedBlockHeight)}`));
+
+        if (syncedBlockHeight === blockHeight) {
```
LiranCohen, line 99: Could this check ever be true if block headers and block height have not been synced? So Height and Headers are both zero?

Not sure if this could be the case but maybe make sure headers are not zero?

xinaxu (dismissed):

lib/bitcoin/BitcoinClient.ts
thehenrytsai: Good call! I felt it was edge case as syncing header is literally the first thing bitcoin core does when it fires up, but better safe than sorry, 'coz if network is down upon start up or something...

csuwildcat (approved):

csuwildcat:
Thank you for doing this, Henry, I know this cadence/wait thing has tripped folks up in the past, so to see it go is really cool.

LiranCohen:
Yeah this is an excellent change, good work!

PR #1196, opened 2022-07-28
docs(spec): add also known as patches

troyronda:


csuwildcat:
We should probably tag a patch version for this, right?

csuwildcat (approved):

kdenhartog (approved):

OR13 (approved):

Issue #1198, opened 2022-09-14
DIDComm Messaging Spec demands array for service endpoint which isn't allowed by Sidetree


nikolockenvitz:
Hi, while trying to implement DIDComm, a colleague of mine noticed that there is a problem with Sidetree-/ION-based DIDs.

## Context & Problem

In particular, the DIDComm messaging protocol demands an array for the serviceEndpoint field (in the service property of a DID document). In their [spec](https://identity.foundation/didcomm-messaging/spec/#service-endpoint), they provide the following example:

```json
{
    "id": "did:example:123456789abcdefghi#didcomm-1",
    "type": "DIDCommMessaging",
    "serviceEndpoint": [{
        "uri": "https://example.com/path",
        "accept": [
            "didcomm/v2",
            "didcomm/aip2;env=rfc587"
        ],
        "routingKeys": ["did:example:somemediator#somekey"]
    }]
}
```

However, this does not align with the Sidetree specification, which does not allow arrays. Sidetree only allows strings (URIs) and JSON objects, but no arrays:

https://github.com/decentralized-identity/sidetree/blob/bcc0cdb0f9263d15a95825faf94a7cb2240a05ba/docs/spec/patches.md?plain=1#L145

But actually, the DIDComm spec is compliant with the current [DID specification](https://www.w3.org/TR/did-core/), because there it says that

> The value of the serviceEndpoint property MUST be a [string](https://infra.spec.whatwg.org/#string), a [map](https://infra.spec.whatwg.org/#string), or a [set](https://infra.spec.whatwg.org/#ordered-set) composed of one or more [strings](https://infra.spec.whatwg.org/#string) and/or [maps](https://infra.spec.whatwg.org/#string).
> -- [Decentralized Identifiers v1.0, 5.4 Services](https://www.w3.org/TR/did-core/#services)

As you can see, there is a mismatch between the DID spec and the Sidetree spec.

## Requested Action

I think that Sidetree should also support arrays for the serviceEndpoint field and not discard corresponding PATCH actions (as the spec demands currently). Otherwise, Sidetree-based DID methods become unusable for protocols that build on top of the DID standard.

Any takes? Or does anyone know a good workaround how to use DIDComm with Sidetree-based DID methods (e.g. ION)?

csuwildcat:
[...]

PR #1200, opened 2022-09-20
Support array for serviceEndpoint

nikolockenvitz:
related to #1198 @csuwildcat

should the breaking change notice be part of the merge commit?


OR13 (dismissed):

csuwildcat:
@thehenrytsai looks like a good one to pull in, given the DID Comms enablement

troyronda (approved):
Makes sense to support an array of service endpoints.

BenjaminMoe (approved):

csuwildcat:
@nikolockenvitz I would add the breaking change to the commit, yes

lib/core/versions/latest/DocumentComposer.ts
```diff
@@ -345,22 +345,26 @@ export default class DocumentComposer {
...
-          throw new SidetreeError(
-            ErrorCode.DocumentComposerPatchServiceEndpointStringNotValidUri,
-            `Service endpoint string '${serviceEndpoint}' is not a valid URI.`
-          );
-        }
-      } else if (typeof serviceEndpoint === 'object') {
-        // Allow `object` type only if it is not an array.
-        if (Array.isArray(serviceEndpoint)) {
-          throw new SidetreeError(ErrorCode.DocumentComposerPatchServiceEndpointCannotBeAnArray);
+      // transform URI strings and JSON objects into array so that we can run validations more easily
```
thehenrytsai, line 348: Thanks for the awesome change! Can you copy the change under `versions/latest` (so far just this file) and overwrite [versions/1.0](https://github.com/decentralized-identity/sidetree/tree/master/lib/core/versions/1.0) folder also?  I don't think the change is worth divergent behavior and cutting a v1.1 for.

lib/core/versions/latest/models/ServiceModel.ts
```diff
@@ -4,5 +4,5 @@
 export default interface ServiceModel {
   id: string;
   type: string;
-  serviceEndpoint: string | object;
+  serviceEndpoint: string | object | Array<string | object>;
```
thehenrytsai, line 7: Probably no need for this change, array is an object.

lib/core/versions/latest/DocumentComposer.ts
```diff
@@ -345,22 +345,26 @@ export default class DocumentComposer {
...
+              `Service endpoint string '${serviceEndpoint}' is not a valid URI.`
+            );
+          }
+        } else if (typeof serviceEndpoint === 'object') {
+          // Allow `object` type only if it is not an array.
+          if (Array.isArray(serviceEndpoint)) {
+            throw new SidetreeError(ErrorCode.DocumentComposerPatchServiceEndpointCannotBeAnArray);
+          }
+        } else {
+          throw new SidetreeError(ErrorCode.DocumentComposerPatchServiceEndpointMustBeStringOrNonArrayObject);
```
thehenrytsai, line 366: We will need the same change in [ion-sdk](https://github.com/decentralized-identity/ion-sdk/blob/38f4cbea1fed7dacdc2dc63d9a346eeffe0b4e66/lib/IonRequest.ts#L328). Totally get it if you don't have time to make those changes in `ion-sdk`, so the PR will suffice to be approved if you can file an issue in the `ion-sdk` repo to make the same changes! Thanks!

thehenrytsai (changes requested):
Sorry for taking so long to get back. The PR looks good overall, just a couple of comments for this PR to actually take effect.

lib/core/versions/latest/models/ServiceModel.ts
nikolockenvitz: That's true, I just thought it might be better to understand if it's explicit. But I can also revert this change if you like.

lib/core/versions/latest/DocumentComposer.ts
nikolockenvitz: Happy to implement this in ion-sdk as well. I will still create an issue as well, as I am not yet 100% sure when I will find time for the implementation. So I will open the issue, once this PR is merged, right?

nikolockenvitz: Hi @thehenrytsai, I updated ion-sdk (decentralized-identity/ion-sdk#30) as well

nikolockenvitz:
@thehenrytsai can you please check whether all your concerns are addressed or point out required changes?

PR #1201, opened 2022-09-28
Handle HTTP 503 statuses returned from blockchain service

viathefalcon:
Change to not throw an error if the underlying blockchain service is unavailable when retrieving lock objects.

### DIFF
--- a/lib/core/Blockchain.ts
+++ b/lib/core/Blockchain.ts
@@ -214,6 +214,10 @@ export default class Blockchain implements IBlockchain {
     if (response.status === HttpStatus.NOT_FOUND) {
       return undefined;
     }
+
+    // Treat a HTTP 503 from the blockchain service as a non-error and
+    // return `undefined` such that a minimum-size batch will be
+    // written instead of the batch write being skipped
     if (response.status === HttpStatus.SERVICE_UNAVAILABLE) {
       return undefined;
     }
--- a/lib/core/Blockchain.ts
+++ b/lib/core/Blockchain.ts
@@ -201,6 +201,9 @@ export default class Blockchain implements IBlockchain {
     if (response.status === HttpStatus.NOT_FOUND) {
       return undefined;
     }
+    if (response.status === HttpStatus.SERVICE_UNAVAILABLE) {
+      return undefined;
+    }
 
     if (response.status !== HttpStatus.OK) {
       throw new SidetreeError(CoreErrorCode.BlockchainGetLockResponseNotOk, `Response: ${responseBodyString}`);
@@ -212,6 +215,9 @@ export default class Blockchain implements IBlockchain {
   public async getWriterValueTimeLock (): Promise<ValueTimeLockModel | undefined> {
     const response = await this.fetch(this.writerLockUri);
     if (response.status === HttpStatus.NOT_FOUND) {
+      return undefined;
+    }
+    if (response.status === HttpStatus.SERVICE_UNAVAILABLE) {
       return undefined;
     }
 
--- a/lib/core/Blockchain.ts
+++ b/lib/core/Blockchain.ts
@@ -201,9 +201,6 @@ export default class Blockchain implements IBlockchain {
     if (response.status === HttpStatus.NOT_FOUND) {
       return undefined;
     }
-    if (response.status === HttpStatus.SERVICE_UNAVAILABLE) {
-      return undefined;
-    }
 
     if (response.status !== HttpStatus.OK) {
       throw new SidetreeError(CoreErrorCode.BlockchainGetLockResponseNotOk, `Response: ${responseBodyString}`);
--- a/tests/core/Blockchain.spec.ts
+++ b/tests/core/Blockchain.spec.ts
@@ -380,23 +380,6 @@ describe('Blockchain', async () => {
       done();
     });
 
-    it('should return undefined if service-unavailable is returned by the network call.', async (done) => {
-      const blockchainClient = new Blockchain('unused');
-
-      const mockFetchResponse = {
-        status: 503,
-        body: ''
-      };
-
-      spyOn(blockchainClient as any, 'fetch').and.returnValue(Promise.resolve(mockFetchResponse));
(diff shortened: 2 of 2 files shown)


thehenrytsai (changes requested):
Thanks @viathefalcon for the PR!

Upon a look, changes to `getValueTimeLock()` definitely cannot be made as is because it could lead to diverging behavior in the Observer depending on network health of a running node, the error thrown is to allow re-evaluation of the transaction again later. It does look like there could be an improvement however (but even this counts as a fork so we need to make sure this would not affect ION operations in the past): it looks like `getValueTimeLock()` didn't have to called really if the number of operations doesn't exceed `maxNumberOfOperationsForNoValueTimeLock` in `verifyLockAmountAndThrowOnError(...)`


`getWriterValueTimeLock()` on the other hand is safe(r) to change since it only the Writer. But can you also explain the intent (perhaps by adding comments inline) on why the desire to continue the write a potential smaller batch here and not wait for the call to exceed? I agree a case can be made here to write a small batch here, but then it begs the question why not do the same for the error on line 225 right?

Let me know if I am not making any sense!

viathefalcon:
@thehenrytsai, thank you for looking into this.

The motivation for this change is that we have sidetree and the blockchain service deployed via Kubernetes which will occasionally respond to sidetree with a `503`, e.g. when the blockchain service's pod is being (re)deployed, and this is reported as a failure.

When I looked at how sidetree handled responses from the blockchain service, they seemed to fall into three categories: success, failure and "neither of the above". A "service (temporarily) unavailable" seemed a better fit for the last category than the second (or the first). It's not so much "something is wrong - stop what you're doing" as it is "call me back later when I might have a better answer for you". 

I did not appreciate the downstream effects of the change to `getValueTimeLock()` (I did look, but the semantics didn't register with me, and it didn't seem to be called out in the Blockchain REST API doc, either). But, in practice I'm only seeing the error reported against `getWriterValueTimeLock()`, so I'm content to undo the change to `getValueTimeLock()`.

viathefalcon:
@thehenrytsai would you be (more) open to a different approach, of accepting an instance of `IBlockchain` as a parameter to the constructor to https://github.com/decentralized-identity/sidetree/blob/bcc0cdb0f9263d15a95825faf94a7cb2240a05ba/lib/core/Core.ts#L53 with a default value of the current implementation (https://github.com/decentralized-identity/sidetree/blob/master/lib/core/Blockchain.ts), allowing callers to inject their own implementation?

Thanks!

thehenrytsai:
@viathefalcon, sorry for the delay. Your suggestion to focus on changes to `getWriterValueTimeLock()` sounds good to me, my suggestion would then be to:

1. treat line 221 ~ 224 the same (return `undefined`)
2. add inline comment to clearly explain the intent/why and the consequence (that a minimum batch will be written as opposed to a failure that causes this batch write to be skipped).

viathefalcon:
@thehenrytsai done!

lib/core/Blockchain.ts
```diff
@@ -215,6 +215,13 @@ export default class Blockchain implements IBlockchain {
...
+    // Treat a HTTP 503 from the blockchain service as a non-error and
+    // return `undefined` such that a minimum-size batch will be
+    // written instead of the batch write being skipped
+    if (response.status === HttpStatus.SERVICE_UNAVAILABLE) {
+      return undefined;
+    }
+
     const responseBodyString = await ReadableStream.readAll(response.body);
     if (response.status !== HttpStatus.OK) {
       throw new SidetreeError(CoreErrorCode.BlockchainGetWriterLockResponseNotOk, `Status: ${response.status}. Response: ${responseBodyString}.`);
```
thehenrytsai, line 227: Shouldn't this also result in minimum-size batch to written also?

viathefalcon: @thehenrytsai: Ah, sorry it did not register with me earlier that you meant to change the behaviour in response to _all_ non-200/404 status codes. My intent is to only change the behaviour in response to a 503, to be the same as to a 404.

We have a deployment in which a sort-of proxy will occasionally respond to requests to `${uri}/writerlock` with a 503. In the existing implementation of `Blockchain.ts`, this will result in an error being thrown. By the time the error makes it to the point in the call stack at which it is caught, it cannot be distinguished from other (actual) errors, and appears as a failure even though it isn't really.

So all I want to do is not treat a 503 as an error; other status codes are beyond the scope of this PR.

thehenrytsai: I understand as I've seen that occasional behavior. = ]

At minimum you'll need to update the interface description with this new behavior since this is a deviation to existing contract.
```javascript
  /**
   * Gets the lock object required for batch writing.
   *
   * @returns the lock object if one exist; undefined otherwise.
   * @throws SidetreeError with ErrorCode.ValueTimeLockInPendingState if the lock is not yet confirmed on the blockchain.
   */
  getWriterValueTimeLock (): Promise<ValueTimeLockModel | undefined>;
```

But from interface design standpoint, it is rather odd that `undefined` is returned for one error but not for another, hence the suggestion to just written `undefined` for all errors as a compromise while keeping the interface clean.

viathefalcon: Hi @thehenrytsai, I don't mean to seem like I'm insisting on introducing any inconsistency. I'd happily abandon this PR if you could approve https://github.com/decentralized-identity/sidetree/pull/1203, instead; this would enable us to customise the behaviour of `Blockchain.ts` to accommodate our deployment environment without changing the semantics of the interface here.

thehenrytsai: Hey @viathefalcon, you don't sounds like insistent at all, and seriously thanks for looking into any of this!

Your contribution and insights here are super valuable and thus I really want to figure out a happy middle ground to get this PR checked in so your work doesn't go to waste and everyone can benefit! But I also know your time is limited and valuable, so if you can't make further investment for this one I totally understand.

I have reviewed, approved, and merged your dependency injection PR to at least unblock you.

PR #1204, opened 2022-10-13
fix(ref-imp): update mongo db version

ANDuff:
Update to mongo db version to prevent error as noted here: https://jira.mongodb.org/browse/NODE-3252.

Previously if Mongo lost connection it would not recover without restarting the sidetree consuming app. This change makes the connection recover and sidetree processing to resume without restarting.

### DIFF
--- a/lib/common/MongoDbStore.ts
+++ b/lib/common/MongoDbStore.ts
@@ -64,7 +64,7 @@ export default class MongoDbStore {
       logger: MongoDbStore.customLogger,
       monitorCommands: true,
       loggerLevel: 'error',
-      useUnifiedTopology: true
+      useUnifiedTopology: true // prevents unrecoverable lost of connection: https://jira.mongodb.org/browse/NODE-3252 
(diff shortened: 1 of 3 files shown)


lib/common/MongoDbStore.ts
```diff
@@ -63,7 +63,8 @@ export default class MongoDbStore {
       useNewUrlParser: true,
       logger: MongoDbStore.customLogger,
       monitorCommands: true,
-      loggerLevel: 'error'
+      loggerLevel: 'error',
+      useUnifiedTopology: true
```
thehenrytsai, line 67:
```suggestion
      useUnifiedTopology: true // prevents unrecoverable lost of connection: https://jira.mongodb.org/browse/NODE-3252 
```

thehenrytsai (changes requested):
This is amazing!! Just a minor suggestion.

thehenrytsai (approved):

